PR #801, opened 2014-12-19
WELD-1808 EnhancedListener should not shutdown Weld if

mkouba:
org.jboss.weld.environment.servlet.Listener is registered


jharting:
Merged, thanks!

PR #997, opened 2015-05-29
Backport of WELD-1478 to EAP 6. Possible memory leak in HttpConversationContextImpl.

rpelisse:
[BZ1224358](https://bugzilla.redhat.com/show_bug.cgi?id=1224358) - Backport of WELD-1478 possible memory leak in HttpConversationContextImpl to 1.1.x for EAP.

PR #1019, opened 2015-06-10
Backport of WELD-1478 to EAP 6. Possible memory leak in HttpConversationContextImpl.

rpelisse:
Superseed [PR997](https://github.com/weld/core/pull/997).


weld-tester:
Can one of the admins verify this patch?

rpelisse:
Houston, we got a problem :(

I've ran a longer test today, just for the sake of it, and to ensure that new commit had not change the behavior, and it turn out the memory leak is still here. I've tried also with the first commit, also on a longer scale, and the memory leak is still here :(

I need to reinvestigate this, so closing the PR in the meantime.

PR #1095, opened 2015-07-27
WELD-2001 temporary workaround for WFLY-4915.

tremes:


jharting:
ok to test

jharting:
Only WildFly 10.0.0.Alpha5 is affected by this bug. Since Alpha6 is available already I am going to reject this PR.

PR #1087, opened 2015-07-23
WELD-2002 Switch to jboss.org templates for reference documentation

jharting:

PR #1099, opened 2015-07-27
WELD-2002 Switch to jboss.org templates for reference documentation

mkouba:
This should replace https://github.com/weld/core/pull/1087


jharting:
Merged, thanks!

PR #1103, opened 2015-07-29
WELD-2003 Improve Weld.initialize() javadoc so that it is clear that a

mkouba:
shutdown hook is registered automatically


mkouba:
This should be backported to 2.3 as well. I'm not going to create a separate branch as it's javadoc update only.

jharting:
Merged, thanks!

PR #1105, opened 2015-07-30
Use weld-se-shaded in tck runner.

tremes:
I think we can use weld-se-shaded artifact as dependency for TCK SE tests execution. It is little bit more clear.


mkouba:
retest this please

mkouba:
Merged, thanks!

PR #1110, opened 2015-08-12
WELD-1962 Implicit bean archive support in weld-se

mkouba:
- support implicit bean archives without beans.xml
- also add weld-se-parent and weld-se-tests

### DIFF
--- a/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/ClassPathBeanArchiveScanner.java
+++ b/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/ClassPathBeanArchiveScanner.java
@@ -23,7 +23,6 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.security.AccessController;
-import java.util.Enumeration;
 import java.util.List;
 import java.util.Set;
 import java.util.regex.Pattern;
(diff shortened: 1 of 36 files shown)


mkouba:
I'm going to write some more tests. Unfortunately, `arquillian-container-se` does not support directory class path entries yet.

jharting:
retest this please

mkouba:
Waiting until https://github.com/tremes/arquillian-container-se/pull/7 is merged/reviewed.

environments/se/tests/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+      <dependency>
+         <groupId>org.jboss.arquillian.junit</groupId>
+         <artifactId>arquillian-junit-container</artifactId>
+         <scope>test</scope>
+      </dependency>
+
+      <dependency>
+         <groupId>org.jboss.arquillian.container</groupId>
+         <artifactId>container-se-managed</artifactId>
+         <!-- Override temporarily - until the TCK switches to the new API -->
```
mkouba: This should be removed if merged after https://github.com/weld/core/pull/1121

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/ClassPathBeanArchiveScanner.java
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+    }
+
+    private void scanJarFile(File entryFile, ImmutableList.Builder<ScanResult> results) throws IOException {
+        try (ZipFile zip = new ZipFile(entryFile)) {
+            Enumeration<? extends ZipEntry> entries = zip.entries();
+            BeansXml beansXml = null;
+            boolean extensionFound = false;
+            while (entries.hasMoreElements()) {
```
jharting: Unless I am missing something it would be likely better to use this method: https://docs.oracle.com/javase/8/docs/api/java/util/zip/ZipFile.html#getEntry-java.lang.String-
instead of iterating over all the entries here.

mkouba: Yes, definitely. I forgot to update this part of the code after I found this method during the work on https://github.com/weld/core/pull/1119 :-)

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -627,7 +629,7 @@ public Weld setClassLoader(ClassLoader classLoader) {
      * @param resourceLoader
      * @param bootstrap
      */
-    protected Deployment createDeployment(ResourceLoader resourceLoader, CDI11Bootstrap bootstrap) {
```
jharting: I've seen the Weld class being subclassed more than once. Changing the signature of this protected method is likely to break those subclasses.

mkouba: Ok. I will revert the signature change.

jharting:
Merged, thanks!

PR #1115, opened 2015-08-14
Probe bean detail - various fixes and improvements

mkouba:
- do not include delegate injection points in dependency graph
- show associated decorators
- show beans a decorator is associated to
- show class-level interceptor bindings
- don't render dependeny graph if there are no data


jharting:
Merged, thanks!

PR #1116, opened 2015-08-17
Fix UndertowSmokeTest

mkouba:
- the test currently passes because the Weld-specific undertow extension
is loaded no matter what container is detected by Weld


jharting:
Merged, thanks!

PR #1119, opened 2015-08-17
WELD-2011 Simplify JandexIndexBeanArchiveHandler

mkouba:
- index cache is not needed
- fail fast if a bean reference is not a zip file
- open archive and try to get index file instead of using URL


jharting:
Merged, thanks!

PR #1128, opened 2015-08-27
WELD-2017 Fix conversation context deactivation

mkouba:
- state threadlocal must be removed during deactivation


jharting:
Merged, thanks!

PR #1149, opened 2015-09-10
WELD-2025 Improve JSON escaping in Probe

mkouba:
- also limit the portion of toString() data returned to the client and
stored for events
- remove unused JS functions


mkouba:
Merged.

PR #1153, opened 2015-09-18
WELD-2032 update Jetty integration to use new classes.

tremes:


mkouba:
I disagree to drop support of previous versions (e.g. 9.2) of Jetty in Weld 2.3. Instead we should probably encourage users to make use of the CDI/Weld Jetty module (we already mention the module in FAQ and docs).

tremes:
Yes it's reasonable. So I am going to close this one and will update only the doc.

PR #1161, opened 2015-10-02
WELD-2035 add warning log.

tremes:


impl/src/main/java/org/jboss/weld/logging/BootstrapLogger.java
```diff
@@ -194,4 +194,8 @@
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 145, value = "Bean identifier index built:\n  {0}", format = Format.MESSAGE_FORMAT)
     void beanIdentifierIndexBuilt(Object info);
+
+    @LogMessage(level = Level.WARN)
+    @Message(id = 146, value = "Method BeforeBeanDiscovery.addAnnotatedType(AnnotatedType<?>) is deprecated!\n", format = Format.MESSAGE_FORMAT)
```
mkouba: Please remove the `\n` char. Maybe we should also add some more info, e.g. the underlying java class of the given AT.

impl/src/main/java/org/jboss/weld/logging/BootstrapLogger.java
```diff
@@ -194,4 +194,8 @@
     @LogMessage(level = Logger.Level.DEBUG)
     @Message(id = 145, value = "Bean identifier index built:\n  {0}", format = Format.MESSAGE_FORMAT)
     void beanIdentifierIndexBuilt(Object info);
+
+    @LogMessage(level = Level.WARN)
+    @Message(id = 146, value = "Method BeforeBeanDiscovery.addAnnotatedType(AnnotatedType<?>) is used for {0}. This method is deprecated!", format = Format.MESSAGE_FORMAT)
```
mkouba: I think we should mention the version it's deprecated from... what about using the text from javadoc as a base?
> This method is deprecated from CDI 1.1 and addAnnotatedType(AnnotatedType, String) should be used instead.

mkouba:
Merged, thanks!

PR #1170, opened 2015-10-09
Weld Probe: Added test profile which allows to execute all tests (int…

manovotn:
…egration and ftest) with one command.

Given the latest test addition to Probe tests, we now have integration and ftests. While it is good to have them separated I think we might also make use of a profile which will fire off all the tests. 
For instance our release Jenkins jobs will surely want to run the whole test suite.

WDYT @mkouba @tremes ?

### DIFF
--- a/probe/tests/pom.xml
+++ b/probe/tests/pom.xml
@@ -295,6 +295,62 @@
                 </plugins>
             </build>
         </profile>
+        
+        <profile>
+            <id>alltests</id>
+            <build>
+                <plugins>
+                    <plugin>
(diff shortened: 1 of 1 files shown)


tremes:
Well I think it's not neccessary. I think it makes more sense to execute it separately especially with fact that ftests needs environment with firefox installed. 
I prefer to have one extra line of Maven execution (or even completely new job) than adding this profile. 
I am not also sure if this new profile will run on Tomcat.

mkouba:
Let's leave it separated.

PR #1178, opened 2015-10-13
Weld-2033 document enhancements to Weld SE bootstrap. Also added test…

manovotn:
… for multiple Weld instance creation.

I went through the docs and added a section about bootstrap new features.
Furthermore I added a short note to Weld builder javadoc to show the capabilities it has.
When inspecting tests I found out that we have it covered in WeldBuilderTest class except for a test which would verify creation of multiple Weld instances, so I created one.

### DIFF
--- a/environments/se/core/src/test/java/org/jboss/weld/environment/se/test/builder/WeldBuilderTest.java
+++ b/environments/se/core/src/test/java/org/jboss/weld/environment/se/test/builder/WeldBuilderTest.java
@@ -62,7 +62,7 @@ public class WeldBuilderTest {
         assertNull(WeldContainer.instance("FOO"));
         // Test alternatives selected for the synthetic BDA
         try (WeldContainer container = weld.beanClasses(Foo.class, Bar.class, Cat.class).alternatives(Bar.class)
-                .alternativeStereotypes(AlternativeStereotype.class).initialize()) {
+            .alternativeStereotypes(AlternativeStereotype.class).initialize()) {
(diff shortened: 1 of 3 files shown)


environments/se/core/src/test/java/org/jboss/weld/environment/se/test/builder/WeldBuilderTest.java
```diff
@@ -211,4 +211,19 @@ public void testNoBeanArchivesFound() {
         new Weld().disableDiscovery().initialize();
     }
 
+    @Test
+    public void testMultipleWeldInstancesCreated() {
```
mkouba: `WeldBuilderTest.testInitializedContainers()` is already testing multiple containers. So maybe we could just rename the original test method and add the `WeldContainer.getRunningContainerIds()` assertions?

mkouba:
Looks good!

mkouba:
add to whitelist

mkouba:
ok to test

environments/se/core/src/test/java/org/jboss/weld/environment/se/test/builder/WeldBuilderTest.java
manovotn: Hmm... I must have missed that test. Anyway I will modify the PR as you suggested, it seems pretty reasonable.

mkouba:
Merged, thanks!

PR #1179, opened 2015-10-15
Probe - events improvements

mkouba:
- observer methods view - add bean archive info for declaring beans
- event info storage


mkouba:
Merged.

PR #1182, opened 2015-10-16
Weld-2033 document enhancements to Weld SE bootstrap.

manovotn:
Equal to /pull/1178 but against master.


mkouba:
Merged, thanks!

PR #1184, opened 2015-10-22
Test sanity check

manovotn:
This is a result of test sanity check in tests-arquillian module, here is some breakdown of what was done:

* review test sanity, most tests (unsurprisingly) make sense and test what they claim to
* several tests were fixed/had asserts added
* reviewed @Ignored tests and added some back to our test suite
* checked TODOs and FIXMEs, reduce their amount
* 4 tests had their names changed from 'TestBean...' to 'Bean...Test' to align with majority
* empty test methods were reviewed -> methods now have a name which makes sense (intead of 'test' or 'go') and if they are empty, a commentary was added explaining whether an exception is expected
* no tests were removed (I rather tried to justify their existence or alter them to have some)

Here is a breakdown of a test run:
* 2.3 HEAD
 -> Tests run: 977, Failures: 0, Errors: 0, Skipped: 21
* 2.3 testSanityCheck
 -> Tests run: 981, Failures: 0, Errors: 0, Skipped: 15

I left the PR as several commits because the changes affect piles of files but I can squash them together if the need be.

WDYT?


manovotn:
Will inspect the failure + I noticed all the formatting changes.
I will check what can be done to make my NetBeans behave.

manovotn:
Updated the PR...
I cleansed the code from the formatting mess I had there before. Mostly it is just small adjustments, I added some test and altered others. Also there was one removed (NonProxyableTest) because it was ignored and has no real value (was expecting failure where I couldn't see reason for it).

mkouba:
Ok, do you plan a similar sanity check for master branch?

manovotn:
Definitely. In fact part of that work is already done on my branche, but I yet have to take a deeper look at it as I suppose there will be multiple details/changes.

mkouba:
Merged, thanks!

PR #1185, opened 2015-10-23
WELD-2051 Allow to skip the registration of a shutdown hook

mkouba:
- also fix ShutdownHookTest (Undertow was not stopped correctly)

### DIFF
--- a/environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
+++ b/environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
@@ -16,6 +16,7 @@
  */
 package org.jboss.weld.environment.se;
 
+import java.security.AccessController;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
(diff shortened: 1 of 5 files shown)


environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
```diff
@@ -152,6 +154,11 @@ static WeldContainer initialize(String id, WeldManager manager, Bootstrap bootst
         return weldContainer;
     }
 
+    private static boolean isShutdownHookNeeded() {
+        String property = AccessController.doPrivileged(new GetSystemPropertyAction(Weld.SHUTDOWN_HOOK_SYSTEM_PROPERTY));
+        return property == null || Boolean.FALSE.equals(Boolean.valueOf(property));
```
astefanutti: Shouldn't that be `TRUE`?

mkouba: No, `org.jboss.weld.se.shutdownHook=true` means *skip* the registration. I will change the system property name... this one is confusing. Thanks for feedback!

mkouba: @astefanutti When I think about it... it will be better to use `org.jboss.weld.se.shutdownHook` (and follow the Weld config convention) and change this logic. So that one would have to use `org.jboss.weld.se.shutdownHook=false` to skip the registration.

astefanutti: @mkouba yes I agree.

mkouba:
Merged.

PR #1186, opened 2015-10-23
WELD-2051 Allow to skip the registration of a shutdown hook

mkouba:
- also fix ShutdownHookTest (Undertow was not stopped correctly)


mkouba:
Merged.

PR #1192, opened 2015-10-29
WELD-2055 added a profile to exclude shutdown hook test on Windows. M…

manovotn:
…entioned shutdown hook behaviour in docs.


mkouba:
Merged, thanks!

PR #1196, opened 2015-11-05
WELD-2059 fail if observer method is EJB business remote method.

tremes:


impl/src/main/java/org/jboss/weld/bean/DisposalMethod.java
```diff
@@ -123,7 +123,7 @@ private void checkDisposalMethod(EnhancedAnnotatedMethod<T, ? super X> enhancedA
         }
         if (declaringBean instanceof SessionBean<?>) {
             SessionBean<?> sessionBean = (SessionBean<?>) declaringBean;
-            Set<MethodSignature> businessMethodSignatures = sessionBean.getBusinessMethodSignatures();
+            Set<MethodSignature> businessMethodSignatures = sessionBean.getLocalBusinessMethodSignatures();
```
mkouba: I wonder whether a remote business method may also be a producer/disposer method?

tremes: Good point. I think it could be nowadays however the question is whether it makes sense when you call this method "remotely" - which means you call it by name and do not really on CDI to produce the instance for you. I think it could be worth of spec issue.

tremes: Yes we have only: 
| If a non-static method of a session bean class has a parameter annotated @Disposes , and the
| method is not a business method of the session bean, the container automatically detects the
| problem and treats it as a definition error."

tremes:
Updated to include remote business methods in disposer methods.

impl/src/main/java/org/jboss/weld/bean/DisposalMethod.java
```diff
@@ -123,7 +123,9 @@ private void checkDisposalMethod(EnhancedAnnotatedMethod<T, ? super X> enhancedA
...
+            // see CDI-571 and remove when it's agreed
+            businessMethodSignatures.addAll(sessionBean.getRemoteBusinessMethodSignatures());
```
mkouba: `getLocalBusinessMethodSignatures()` returns an unmodifiable set...

mkouba:
Merged, thanks!

PR #1199, opened 2015-11-06
Add sitemap for html documentation.

tremes:


docs/reference/src/main/sitemap/sitemap.xml
```diff
@@ -0,0 +1,143 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+    <url>
+        <loc>http://docs.jboss.org/weld/reference/latest-master/en-US/html/part-1.html</loc>
```
mkouba: We should probably make the url base, i.e. `http://docs.jboss.org/weld/reference/latest-master/en-US/html`, configurable. A property in docs/reference/pom.xml should work. What do you think?

tremes: Yeah I thought about it shortly. I am not sure if it's doable only with some maven plugin but will check.

mkouba:
Merged, thanks!

PR #1200, opened 2015-11-09
WELD-2056 Proxy class initializer - add ACC_STATIC flag

mkouba:
- the ACC_STATIC flag should only be required for class files with
version number 51.0 or above and weld proxies are 50.0 (java6)
- see also DELTASPIKE-1010 adn https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-2.html#jvms-2.9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: java
 jdk:
+  - oraclejdk8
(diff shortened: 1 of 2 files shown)


mkouba:
Merged.

PR #1206, opened 2015-11-12
WELD-2061 Probe should be part of the OSGi bundle

mkouba:


tremes:
I am not really sure how I can properly verify our OSGI bundle. Maybe we should think of some reasonable way of verification.

mkouba:
@tremes Running the Weld Paint Example and PAX CDI test suite should be enough.

tremes:
I am not aware of any such testsuite and I think running our example doesn't mean lot wrt verification of bundle itself.

mkouba:
Merged.

PR #1211, opened 2015-11-15
Rolling Upgrade session state fix: https://issues.jboss.org/browse/WELD-2064

lprimak:
This patch enables a suffixed version to be ignored by Weld and thus able to upgrade applications with different bean archive names without session or bean loss in a cluster

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/BeanDeployment.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/BeanDeployment.java
@@ -104,7 +104,8 @@ public class BeanDeployment {
         services.add(WSApiAbstraction.class, new WSApiAbstraction(resourceLoader));
         services.add(InterceptorsApiAbstraction.class, new InterceptorsApiAbstraction(resourceLoader));
         services.add(AnnotationApiAbstraction.class, new AnnotationApiAbstraction(resourceLoader));
-        this.beanManager = BeanManagerImpl.newManager(deploymentManager, beanDeploymentArchive.getId(), services);
+        this.beanManager = BeanManagerImpl.newManager(deploymentManager,
+                WeldBootstrap.stripArchiveIdSuffix(beanDeploymentArchive.getId()), services);
         services.add(InjectionTargetService.class, new InjectionTargetService(beanManager));
 
         services.get(WeldModules.class).postBeanArchiveServiceRegistration(services, beanManager, beanDeploymentArchive);
--- a/impl/src/main/java/org/jboss/weld/bootstrap/WeldBootstrap.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/WeldBootstrap.java
@@ -66,7 +66,7 @@ public class WeldBootstrap implements CDI11Bootstrap {
 
     @Override
     public synchronized Bootstrap startContainer(String contextId, Environment environment, Deployment deployment) {
-        weldRuntime = weldStartup.startContainer(contextId, environment, deployment);
+        weldRuntime = weldStartup.startContainer(stripArchiveIdSuffix(contextId), environment, deployment);
         return this;
     }
 
@@ -150,4 +150,12 @@ public class WeldBootstrap implements CDI11Bootstrap {
             throw BootstrapLogger.LOG.callingBootstrapMethodAfterContainerHasBeenInitialized();
         }
     }
+
+    static String stripArchiveIdSuffix(String archiveId) {
+        int idx = archiveId.indexOf(ARCHIVE_SUFFIX_DELIMITER);
+        return archiveId.substring(0, idx >= 0 ? idx : archiveId.length());
+    }
+
+
+    private static final String ARCHIVE_SUFFIX_DELIMITER = "..";
 }


mkouba:
ok to test

mkouba:
In order to make this work we would have to force the users to use the fixed suffix delimiter in the name of the deployment - not very user-friendly. Also this could break existing applications having this suffix in its name already. And what about EARs? 

Also note that `BeanDeploymentArchive.getId()` does not necessarily have to correspond to the real name of the deployment (it's up to the Weld integrator).

The session state loss is also questionable. If you change a bean class you wouldn't be able to deserialize a bean instance anyway.

In any case, "rolling upgrades" is an interesting use-case and we should probably address this requirement.

lprimak:
Hi Martin,

I tried to make the patch as least intrusive as possible with the least impact or sideffects.
The particular delimiter is the easiest way and with the least impact I could think of.

It will not break existing applications.  Even if an existing application uses the suffix, the only change is that the suffix is ignored, which won't break the application itself.  I have tested this.
EARs work as designed as well.  I have tested EAR files as well.

As far as BeanDeploymentArchive.getId() - it doesn't need to correspond to the name of deployment, it just has to be the same across versioned deployments, which is what this patch achieves.
Even if it's winds up not being unique due to stripping of the suffix, it will works as designed. I have tested this as well.

My #1 priority for this patch was to maintain stability and not to break existing things, and I have tested many, many corner cases.

As far as the session state loss, yes, if you change the serialized bean's data in an incompatible way, across versions, it will not be deserialized.  But this is but a very, very small case of application upgrades.  For most stable, production applications, serialized beans (at least serialized state data) change very little, so even if the functionality of the serialized class changes, unless the data itself changes in an incompatible way, the data will be deserialized properly (99% of the upgrade cases).

This makes this functionality very valuable indeed.

----
FTY, I have tested / debugged to come up with this fix for 3 weeks, >10hrs per day, I worked very hard on this, and want to contribute it back to the community.  
I know it's not the first instinct of project maintainers to accept patches from outside committers without prior experience with your projects.
So, I just want to let you know that I already have some credibility submitting patches into JavaEE ecosystem, including GlassFish, Payara, Jersey, and Mojarra.
Hopefully, this will alleviate some concerns about the quality of my work.
Here are some examples (all accepted patches):

https://github.com/payara/Payara/issues/455
https://java.net/jira/browse/JERSEY-2626
https://java.net/jira/browse/JAVASERVERFACES-3673
https://java.net/jira/browse/JAVASERVERFACES-3621

As you can see, most of these issues are leading up to successful rolling application upgrades,
this Weld issue is the last one, as my patched Payara version successfully works with rolling applications upgrades.

Thank you for your consideration

mkouba:
@lprimak I don't have any concerns about the quality of your work. I just want to be sure it does not break anything. It's good to hear you've tested this thoroughly. On the other hand, Weld is not only used in Payara/GlassFish and so we must review the patch carefully.

lprimak:
Understandable.  Let me know if there is anything I can do to get this patch incorporated.

mkouba:
@lprimak If I read https://github.com/payara/Payara/issues/455 correcly, you're not using the GlassFish Application Versions feature, right? Could you describe your rolling upgrades in more detail? I wonder whether it's possible to use similar technique outside of GlassFish/Payara world.

Also if we decide to merge this patch we should make the delimiter configurable (and leave the bean archive id as it is by default) - don't worry about the implementation right now.

lprimak:
> If I read payara/Payara#455 correctly, you're not using the GlassFish Application Versions feature, right?

Correct.  Unfortunately, GlassFish versioning feature is fundamentally broken and not maintained anymore.  For example, any features >JavaEE 6 don't work, like ExecutorService, Batch processing, etc.  This makes GF application versioning unusable.

> I wonder whether it's possible to use similar technique outside of GlassFish/Payara world.

Absolutely.  Payara has built-in Hazelcast support, which decouples session replication from the GlassFish cluster.  I am sure same functionality can be implemented within JBoss / JGroups.
The reason GlassFish clusters don't work in this scenario is because application versioning is broken.
All that needs to happen is individual deploy of clustered apps on individual cluster nodes.

> Could you describe your rolling upgrades in more detail?

We have a modern JavaEE 7/JSF/PrimeFaces web application that is fluid, i.e. needs to be updated multiple times a day.  We don't want all users to get logged out or get downtime when the application gets updated.  We also don't want to have blue/green deploy, because we run many, many VMs don't want to double our resources just for blue/green deploy.  
Solution?  Rolling upgrades.  Here is the workflow:
- Remove node from load balancer (disable application server HTTP/AJP listener)
- un-deploy the existing application
- deploy the new version of the application
- re-enable the HTTP/AJP listener (add it back to the load balancer)
- wait for the LB uptake
- repeat on all other nodes

We can do the upgrade on one node, or do it on all of them, depending of the risk of the upgrade.
The key is that the session state (and all the EJB/CDI beans) get preserved, so the users don't get logged out nor lose their @ViewScoped beans so they continue to interact with the app via Ajax like nothing ever happened.
Same is true if the node fails.  Everything silently fails over to the next available node.

> Also if we decide to merge this patch we should make the delimiter configurable

Sounds good.  Delimited could be configured via system property.

> (and leave the bean archive id as it is by default) - don't worry about the implementation right now.

I would be really, really careful here.  This point is exactly where GlassFish application versioning implementation faltered.  I don't want Weld to go on the same wrong path.  As far as I've seen, there are many, many, places that bdaId is used.  All those places would potentially be replaced with the "stripped" Id.  As time goes by, someone unaware of version stripping would inevitably use the wrong ID and the functionality would be broken instantly.  I think it would be much, much easier to keep bdaId the stripped version, as there are no sideffects, and a new "stripped" ID data point would just needlessly add complexity and potential for errors.
Also, please keep in mind that this functionality doesn't yield easily to automated testing, so that adds significant risk that if a more complex (and error-prone) path is chosen for implementation, since it won't be properly tested, it will degrade and break over time.
My proposed implementation is brain-dead simple and should stand the test of time, even if someone in the future, unaware of version stripping, will make modifications.

mkouba:
> Solution? Rolling upgrades. Here is the workflow:

Well, if you do undeploy/deploy then why not use the same name for the deployment?

lprimak:
> Well, if you do undeploy/deploy then why not use the same name for the deployment?

I wish it was that simple.  That's the first thing I tried to get working.
GlassFish has a concept of a domain, which is a central controller that you can use to control remote nodes.  These remote nodes can be, but are not necessarily form a cluster.  Unfortunately, any applications deployed to the domain have to have unique names.

If I were to undeploy the app (and then redeploy it under the same name), it would first be removed from all the remote nodes, thus defeating the purpose of rolling upgrades.

The undeploy/deploy in the above workflow is on an individual remote node, and not on the domain controller.  These "undeploy/deploy" actually means "disable/enable this app on this remote node"
which actually creates a "reference" of the application on the node by the domain controller.

I think JBoss has something similar also.

lprimak:
Any updates on this?  I would love to get this incorporated into Weld 3

mkouba:
See also https://issues.jboss.org/browse/WELD-2064?focusedCommentId=13139540&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13139540

lprimak:
New branch ( I had to re-fork)
https://github.com/flowlogix/weld-core/commits/rolling-upgrade-patch-2.2.16

PR #1223, opened 2015-11-25
WELD-2019 Ignore synthetic methods

mkouba:
- add jdk8 profile to weld-core-test-arquillian 
- cleanup arquillian profiles in weld-core-test-arquillian


mkouba:
We need to upgrade our CI job - EAP 6.3.x does not support JDK8 lambdas properly.

mkouba:
Merged.

PR #1225, opened 2015-11-26
WELD-2070 introduce JandexServletContextBeanArchiveHandler.

tremes:


environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/deployment/JandexServletContextBeanArchiveHandler.java
```diff
@@ -0,0 +1,77 @@
...
+    public BeanArchiveBuilder handle(String path) {
+
+        BeanArchiveBuilder builder = super.handle(path);
+        if (builder == null) {
+            return null;
+        }
+        for (String className : builder.getClasses()) {
+            // construct path again - add WEB-INF/classes prefix, replace dots and add class suffix
+            addToIndex(servletContext
+                    .getResourceAsStream(WebAppBeanArchiveScanner.WEB_INF_CLASSES_FILE_PATH + File.separator + className.replace(DOT, SLASH) + Files.CLASS_FILE_EXTENSION));
```
tremes: This is not very nice but I think it's better than to mostly copy  method ServletContextBeanArchiveHandler#handleResourcePath from the ancestor.

mkouba: I think a better approach would be to add new protected method to the `ServletContextBeanArchiveHandler`, e.g. `void add(String path)` and put the "adding" logic there. `JandexServletContextBeanArchiveHandler` could then override this method. See also `JandexFileSystemBeanArchiveHandler.add(Entry, BeanArchiveBuilder)`. What do you think?

tremes: Yes sounds like an good idea.Thank you.

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/jandex/Jandex.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @author Tomas Remes
+ */
+public class Jandex {
+
+    public static final String INDEX_ATTRIBUTE_NAME = JandexDiscoveryStrategy.class.getPackage().getName() + ".index";
+
+    private static final String JANDEX_INDEX_CLASS_NAME = "org.jboss.jandex.Index";
+    public static final String JANDEX_DISCOVERY_STRATEGY_CLASS_NAME = "org.jboss.weld.environment.deployment.discovery.jandex.JandexDiscoveryStrategy";
+    public static final String JANDEX_SERVLET_CONTEXT_BEAN_ARCHIVE_HANDLER = "org.jboss.weld.environment.servlet.deployment.JandexServletContextBeanArchiveHandler";
```
mkouba: I'm not so sure it makes sense to put a Weld Servlet specific class name here...

environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/WeldServletLifecycle.java
```diff
@@ -289,8 +294,17 @@ protected CDI11Deployment createDeployment(ServletContext context, CDI11Bootstra
...
 
         DiscoveryStrategy strategy = DiscoveryStrategyFactory.create(resourceLoader, bootstrap, typeDiscoveryConfiguration.getKnownBeanDefiningAnnotations());
-        strategy.registerHandler(new ServletContextBeanArchiveHandler(context));
-
+        if (Jandex.isJandexAvailable(resourceLoader)) {
+            try {
+                Class<? extends BeanArchiveHandler> handlerClass = Reflections.loadClass(resourceLoader, JANDEX_SERVLET_CONTEXT_BEAN_ARCHIVE_HANDLER);
+                strategy.registerHandler((SecurityActions.newConstructorInstance(handlerClass, new Class<?>[] { ServletContext.class }, context)));
+            } catch (Exception e) {
+                throw CommonLogger.LOG.unableToInstantiate(JandexServletContextBeanArchiveHandler.class,
```
mkouba: Use {{JANDEX_SERVLET_CONTEXT_BEAN_ARCHIVE_HANDLER}} instead of class.

mkouba:
Merged, thanks!

PR #1241, opened 2015-12-07
WELD-1811 add warning when decorator declares inappropriate constructor.

tremes:


mkouba:
This will only work if `FastProcessAnnotatedTypeResolver` is used, i.e. if an integrator provides `ClassFileServices` to expose its bytecode scanning facility.

tremes:
Ah I see. I will try to update it.

mkouba:
Merged, thanks!

PR #1252, opened 2016-01-05
WELD-2084 ProxyFactory - change the way bean class methods are added

mkouba:
- if the bean class is abstract, then add methods from all interfaces
implemented by any abstract class from the hierarchy


mkouba:
Merged.

PR #1256, opened 2016-01-07
WELD-2088 Replace NPE with more meaningful error description

mkouba:


manovotn:
+1 for the exception description in the PR.
Tested with latest WFLY and it is now obvious what is going on and why :)

Just a side note - for the sake of clarity, shouldn't the test belong to a package named accordingly to issue? E.g. Weld2088 rather than Weld1177?

mkouba:
> E.g. Weld2088 rather than Weld1177?
Well, imho it's a bad practice to have an issue number in the name of a testcase. I just didn't wanted to rename the test and I also wanted to avoid code duplication...

mkouba:
Merged.

PR #1265, opened 2016-01-13
WELD-2079 Do not register ScheduledExecutorServiceFactory by default

mkouba:
Travis CI build will fail because Weld API 2.3.SP1 is not in Maven central repo yet.


mkouba:
Merged.

PR #1268, opened 2016-01-16
Serialization exception causes fatal error

lprimak:
Basically, readResolve() calls a method that fails if an incompatible application saved serialized CDI objects.  These failures cause all subsequent application requests to fail for the same session.
This fix properly handles exceptions and throws InvalidObjectException instead, which causes the session to be expired and the user can log in again, and not have this issue

Here is the thread for this issue:
https://developer.jboss.org/thread/266577?start=0&tstart=0


mkouba:
Merged, thanks!

PR #1279, opened 2016-02-01
WELD-2100 Correct documentation, section 11.2 was outdated.

manovotn:


docs/reference/src/main/asciidoc/events.asciidoc
```diff
@@ -61,9 +61,16 @@ public void afterDocumentUpdate(@Observes @Updated Document document) { ... }
 -----------------------------------------------------------------------------
 
 An observer method need not specify any event qualifiers—in this case it
-is interested in _only unqualified_ events of a particular type. If it
-does specify qualifiers, it's only interested in events which have those
-qualifiers.
+is interested in every event fired. Such observer will trigger on both events shown below:
```
mkouba: I would also mention the resolution rules here, e.g."...is interested in every event **whose type is assignable to the observed event type**."

mkouba:
Merged, thanks!

PR #1283, opened 2016-02-03
Weld-2103, document configuration option for proxying Java types with…

manovotn:
… final methods.

I was also thinking of adding a note or something saying this should only be used in corner cases, WDYT?


mkouba:
Merged, thanks!

PR #1285, opened 2016-02-08
WELD-2102 Weld SE - support synthetic container lifecycle event observers

mkouba:


environments/se/core/src/main/java/org/jboss/weld/environment/se/ContainerLifecycleObserver.java
```diff
@@ -0,0 +1,673 @@
...
+        }
+
+    }
+
+    /**
+     * A synthetic extension, i.e. a collection of synthetic container lifecycle observers.
+     *
+     * @author Martin Kouba
+     */
+    public static class ContainerLifecycleObserverExtention implements SyntheticExtension {
```
manovotn: I think there is a typo here, 'ContainerLifecycleObserver**Extention**' should be 'ContainerLifecycleObserver**Extension**'.

mkouba: You're right, thanks!

mkouba:
Merged.

PR #1289, opened 2016-02-11
Fixed Probe test failures. Refactored to leverage some of Graphene ad…

manovotn:
…vantages such as @Page.


mkouba:
Merged, thanks!

PR #1293, opened 2016-02-29
WELD-2113 introduce configuration properties for conversation timeout

tremes:
and conversation conccurent access timeout.


manovotn:
We might also want to update doc section 5.3.3. Conversation timeout and mention that the timeout is configurable. 
Furthermore somewhere in the doc we should say that this timeout configuration is only a 'hint' for the container as the conversation might get destroyed earlier due to HTTP session invalidation or HTTP session timeout.

mkouba:
@manovotn Good points. Otherwise looks good.

tremes:
Yes I am going to expand `5.3.3. Conversation timeout`. Wrt the second point - we have following (which seems to me sufficient) in `22.1. Managing the built in contexts`:
>The bound session context works in much the same way, excepting that invalidating and deactivating the session context causes the any conversations in the session to be destroyed as well.

Or do you guys have any other suggestions how to improve doc?

mkouba:
@tremes @manovotn I think the regular users usually don't read `Managing the built in contexts` section. I would add this info to `5.3.3. Conversation timeout`.

manovotn:
@mkouba +1

mkouba:
Merged, thanks!

PR #1294, opened 2016-02-29
WELD-2113 introduce configuration properties for conversation timeout

tremes:
and conversation conccurent access timeout.


impl/src/main/java/org/jboss/weld/context/AbstractConversationContext.java
```diff
@@ -93,8 +93,9 @@
     public AbstractConversationContext(String contextId, BeanIdentifierIndex beanIdentifierIndex) {
```
mkouba: I would rather change the constructor to accept `ServiceRegistry`...

mkouba:
Merged, thanks!

PR #1295, opened 2016-03-04
WELD-2115, refactor BOM.

manovotn:
Re-worked BOM - it now only contains ```dependencyManagement``` section and no direct dependencies.
Also it has no parent in order to avoid fetching unnecessary stuff (depending on weld-core-parent brought in loads of other things).

@tremes WDYT? Does this solution make sense? Or should we do it some other way, e.g. create common parent  somewhere?

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -5,13 +5,6 @@
     <packaging>pom</packaging>
     <version>2.3.4-SNAPSHOT</version>
 
-    <parent>
-        <groupId>org.jboss.weld</groupId>
-        <artifactId>weld-core-parent</artifactId>
-        <version>2.3.4-SNAPSHOT</version>
-        <relativePath>../pom.xml</relativePath>
-    </parent>
-
     <name>Weld Core BOM</name>
 
     <!-- Minimal project metadata, for more see parent/pom.xml -->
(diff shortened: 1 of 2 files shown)


tremes:
@manovotn This could make sense, but I think we don't want to manage versions at the second additional place therefore we will probably need parent structure. Second question is inclusion of api dependency in this bom. I think it's not neccessary since we still have weld-api-bom but it's questionable.

joergsesterhenn:
There is no need to manage versions in two places. Your parent pom could/should import the BOM as well.

manovotn:
Now I see I made a mistake there, there is no need to have ```weld.core.version``` property, this can be achieved using ```${project.version}``` which has to be updated anyway. 

Back to your comment, Tomas I though about both problems..
1) Version management
This BOM only has one version - API, which is not much of a bother I think since API does not change that often. Not much of a bother but I see the point.
2) API in core BOM
I think it should be there, as the API (and SPI) version differs from core ones. This is what BOM should do - allow to easily align fitting versions.

manovotn:
Thanks for comment joergsesterhenn, it crossed my mind as well. Although we manage all versions in ```weld-core-parent``` and using this solution would mean that we would split version management into two places anyway (API in BOM, the rest in parent pom).

manovotn:
@mkouba I modified `weld-core-parent` to include a `dependencyManagement` on `weld-api-parent`. (same change we applied to master)

mkouba:
Merged, thanks!

PR #1300, opened 2016-03-10
WELD-2064 with configurable parameter and an automated test

manovotn:
Built on top of Lenny's original commit.

Introduces a configurable property allowing to set a delimiter for a bean archive ID.
I am still trying to figure out the bootstrapping part in WeldStartup. 
Also there is an automated test added which I believe could do the right thing.

In the current form, the tests-arquillian as well as unit tests in core impl pass, though I am unsure it is a correct solution.

Furthermore the names of the properties and the helper class are a subject for discussion :)

### DIFF
--- a/docs/reference/src/main/asciidoc/configure.asciidoc
+++ b/docs/reference/src/main/asciidoc/configure.asciidoc
@@ -187,6 +187,21 @@ This optimization is used to reduce the HTTP session replication overhead. Howev
 
 NOTE: This optimization is disabled by default in <<weld-servlet,Servlet containers>>.
 
+==== Rolling upgrades ID delimiter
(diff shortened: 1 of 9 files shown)


manovotn:
Added a doc section to 'Weld Configuration' .

manovotn:
Removed ```Container.currentId``` as it was retaining incorrect version of archive ID and it was never used.

lprimak:
This is looking great!  I am wondering when this will be merged,
I would love this to be back-ported to Weld 2.3 as well
thanks!

mkouba:
Closed in favor of https://github.com/weld/core/pull/1366 (all commits included)

PR #1301, opened 2016-03-11
WELD-1905 introduce interceptor activating @ThreadScope

tremes:


tremes:
I am not sure whether adding ThreadContext as a bean is the best solution.

mkouba:
You don't need this - just inject `WeldSEBeanRegistrant` and use `getThreadContext()`.

mkouba:
Merged, thanks!

PR #1304, opened 2016-03-23
Duplicate statement in WoodlandExtension.java

kilida:


tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/WoodlandExtension.java
```diff
@@ -77,7 +77,6 @@ public static void reset() {
         injectCalled = false;
         postConstructCalled = false;
         preDestroyCalled = false;
-        preDestroyCalled = false;
```
mkouba: I believe it should be replaced with `produceCalled = false`. Could you modify your pull request? Thanks for noticing!

kilida:
Patch updated to reflect @mkouba's comment

mkouba:
Merged, thanks!

PR #1306, opened 2016-03-31
WELD-2123 use same logic to obtain BeanManager as in WeldInitialListener

tremes:


impl/src/main/java/org/jboss/weld/servlet/WeldTerminalListener.java
```diff
@@ -61,6 +70,23 @@ public void sessionCreated(HttpSessionEvent httpSessionEvent) {
 
     @Override
     public void sessionDestroyed(HttpSessionEvent event) {
+        final ServletContext ctx = event.getSession().getServletContext();
+        // First try to use the context id obtained from the servlet context (OSGi, Servlet containers, etc.)
+        synchronized (this) {
```
mkouba: You should also check the value ahead of the synchronized block. It's also good to minize the access to the volatile variable. Although the performance does not really matter here. See also variants of double-checked locking: https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java.

mkouba:
Merged, thanks!

PR #1313, opened 2016-04-04
WELD-2122 Support libs on Tomcat with unpackWARs=false

mkouba:


manovotn:
Tested manually with Tomcat 8 using numberguess example. I added a depedency on a cdi-enabled jar which contained an app scoped bean.
Worked like a charm.

mkouba:
Merged.

PR #1315, opened 2016-04-05
WELD-2120 don't notify AFTER_SUCCES transactional observer when

tremes:
transaction was rollbacked

### DIFF
--- a/impl/src/main/java/org/jboss/weld/event/ResolvedObservers.java
+++ b/impl/src/main/java/org/jboss/weld/event/ResolvedObservers.java
@@ -16,8 +16,7 @@
  */
 package org.jboss.weld.event;
 
-import static com.google.common.collect.ImmutableSet.copyOf;
-
+import static com.google.common.collect.ImmutableList.copyOf;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -27,9 +26,8 @@ import javax.enterprise.event.TransactionPhase;
 import javax.enterprise.inject.spi.EventMetadata;
 import javax.enterprise.inject.spi.ObserverMethod;
 
-import org.jboss.weld.util.Observers;
-
 import com.google.common.collect.ImmutableSet;
+import org.jboss.weld.util.Observers;
 
 /**
  * Immutable information about observer methods resolved for a type/qualifiers combination.
@@ -40,7 +38,7 @@ import com.google.common.collect.ImmutableSet;
  */
 public class ResolvedObservers<T> {
 
-    private static final ResolvedObservers<Object> EMPTY = new ResolvedObservers<Object>(Collections.<ObserverMethod<? super Object>>emptySet(), Collections.<ObserverMethod<? super Object>>emptySet(), false) {
+    private static final ResolvedObservers<Object> EMPTY = new ResolvedObservers<Object>(Collections.<ObserverMethod<? super Object>>emptyList(), Collections.<ObserverMethod<? super Object>>emptyList(), false) {
         public boolean isEmpty() {
             return true;
         }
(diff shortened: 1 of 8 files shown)


impl/src/main/java/org/jboss/weld/event/TransactionalObserverNotifier.java
```diff
@@ -73,8 +78,32 @@ protected TransactionalObserverNotifier(String contextId, TypeSafeObserverResolv
...
-                deferNotification(event, metadata, observer);
+            try {
+                for (ObserverMethod<? super T> observer : observers) {
+                    deferNotification(event, metadata, observer);
+                }
+            } catch (Exception e) {
+                if (e.getCause() instanceof RollbackException || e.getCause() instanceof IllegalStateException) {
+
+                    List<ObserverMethod<? super T>> filteredObservers = new ArrayList<ObserverMethod<? super T>>();
+                    filteredObservers.addAll(observers);
```
mkouba: Why not `new ArrayList<>(observers)`?

impl/src/main/java/org/jboss/weld/event/TransactionalObserverNotifier.java
```diff
@@ -73,8 +78,32 @@ protected TransactionalObserverNotifier(String contextId, TypeSafeObserverResolv
...
+                    for (ObserverMethod<? super T> observer : observers) {
```
mkouba: I think a better approach would be to iterate over `filteredObservers` and use iterator to remove AFTER_SUCCESS observers.

impl/src/main/java/org/jboss/weld/event/TransactionalObserverNotifier.java
```diff
@@ -73,8 +78,32 @@ protected TransactionalObserverNotifier(String contextId, TypeSafeObserverResolv
...
+                        if (observer.getTransactionPhase().equals(TransactionPhase.AFTER_SUCCESS)) {
+                            filteredObservers.remove(observer);
+                        }
+                    }
+
+                    Collections.sort(filteredObservers, new Comparator<ObserverMethod<? super T>>() {
+                        @Override
+                        public int compare(ObserverMethod<? super T> o1, ObserverMethod<? super T> o2) {
+                            return o1.getTransactionPhase().compareTo(o2.getTransactionPhase());
```
mkouba: This will compare enums by ordinals. Although it's unlikely that ordering of `TransactionPhase` values will change it might be better to compare by values toString()...?

impl/src/main/java/org/jboss/weld/event/TransactionalObserverNotifier.java
```diff
@@ -73,8 +78,32 @@ protected TransactionalObserverNotifier(String contextId, TypeSafeObserverResolv
...
+                        }
+                    });
+                    notifySyncObservers(ImmutableSet.copyOf(filteredObservers), event, metadata);
```
mkouba: `ImmutableSet` is not necessary...

tremes:
This will need retest. The job wasn't able to fetch the git repository.

mkouba:
retest this please

impl/src/main/java/org/jboss/weld/event/ResolvedObservers.java
```diff
@@ -64,24 +61,24 @@ public boolean isEmpty() {
                 metadataRequired = true;
             }
         }
-        return new ResolvedObservers<T>(copyOf(immediateObservers), copyOf(transactionObservers), metadataRequired);
+        return new ResolvedObservers<T>(immediateObservers, transactionObservers, metadataRequired);
```
mkouba: I would use `ImmutableList.copyOf()` here...

tremes:
retest this please

mkouba:
retest this please

impl/src/main/java/org/jboss/weld/event/TransactionalObserverNotifier.java
```diff
@@ -73,8 +77,33 @@ protected TransactionalObserverNotifier(String contextId, TypeSafeObserverResolv
             // Transaction is not active - no deferred notifications
             notifySyncObservers(observers, event, metadata);
         } else {
-            for (ObserverMethod<? super T> observer : observers) {
-                deferNotification(event, metadata, observer);
+            try {
+                for (ObserverMethod<? super T> observer : observers) {
+                    deferNotification(event, metadata, observer);
```
mkouba: Theoretically, it's possible that some registrations pass. So in the catch block we should only process the observers for which the registration failed. E.g. create a copy of observers outside the catch block, remove each observer for which the registration passed and then process the copy in the catch block.

mkouba:
Merged, thanks!

PR #1318, opened 2016-04-12
WELD-2135 Weld SE builder configuration

mkouba:
- allow to configure isolation, development mode and registration of
shutdown hook manually


mkouba:
Merged.

PR #1320, opened 2016-04-12
WELD-2134 log actions from extensions.

tremes:


impl/src/main/java/org/jboss/weld/bootstrap/events/ContainerEvent.java
```diff
@@ -34,7 +34,7 @@
      * for implementing ProcessSyntheticAnnotatedType properly. The information must be set by an
      * ObserverMethod implementation before invoking the target observer method and unset once the notification is complete.
      */
-    private Extension receiver;
+    protected Extension receiver;
```
mkouba: Why not use `getReceiver()`?

impl/src/main/java/org/jboss/weld/logging/BootstrapLogger.java
```diff
@@ -200,4 +202,29 @@
     @LogMessage(level = Level.WARN)
     @Message(id = 147, value = "Decorator {0} declares inappropriate constructor therefore will not available as a managed bean!", format = Format.MESSAGE_FORMAT)
     void decoratorWithNonCdiConstructor(String clazzName);
+
+    @LogMessage(level = Logger.Level.DEBUG)
+    @Message(id = 148, value = "Extension {0} sets annotatedType for {1}", format = Format.MESSAGE_FORMAT)
```
mkouba: What about the following template: `ProcessAnnotatedType.setAnnotatedType() called by extension {0} for {1}`?

impl/src/main/java/org/jboss/weld/bootstrap/events/AfterTypeDiscoveryImpl.java
```diff
@@ -76,6 +77,7 @@ protected AfterTypeDiscoveryImpl(BeanManagerImpl beanManager, BeanDeploymentArch
     public void addAnnotatedType(AnnotatedType<?> type, String id) {
         checkWithinObserverNotification();
         addSyntheticAnnotatedType(type, id);
+        BootstrapLogger.LOG.addAnnotatedTypeCalled(receiver.getClass().getName(), type.getJavaClass());
```
mkouba: I think `getReceiver().getClass()` should be enough. And we could consider `type.toString()` instead of `type.getJavaClass()`?

impl/src/main/java/org/jboss/weld/bootstrap/events/ProcessSimpleInjectionTarget.java
```diff
@@ -45,6 +45,7 @@ public ProcessSimpleInjectionTarget(BeanManagerImpl beanManager, AnnotatedType<X
     public void setInjectionTarget(InjectionTarget<X> injectionTarget) {
         checkWithinObserverNotification();
         this.injectionTarget = injectionTarget;
+        BootstrapLogger.LOG.setInjectionTargetCalled(getReceiver().getClass(), injectionTarget.getClass());
```
mkouba: I think `injectionTarget.getClass()` does not make sense here. It will be better to simply pass the IT instance. And the same probably applies to extension. Also I think it would be better to log both the original and the new value for all `ProcessX` events, e.g. something like:
`ProcessInjectionTarget.setInjectionTarget() called by {0}: {1} replaced by {2}`. What do you think?

tremes: What's the difference between passing instance and instance.getClass? It will be bit verbose but yes we can log both.

mkouba: `toString()` may be overriden showing some important info. Otherwise `Object.toString()` always contains the class name.

tremes: Ok if we pass only `getReceiver()` for extension param then we wiill display something like `com.sun.faces.application.view.ViewScopeExtension@69309707` that's not so nice. Same for annotatedType etc.

mkouba: I find it quite nice :-). One can see the class name and the hashcode of the instance. Don't forget it's only a debug message, it does not have to be nice.

mkouba:
Looks good. I'm going to merge this pull request tomorrow.

mkouba:
Merged, thanks!

PR #1328, opened 2016-04-18
WELD-2110 Relax Event.select() validation

mkouba:
- the event type may contain wildcards


tests-arquillian/src/test/java/org/jboss/weld/tests/event/subtype/Observers.java
```diff
@@ -17,17 +34,26 @@ void observeBar(@Observes Bar bar) {
         this.bar = bar;
     }
 
-    public Bar getBar() {
+    void observeUpdatedList(@Observes Baz<?> baz) {
```
manovotn: Probably forgot to rename this method to something like `observeBaz`?

mkouba:
Merged.

PR #1331, opened 2016-04-20
Test dependendy update for Arquillian and htmlUnit.

manovotn:
This should resolve the `ftest` problems we came across lately.


mkouba:
Merged, thanks!

PR #1336, opened 2016-04-22
WELD-2125 expose interceptorBuilder in Weld builder.

tremes:


impl/src/main/java/org/jboss/weld/bootstrap/events/InterceptorBuilderImpl.java
```diff
@@ -49,6 +50,12 @@
 
     private BeanManagerImpl beanManager;
 
+    private BeanDeploymentFinder beanDeploymentFinder;
+
+    public InterceptorBuilderImpl() {
+        this(null);
```
tremes: I personally don't like this much but didn't invent anything better in this case.

mkouba:
Merged, thanks!

PR #1342, opened 2016-04-29
WELD-2144 add dependency to weld-core-jsf

tremes:


mkouba:
Maybe we should also revise `org.jboss.weld.servlet:weld-servlet` pom.xml... weld-core-impl and weld-core-jsf seem to be redundant (but rather double check this assumption ;-).

tremes:
Yes I realized that. Both should be likely redundant.

tremes:
Hm weird SE tests failure at JDK7....

mkouba:
Merged, thanks!

PR #1357, opened 2016-05-17
QualifierInstance - improve failure description

mkouba:


caarlos0:
Nice!

Any tips on how I can add this to an existing wildfly distrubution?

manovotn:
@caarlos0 It is about replacing jars only.
Under Wildfly folder go to `modules/system/layers/base/org/jboss/weld`. There you will find all jars with Weld. Then:
* Build Weld with the PR
* Replace the affected jar (`core` I believe)
* If the newly created jar has a different name, reflect this name under `module.xml`

That should do the trick :)

mkouba:
For WildFly 10 I'd rather recommend to use the update script, i.e.:
````
git clone git@github.com:weld/core.git
cd core
mvn clean install -DskipTests
export JBOSS_HOME="/path/to/wildfly10"
mvn clean package -f jboss-as/pom.xml -Pupdate-jboss-as
````
@caarlos0 I know your're trying to migrate to WildFly 9. But you could go with WildFly 10 for the testing purpose.

caarlos0:
@mkouba my app is java 7, will not work on wildfly 10 :/

caarlos0:
@manovotn yeah, seems like that will work, will try that! Thanks!

mkouba:
> @mkouba my app is java 7, will not work on wildfly 10 :/
Why not? You should be able to run an app compiled with java 7 on java 8.

caarlos0:
@mkouba real big app, lot's of "weird" and old deps, I doubt it will work... I'll try to update wildfly's 9 weld first...

mkouba:
@caarlos0 Sure, no problem. Let us know.

caarlos0:
@mkouba @manovotn Thanks for your help, I backported this fix to weld 2.2 [#1359] and found out the problem!

mkouba:
Merged.

PR #1364, opened 2016-05-24
WELD-2157 add dependency on jaxws-api to prevent CNFE with Java 9.

manovotn:
Adds a dependency on `jaxws-api` into jboss-tck-runner.
This eliminates the CNFE when running with JDK 9 (EA).

Note: I haven't yet pinpointed why is this happening and why exactly do we need this dependency

The modification in surefire plugin is just from deprecated approach to currently used one (according to docs, the behaviour equals).


manovotn:
We have agreed that this issue should be  resolved on TCK level since the annotations come from there.
Hence I am closing this PR and will open another one against TCK.

PR #1365, opened 2016-05-24
WELD-2155 added explicit dependency on BCEL and modified Formats acco…

manovotn:
…rdingly.

This is the first approach I tried to resolve the problem with inaccessible BCEL classes in JDK 9.
I added explicit dependency and rewrote Formats class to avoid reflections and use the dependency directly.

Feel free to reject it, if you think we shouldn't go this way :)

### DIFF
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -79,6 +79,12 @@
             <version>1.3.2</version>
             <optional>true</optional>
         </dependency>
+        
+        <dependency>
+            <groupId>org.apache.bcel</groupId>
+            <artifactId>bcel</artifactId>
+            <version>5.2</version>
+        </dependency>
 
         <!-- test dependencies -->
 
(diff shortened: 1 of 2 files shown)


mkouba:
I don't think we want to add a hard dependency to `bcel` just because of this "get line number" feature. Also I don't think we should remove the current solution entirely - it works on most JDK8s.

We could extend the current solution (i.e. also try to load `org.apache.bcel` classes via reflection and recommend users to add an optional `bcel` dependency). But that's not very user-friendly.

I think we should explore the JDK9 thoroughly and find out whether there is some way to do this without a new dependency.

manovotn:
> (i.e. also try to load org.apache.bcel classes via reflection...
I don't like that one either.

Imho we should go for something which won't put any additional burden on users, something working by default. So yea, let us explore meanwhile.

manovotn:
Closing PR - we will not go this way ATM. There is a possibility to export these internal classes into a module where Weld can still use them, should the user want/need that.

PR #1370, opened 2016-05-30
WELD-2163 document org.jboss.weld.configuration.spi.ExternalConfigura…

manovotn:
…tion.

### DIFF
--- a/docs/reference/src/main/asciidoc/configure.asciidoc
+++ b/docs/reference/src/main/asciidoc/configure.asciidoc
@@ -220,6 +220,37 @@ to wait on the conversation concurrent lock.
 |`org.jboss.weld.conversation.concurrentAccessTimeout` |1000 ms| Conversation concurrent access timeout represent the maximum time to wait on the conversation concurrent lock.
 |=======================================================================
 
+=== Defining external configuration
+
+Weld allows integrators to provide an external configuration - a class which implements `org.jboss.weld.configuration.spi.ExternalConfiguration` interface. 
+This interface has `getConfigurationProperties` method which returns a `Map` with provided configuration and also inherits a `cleanup` method because it extends `org.jboss.weld.bootstrap.api.Service`. 
+Below is a short example of `ExternalConfiguration` implementation:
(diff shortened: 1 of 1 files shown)


docs/reference/src/main/asciidoc/configure.asciidoc
```diff
@@ -1,6 +1,7 @@
 [[configure]]
 == Configuration
 
+[[weld-configuration]]
```
mkouba: The id of this section is currently autogenerated and has a value of `_weld_configuration`. We should probably keep this value so that for example existing links do not break.

docs/reference/src/main/asciidoc/configure.asciidoc
```diff
@@ -220,6 +221,35 @@ to wait on the conversation concurrent lock.
 |`org.jboss.weld.conversation.concurrentAccessTimeout` |1000 ms| Conversation concurrent access timeout represent the maximum time to wait on the conversation concurrent lock.
 |=======================================================================
 
+=== Defining external configuration
+
+Weld allows integrators to provide an external configuration - a class which implements `org.jboss.weld.configuration.spi.ExternalConfiguration` interface. This interface has `getConfigurationProperties` method which returns a `Map` with provided configuration and also inherits a `cleanup` method because it extends `org.jboss.weld.bootstrap.api.Service`. Below is a short example of `ExternalConfiguration` implementation:
+
```
mkouba: It's a good idea to have one sentence per line... it's more readable ;-)

mkouba:
Otherwise looks good!

docs/reference/src/main/asciidoc/configure.asciidoc
manovotn: Cool, just what I was looking for. I copied this solution from other part of the doc since I wasn't sure how this works. Thanks!

manovotn:
Updated both PRs accordingly, thanks for hints.

mkouba:
Merged, thanks!

PR #1387, opened 2016-06-07
Correct links in Weld doc to point into CDI 1.2 specification

manovotn:
I found out Weld doc has linnks to CDI 1.1  spec instead of CDI 1.2.
Grep showed me only these three occurrences, so here is a PR.


mkouba:
Merged, thanks!

PR #1389, opened 2016-06-07
WELD-2175 throw an exception when disposal method is a business metho…

tremes:
…d exposed by a remote business interface.


mkouba:
So I suppose it's not a problem for producer methods where we already check whether a method is declared on one of the declaring bean types, right?

tremes:
Yes it's not a problem for producers.

mkouba:
Merged, thanks!

PR #1392, opened 2016-06-09
WELD-2178 log warning when AT not resolved to a bean.

tremes:


impl/src/main/java/org/jboss/weld/util/Beans.java
```diff
@@ -583,6 +583,13 @@ public static boolean hasBuiltinScope(Bean<?> bean) {
                 || ConversationScoped.class.equals(bean.getScope()) || Dependent.class.equals(bean.getScope());
     }
 
+    public static boolean hasBuiltinScope(AnnotatedType<?> annotatedType) {
+        return annotatedType.getAnnotations().contains(RequestScoped.Literal.INSTANCE) || annotatedType.getAnnotations()
```
manovotn: How about storing `annotatedType.getAnnotations()` into a variable and write the boolean expression using that? It would make the boolean expression much shorter.

tremes: Yes, could be.

mkouba:
Merged, thanks!

PR #1393, opened 2016-06-09
WELD-2172 Probe - allow to test bean availability in a given bean arc…

mkouba:
…hive

### DIFF
--- a/impl/src/main/java/org/jboss/weld/resolution/QualifierInstance.java
+++ b/impl/src/main/java/org/jboss/weld/resolution/QualifierInstance.java
@@ -101,7 +101,15 @@ public class QualifierInstance {
         this(annotationClass, Collections.<String, Object>emptyMap());
     }
 
-    private QualifierInstance(Class<? extends Annotation> annotationClass, Map<String, Object> values) {
+    /**
+     * Constructs a qualifier instance without any checks. This method should be used with care.
+     *
+     * @param annotationClass
+     * @param values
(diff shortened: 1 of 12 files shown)


manovotn:
I took a glance at this via our Numberguess (WAR) and Translator (EAR) samples.
Few things I noticed:
* "With Qualifiers (optional):"
 * We might want to explicitly state what you get if you do not type in anything (which is @ Default)
 * Also I was thinking whether it wouldn't be better to have @Any as default (as in "give me all beans in archive" -> see below)
* Add shortcut to show all beans with any qualifiers in given archive
 * actually, first thing I tried to do with this
 * you now need to use `java.lang.Object` as type and `javax.enterprise.inject.Any` as qualifier to truly see all beans
 * we might add an option to use `*` as a shortcut for this?
* Add table page controls and total sum of beans to the top of the table
 * to allign it with how we have it in other views (for instance 'Beans')
* Typo in hint for "With Required Type:"
 * withing the example there is `java.lang.Integer>[]`, remove `>`

I'll modify our samples a bit to make it more complex and get back if I find anything else. So far I haven't been able to break it, but I won't give up that easily ;-)

manovotn:
Found one odd thing - when browsing for bean which is not visible in that given archive, you get `Unable to parse the required type: org.jboss.weld.examples.translator.SentenceParser`. Switching the archive to one which has the bean then shows results immediately. This can be misleading, inability to parse is imho not the same thing as no such bean found.

mkouba:
Merged.

PR #1394, opened 2016-06-09
WELD-2178 log warning when AT not resolved to a bean.

tremes:


impl/src/main/java/org/jboss/weld/util/Beans.java
```diff
@@ -647,6 +652,13 @@ public static boolean hasBuiltinScope(Bean<?> bean) {
                 || ConversationScoped.class.equals(bean.getScope()) || Dependent.class.equals(bean.getScope());
     }
 
+    public static boolean hasBuiltinScope(AnnotatedType<?> annotatedType) {
+        Set<Annotation> annotations = annotatedType.getAnnotations();
```
mkouba: I would use `AnnotatedType.isAnnotationPresent()` instead (so that those literals are not needed).

impl/src/main/java/org/jboss/weld/logging/BootstrapLogger.java
```diff
@@ -276,4 +276,8 @@
     @Message(id = 166, value = "AfterTypeDiscovery.{3} modified by {0} {2} {1}", format = Format.MESSAGE_FORMAT)
     void typeModifiedInAfterTypeDiscovery(Object extensionName, Object type, Object operation , Object types);
 
+    @LogMessage(level = Level.WARN)
+    @Message(id = 167, value = "Class {0} cannot be resolved to a bean because it does not declare appropriate constructor!", format = Format.MESSAGE_FORMAT)
+    void annotatedTypeNotResolvedToBean(String clazzName);
```
mkouba: The method name and the message should be more specific. Maybe something like _"Class {0} is annotated with {1} but it does not declare an appropriate constructor therefore is not registered as a bean"_...

impl/src/main/java/org/jboss/weld/util/Beans.java
```diff
@@ -647,6 +648,15 @@ public static boolean hasBuiltinScope(Bean<?> bean) {
                 || ConversationScoped.class.equals(bean.getScope()) || Dependent.class.equals(bean.getScope());
     }
 
+    public static Class<? extends Annotation> getBeanDefiningMetaAnnotation(AnnotatedType<?> annotatedType) {
```
mkouba: I think it should be rather something like `getBeanDefiningAnnotationScope()`. This method does not return a meta annotation (e.g. `NormalScope`) but the scope types which are bean defining annotations. 

Otherwise looks perfect!

mkouba:
Merged, thanks!

PR #1397, opened 2016-06-10
WELD-2172 Probe - allow to test bean availability in a given bean arc…

mkouba:
…hive

- also add Object info to contextual instance detail


mkouba:
Merged.

PR #1400, opened 2016-06-13
WELD-2179 Fix private observer/producer/disposer on bean with decorator

mkouba:


manovotn:
Just not to forget it ...
The failure is caused by usage of `Default.Literal.INSTANCE` in tests  which is not available in CDI 1.2.

mkouba:
Merged.

PR #1410, opened 2016-07-05
WELD-2189 prepare Jigsaw modularisation

antoinesd:
Moving classes from org.jboss.weld.environment to servlet subpackage to avoid package spliting across artefacts


mkouba:
Test this please

PR #1411, opened 2016-07-12
WELD-2155 JDK9: force-add java BCEL internal classes to unnamed module

manovotn:
This is probably the easiest solution to WELD-2155. Yet it might be temporary (in a very long term with regard to Java development speed) as this approach might be deprecated.

The idea is to force-export internal java modules (BCEL classes for our `Formats` class) into unnamed module hence making them accessible for Weld. This is done via a profile which only kicks in when running JDK 9.

### DIFF
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -149,6 +149,26 @@
 
     <profiles>
         <profile>
+            <id>jdk9</id>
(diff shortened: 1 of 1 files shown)


mkouba:
Merged, thanks!

PR #1412, opened 2016-07-19
Weld 2093 intercept default methods

tremes:


tremes:
I think we need tu update WildFly to use jboss-classfilewriter 1.2.0.Beta1

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -493,12 +505,22 @@ protected void addAdditionalInterfaces(Set<Class<?>> interfaces) {
         return proxyClass;
     }
 
-    private static ClassFile newClassFile(String name, int accessFlags, String superclass, String... interfaces) {
-        if (CONFIGURABLE_ACCESS_FLAGS) {
-            return new ClassFile(name, accessFlags, superclass, interfaces);
-        } else {
-            return new ClassFile(name, superclass, interfaces);
+    private ClassFile newClassFile(String name, int accessFlags, String superclass, String... interfaces) {
+        WeldClassLoaderResourceLoader resourceLoader = WeldClassLoaderResourceLoader.INSTANCE;
```
mkouba: We should probably use `this.classLoader`

mkouba:
For now we can mark the test with `@Category(EmbeddedContainer.class)`.

tremes:
Closed in favor of https://github.com/weld/core/pull/1415

PR #1416, opened 2016-07-22
WELD-2197 Replace Guava - basic utilities

mkouba:


impl/src/main/java/org/jboss/weld/util/Defaults.java
```diff
@@ -0,0 +1,67 @@
...
+        // TODO Use the appropriate immutable map replacement (and possibly builder)
+        JLS_PRIMITIVE_DEFAULT_VALUES = ImmutableMap.copyOf(values);
+    }
+
+    private static <T> void put(Map<Class<?>, Object> map, Class<T> type, T value) {
+        map.put(type, value);
+    }
+
+    /**
+     * See also JLS8, 4.12.5 Initial Values of Variables.
```
manovotn: `See also JLS8,` -> `JLS7`
I checked them both and they are identical as far as impl goes, but we should probably avoid mentioning J8.

impl/src/test/java/org/jboss/weld/tests/unit/util/collections/WeldCollectionsTest.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2014, Red Hat, Inc., and individual contributors
```
manovotn: `2014` -> `2016` (5 occurences in total)
Just to keep things up to date :)

impl/src/main/java/org/jboss/weld/util/Defaults.java
mkouba: I don't think it's a problem. I believe it's the same for all java versions.

impl/src/test/java/org/jboss/weld/tests/unit/util/collections/WeldCollectionsTest.java
mkouba: This file was created in 2014 so I think it's ok to leave it as it is ;-)

mkouba:
Merged.

mkouba:
retest this please

PR #1418, opened 2016-07-22
WELD-2196 Replace Guava - caches

mkouba:


impl/src/main/java/org/jboss/weld/util/cache/ComputingCache.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+package org.jboss.weld.util.cache;
+
+import org.jboss.weld.util.Consumer;
+
+/**
+ * A simple abstraction for computing cache.
+ * <p>
+ * Implementations are not required to support recursive computations. A cache client must be aware that such computations may result in livelocks, inifinite
+ * loops and other undesired situations.
```
tremes: Typo in the word "infinite."

impl/src/main/java/org/jboss/weld/util/cache/ComputingCache.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @return an immutable iterable of values
+     */
+    Iterable<V> getAllPresentValues();
+
+    /**
+     * Performs the given action for each cached value.
+     * @param consumer the given action
+     */
+    void forEachValue(Consumer<? super V> consumer);
```
tremes: This method seems to be never used so far..

mkouba: You're right, I will remove this method.

mkouba:
Merged.

PR #1428, opened 2016-07-27
WELD-2200 replace MetaDataImpl.

tremes:


tremes:
retest this please

impl/src/main/java/org/jboss/weld/metadata/MetadataImpl.java
```diff
@@ -1,12 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.jboss.weld.metadata;
 
 import org.jboss.weld.bootstrap.spi.Metadata;
 
+@Deprecated
```
mkouba: Should we add `@deprecated` javadoc tag referencing MetadataImpl from API?

tremes: Could be.

mkouba:
Merged, thanks!

PR #1431, opened 2016-07-29
WELD-2195 [Backport] Replace Guava - new collections types

manovotn:


impl/src/main/java/org/jboss/weld/util/collections/AbstractMultimap.java
```diff
@@ -0,0 +1,208 @@
...
+    public boolean isEmpty() {
+        return map.isEmpty();
+    }
+
+    @Override
+    public C get(K key) {
+        C value = map.get(key);
+        if (value == null) {
+            value = supplier.get();
+            C previousValue = map.put(key, value);
```
mkouba: As previously discussed this is not thread safe. We cannot use `Map.computeIfAbsent()` or `Map.putIfAbsent()` introduced in JDK8 so I propose to add `ConcurrentMap<K, C> concurrentMap` field and corresponding param to the constructor. It would be illegal to provide both suppliers.

The `get()` method could have two branches. One for regular map:

```
if (map.get(key) == null) {
     V newValue = supplier.get(key);
     if (newValue != null)
         map.put(key, newValue);
 }
```
And second branch for `ConcurrentMap` - the same as the current version but using `putIfAbsent()` instead of `put()`.
Of course, we would have to modify `SetMultimap` and `ListMultimap`.

impl/src/main/java/org/jboss/weld/bootstrap/SpecializationAndEnablementRegistry.java
```diff
@@ -88,7 +89,14 @@ private void buildAccessibleBeanDeployerEnvironments(BeanManagerImpl manager, Co
...
                 if (isEnabledInAnyBeanDeployment(specializingBean)) {
-                    specializedBeansSet.addAll(result);
+                    for (AbstractBean<?, ?> specializedBean : result) {
+                        // not present yet
+                        if (specializedBeansMap.get(specializedBean) == null) {
+                            specializedBeansMap.put(specializedBean, new AtomicLong(1));
+                        } else {
+                            specializedBeansMap.put(specializedBean, new AtomicLong(specializedBeansMap.get(specializedBean).get() + 1));
+                        }
+                    }
```
mkouba: This produces findbug warning:

> 06:13:08 [INFO] Sequence of calls to java.util.concurrent.ConcurrentHashMap may not be atomic in org.jboss.weld.bootstrap.SpecializationAndEnablementRegistry$BeansSpecializedByBean.apply(Bean) ["org.jboss.weld.bootstrap.SpecializationAndEnablementRegistry$BeansSpecializedByBean"] At SpecializationAndEnablementRegistry.java:[lines 79-115]

I think this should work:
```
AtomicLong value = specializedBeansMap.get(specializedBean);
if (value == null) {
  value = new AtomicLong(1);
  AtomicLong previous = specializedBeansMap.putIfAbsent(specializedBean, value);
  if (previous != null) {
    value = previous;
  }
}
value.incrementAndGet();
```

manovotn: Noticed the warning. Thanks for suggestion.

impl/src/main/java/org/jboss/weld/util/collections/AbstractMultimap.java
manovotn: I like this approach only I would keep the structure pretty much as it is and use `instanceof` check in the constructor to determine whether it is `ConcurrentMap` or not. That way we wouldn't need to make many changes.
Or is there a caveat to `instanceof`?

mkouba: No caveat I'm aware of ;-)

impl/src/main/java/org/jboss/weld/util/collections/AbstractMultimap.java
```diff
@@ -44,18 +45,33 @@
     protected final Supplier<C> supplier;
 
     private final Map<K, C> map;
+    private final ConcurrentMap<K, C> concurrentMap;
+    private final boolean concurrentMapProvided;
```
mkouba: This flag seems unnecessary, just check whether `concurrentMap` is null or not. To make the code cleaner I would add a private method `getMap()`:
```
private Map<K, C> getMap() {
   return concurrentMap != null ? concurrentMap : map;
}
```
This method could be used in most of the methods, e.g.:
```
  public int size() {
    return getMap().size();
  }
```

mkouba:
Merged, thanks!

PR #1437, opened 2016-08-02
Bunch of tweaks to make Weld 2.4 build on JDK9

manovotn:


manovotn:
Will require version upgrade of Weld API and CDI TCK to compile on JDK 9.
Relevant PRs are:
* API - https://github.com/weld/api/pull/44
* CDI TCK - https://github.com/cdi-spec/cdi-tck/pull/108

manovotn:
Note to self:
We should try this again once there is new JDK 9 EA version (currently running on build 129). This version fails several tests with NPE [(link to Pastebin)](http://pastebin.com/2nNsMCch) which comes from JDK internal implementation of Set (that is, from Map) . This does not happen with older EA version - build 119.

mkouba:
Merged, thanks!

PR #1439, opened 2016-08-03
WELD-2186 Review the usage of ArraySet

manovotn:
`ArraySet` and `ArraySetMultimap` were removed and replaced with appropriate replacements which were backported.
Along this I re-applied changed made in WELD-2156.
Last but not least this commit includes a clean up on remaining util methods for Guava which was removed.


impl/src/main/java/org/jboss/weld/annotated/slim/unbacked/UnbackedAnnotatedType.java
```diff
@@ -59,17 +59,17 @@ private UnbackedAnnotatedType(AnnotatedType<X> source, AnnotatedTypeIdentifier i
         for (AnnotatedConstructor<X> constructor : source.getConstructors()) {
             constructors.add(UnbackedAnnotatedConstructor.of(constructor, this, cache));
         }
-        this.constructors = immutableGuavaSet(constructors);
+        this.constructors = immutableSet(constructors);
```
mkouba: I'd use `org.jboss.weld.util.collections.ImmutableSet.Builder` here and below...

impl/src/main/java/org/jboss/weld/bean/attributes/BeanAttributesFactory.java
```diff
@@ -171,7 +171,7 @@ protected void initQualifiers(Set<Annotation> qualifiers) {
             if (qualifiers.isEmpty()) {
                 this.qualifiers = DEFAULT_QUALIFIERS;
             } else {
-                ArraySet<Annotation> normalizedQualifiers = new ArraySet<Annotation>(qualifiers.size() + 2);
+                Set<Annotation> normalizedQualifiers = new HashSet<Annotation>(qualifiers.size() + 2);
```
mkouba: `org.jboss.weld.util.collections.ImmutableSet.Builder`?

impl/src/main/java/org/jboss/weld/event/ObserverMethodImpl.java
```diff
@@ -131,8 +131,8 @@ protected ObserverMethodImpl(final EnhancedAnnotatedMethod<T, ? super X> observe
             }
             injectionPoints.add(injectionPoint);
         }
-        this.injectionPoints = immutableGuavaSet(injectionPoints);
-        this.newInjectionPoints = immutableGuavaSet(newInjectionPoints);
+        this.injectionPoints = immutableSet(injectionPoints);
```
mkouba: `ImmutableSet.Builder` for `injectionPoints`. `newInjectionPoints` seem to be unsued - remove pls.

impl/src/main/java/org/jboss/weld/util/InjectionPoints.java
```diff
@@ -48,11 +47,11 @@ private InjectionPoints() {
     }
 
     public static <T extends WeldInjectionPointAttributes<?, ?>> Set<T> flattenInjectionPoints(List<? extends Set<T>> fieldInjectionPoints) {
-        ArraySet<T> injectionPoints = new ArraySet<T>();
+        Set<T> injectionPoints = new HashSet<T>();
         for (Set<T> i : fieldInjectionPoints) {
             injectionPoints.addAll(i);
         }
-        return injectionPoints.trimToSize();
+        return injectionPoints;
```
mkouba: I think it makes sense to return `ImmutableSet`...

impl/src/main/java/org/jboss/weld/bean/attributes/BeanAttributesFactory.java
manovotn: The point of this change is that `normalizedQualifiers` are send into `SharedObjectCache` as an immutable set?
If so, we need to do it differently (`ImmutableSet.copyOf`), because `Builder` cannot do `remove()` method (L182)

mkouba: Yes, that's the point. Either use `ImmutableSet.copyOf()` or modify the algorithm not to use `remove()` ;-) 

`if (qualifiers.size() == 1) {` could be also "merged" with the nested condition.

manovotn:
Removing `fixme` label, I reflected all the changes mentioned in comments (and via chat). 
@mkouba please review the changes (last commit) and let me know if there is anything fishy :)

mkouba:
Merged, thanks!

PR #1441, opened 2016-08-03
WELD-2209 Use provided ClassLoader/ResourceLoader to load extensions

mkouba:
- also log each extension bean deployed


mkouba:
Merged.

PR #1443, opened 2016-08-03
WELD-2207 fix TODO's and remove environment.servlet.util.ServiceLoader

tremes:


impl/src/main/java/org/jboss/weld/util/ServiceLoader.java
```diff
@@ -232,27 +232,18 @@ private void loadService(String serviceClassName, URL file, int lineNumber) {
         return serviceClass;
     }
 
-    private S prepareInstance(Class<? extends S> serviceClass) {
-        try {
-            // TODO Support the SM
-            Constructor<? extends S> constructor = serviceClass.getDeclaredConstructor();
-            constructor.setAccessible(true);
-            return constructor.newInstance();
-        } catch (LinkageError e) {
```
tremes: I am not sure whether we can remove this.

environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/WeldServletLifecycle.java
```diff
@@ -64,7 +65,6 @@
 import org.jboss.weld.environment.servlet.deployment.WebAppBeanArchiveScanner;
 import org.jboss.weld.environment.servlet.logging.WeldServletLogger;
 import org.jboss.weld.environment.servlet.services.ServletResourceInjectionServices;
-import org.jboss.weld.environment.servlet.util.ServiceLoader;
```
mkouba: Hm, I think you cannot just replace the import - it works with `Iterable<Container>` and not `Iterable<Metadata<Container>>`.

impl/src/main/java/org/jboss/weld/util/ServiceLoader.java
mkouba: Hm, I think we should follow the ServiceLoader from JDK, i.e. catch `Throwable` and throw `ServiceConfigurationError`?

tremes: Well maybe. TBH I don't know the difference between `LinkageError` and `InstantiationException` in the original class so it's probably easier to throw `ServiceConfigurationError`.

mkouba:
Merged, thanks!

PR #1451, opened 2016-08-09
WELD-2216 provide additional info for Instance.get() deployment probl…

tremes:
…ems.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/builtin/InstanceImpl.java
+++ b/impl/src/main/java/org/jboss/weld/bean/builtin/InstanceImpl.java
@@ -50,6 +50,7 @@ import org.jboss.weld.manager.BeanManagerImpl;
 import org.jboss.weld.resolution.Resolvable;
 import org.jboss.weld.resolution.ResolvableBuilder;
 import org.jboss.weld.resolution.TypeSafeBeanResolver;
+import org.jboss.weld.util.InjectionPoints;
 import org.jboss.weld.util.Preconditions;
 import org.jboss.weld.util.collections.WeldCollections;
 import org.jboss.weld.util.reflection.Formats;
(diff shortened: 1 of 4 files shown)


impl/src/main/java/org/jboss/weld/bean/builtin/InstanceImpl.java
```diff
@@ -101,9 +101,17 @@ public T get() {
         if (bean != null) {
             return getBeanInstance(bean);
         } else if (isUnsatisfied()) {
-            throw BeanManagerLogger.LOG.unresolvableElement("Type: " + getType() + "; Qualifiers: " + getQualifiers());
+            throw BeanManagerLogger.LOG.injectionPointHasUnsatisfiedDependencies(ip,
```
mkouba: Including `ip.toString()` in the log message does not make sense here - it's a `DynamicLookupInjectionPoint` instance with no meaningful info.

impl/src/main/java/org/jboss/weld/bean/builtin/InstanceImpl.java
```diff
@@ -101,9 +101,17 @@ public T get() {
...
+                    Formats.formatInjectionPointType(ip.getType()),
+                    Formats.formatAsStackTraceElement(ip),
+                    InjectionPoints.getUnsatisfiedDependenciesAdditionalInfo(ip, getBeanManager()));
         } else {
-            throw BeanManagerLogger.LOG.ambiguousBeansForDependency(WeldCollections.toMultiRowString(allBeans));
+            throw BeanManagerLogger.LOG.injectionPointHasAmbiguousDependencies(ip,
+                    Formats.formatAnnotations(ip.getQualifiers()),
+                    Formats.formatInjectionPointType(ip.getType()),
+                    Formats.formatAsStackTraceElement(ip),
+                    InjectionPoints.getUnsatisfiedDependenciesAdditionalInfo(ip, getBeanManager()));
```
mkouba: This is wrong. Ambiguous dependency means more than one beans. The original `WeldCollections.toMultiRowString(allBeans)` was correct.

impl/src/main/java/org/jboss/weld/bean/builtin/InstanceImpl.java
```diff
@@ -101,9 +101,17 @@ public T get() {
         if (bean != null) {
             return getBeanInstance(bean);
         } else if (isUnsatisfied()) {
-            throw BeanManagerLogger.LOG.unresolvableElement("Type: " + getType() + "; Qualifiers: " + getQualifiers());
+            throw BeanManagerLogger.LOG.injectionPointHasUnsatisfiedDependencies(ip,
+                    Formats.formatAnnotations(ip.getQualifiers()),
+                    Formats.formatInjectionPointType(ip.getType()),
+                    Formats.formatAsStackTraceElement(ip),
```
mkouba: Stack trace element should not be there. Remember that a real injection point is not always available - i.e. `Instance.select()`.

impl/src/main/java/org/jboss/weld/bootstrap/Validator.java
```diff
@@ -922,16 +903,23 @@ private static void reallyValidatePseudoScopedBean(Bean<?> bean, BeanManagerImpl
...
+                if (!(bean instanceof AbstractBuiltInBean<?>)) {
+                    if (!ij.isDelegate()) {
+                        boolean normalScoped = beanManager.isNormalScope(bean.getScope());
+                        if (!normalScoped && !(bean instanceof SessionBean)) {
+                            reallyValidatePseudoScopedBean(bean, beanManager, dependencyPath, validatedBeans);
+                        }
                     }
                 }
             }
+        } catch (AmbiguousResolutionException ex) {
```
mkouba: I would only use the `try-catch` block for `beanManager.resolve()` to make it clearer.

mkouba:
Merged, thanks!

PR #1454, opened 2016-08-11
WELD-2218 do not automatically add interceptor/decorator to set of bean classes and improve deploymentException message

tremes:
… of selected alternative/interceptor/decorator

### DIFF
--- a/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
+++ b/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
@@ -42,7 +42,6 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
-
 import javax.enterprise.context.spi.CreationalContext;
 import javax.enterprise.inject.UnsatisfiedResolutionException;
 import javax.enterprise.inject.Vetoed;
@@ -369,7 +368,7 @@ public class Weld implements ContainerInstanceFactory {
     }
 
     /**
-     * Enable interceptors for a synthetic bean archive. Interceptor classes are automatically added to the set of bean classes.
+     * Replace interceptors for a synthetic bean archive.
      *
      * @param interceptorClasses
      * @return self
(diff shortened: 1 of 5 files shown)


environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -976,4 +984,40 @@ public boolean equals(Object obj) {
...
+        if (!enabledDecorators.isEmpty()) {
+            String result = checkMetadata.apply(enabledDecorators);
+            if (result != null) {
+                throw WeldSELogger.LOG.metadataNotFoundInBeanClasses(Decorator.class.getSimpleName().toLowerCase(), result);
+            }
+        }
+
+    }
+
+    private final Function<List<Metadata<String>>, String> checkMetadata = new Function<List<Metadata<String>>, String>() {
```
mkouba: There's no need to use function here - it could be a stateless private method, i.e. `private String checkMetadata(List<Metadata<String> metadata)`.

I'm missing a simple test :-)

mkouba:
Well, I've been thinking about this for a while and I believe it does not always make sense to add those classes automatically. Imagine an interceptor class from a library - then you will not be able to enable this interceptor in the synthetic bean archive and still have the interceptor class in the library jar. All those interceptors/decorators/alternatives `add` methods should only simulate adding elements in `beans.xml`. What do you think?

mkouba:
PS. I'm sorry I've changed my mind again ;-)

tremes:
TBH I don't get it much. What do you mean by library in SE? Those `add` methods make sense only in the synthetitc bean archive and thus selected or enabled alternative/interceptor/decorator needs to be a bean anyway.
I mean there is no other way for alternative/interceptor/decorator how to become a bean in the synthetic bean archive than explicitly add it as a bean class and enable/select it together. If I am not mistaken. So I am still thinking this could be done in one step instead of two steps.

mkouba:
The point is the synthetic archive may exist side by side "physical" bean archives if you don't disable discovery via `Weld.disableDiscovery()`. E.g. you can have `foo.jar` containing `FooInterceptor.class` and `FooInterceptorBinding.class` on the class path. And bean class `Bar` annotated with `FooInterceptorBinding.class` in your synthetic archive. If `FooInterceptor.class` is not enabled globally, you will have to enable it for the synthetic bean archive. However, if we also add `FooInterceptor.class` to the set of bean classes of the synthetic bean archive, then `FooInterceptor.class` will be found in multiple bean archives and this may result in incorrect behavior.

tremes:
OK got it. Thanks for explanation.:-) So I am going to reintroduce the original proposal with this change - not automatically adding decorators/interceptors to set of bean classes.

mkouba:
Merged, thanks!

PR #1457, opened 2016-08-13
Fixes Issues WELD-2064 and WELD-2220

lprimak:
- fixed a bug when delimiter clashes with suffix
- added appropriate package to osgi bundle


mkouba:
Merged (except for osgi part which is not relevant in master branch), thanks!

PR #1461, opened 2016-08-24
Removing junction plugin from numberguess jsf sample.

manovotn:


manovotn:
Also removed dependency version override, the same version is used in `weld-parent` so it is not needed.

mkouba:
Merged, thanks!

PR #1463, opened 2016-08-24
WELD-2225 stick to new CDI 2.0-EDR2 SE API.

tremes:


environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -391,6 +430,28 @@ public Weld addExtension(Extension extension) {
...
+        }
+        return this;
+    }
+
+    @Override
+    public Weld addExtensions(Class<? extends Extension>... extensionClasses) {
+        Object[] constructorParams = new Object[] {};
+        for (Class<? extends Extension> extensionClass : extensionClasses) {
+            try {
+                Extension extension = extensionClass.getDeclaredConstructor(new Class[] {}).newInstance(constructorParams);
```
mkouba: Please use `org.jboss.weld.environment.se.SecurityActions.newInstance(Class<T>)` instead..

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -391,6 +430,28 @@ public Weld addExtension(Extension extension) {
...
+    public Weld addExtensions(Extension... extensions) {
+        for (Extension extension : extensions) {
+            addExtension(extension);
+        }
+        return this;
+    }
+
+    @Override
+    public Weld addExtensions(Class<? extends Extension>... extensionClasses) {
+        Object[] constructorParams = new Object[] {};
```
mkouba: This variable is useless in most cases. I would rather pass `new Object[] {}` to log method directly.

mkouba:
Merged, thanks!

PR #1465, opened 2016-08-26
Update weld-parent version to 35.

manovotn:


mkouba:
retest this please

manovotn:
Removing hold label, this should be fine now. The problem was that Maven-ant plugin was using bad ant sources (with outdated group id).

mkouba:
Merged, thanks!

PR #1468, opened 2016-08-31
WELD-2227 Fix WeldDeployment.loadBeanDeploymentArchive()

mkouba:
- also synchronize the access to the shared additional bda


mkouba:
Merged.

PR #1473, opened 2016-09-13
Use correct Jandex class when checking presence of new method

robmv:
Code was checking wrongly a class inside Weld that use that same name ClassInfo that the one on Jandex that is where the new method is present.

Patch should be backported to 2.4.0Final that has the same problem.


robmv:
So look like that class is not even on the build class path. What next? load it by reflection?

mkouba:
@robmv Thanks for your report! I've created https://issues.jboss.org/browse/WELD-2231 to track this issue.

Yep, Jandex is not a dependency of the Weld core. We should try to load the class by name, using `org.jboss.weld.resources.spi.ResourceLoader`.

@tremes Tom, please create a new PR based on this one and modify `FastAnnotatedTypeLoader` accordingly.

tremes:
Closing in favor of https://github.com/weld/core/pull/1474

PR #1474, opened 2016-09-14
WELD-2231 update check for ClassInfo.setFlags() method availability.

tremes:


impl/src/main/java/org/jboss/weld/bootstrap/FastAnnotatedTypeLoader.java
```diff
@@ -127,13 +128,19 @@
...
-
-    private boolean isJandexSetFlagsAvailable() {
-        try {
-            Method setFlags = AccessController.doPrivileged(GetDeclaredMethodAction.of(ClassFileInfo.class, "setFlags", short.class));
-            return setFlags != null;
-        } catch (Exception e) {
-            return false;
+    // checking availability of ClassInfo.setFlags method is just workaround for JANDEX-37
+    private boolean initCheckTypeModifiers() {
+        if (Reflections.isClassLoadable(CLASSINFO_CLASS_NAME, new ClassLoaderResourceLoader(classFileServices.getClass().getClassLoader()))) {
```
mkouba: I would first try to load the class and if it's not null then try to find the method - right now, `ClassLoaderResourceLoader` is created and `ResourceLoader.classForName(String)}}` is invoked twice.

impl/src/main/java/org/jboss/weld/bootstrap/FastAnnotatedTypeLoader.java
```diff
@@ -53,17 +55,16 @@
...
         this.fallback = new AnnotatedTypeLoader(manager, transformer, events);
         this.classFileServices = classFileServices;
         this.resolver = resolver;
-        if (isJandexSetFlagsAvailable()) {
-            this.checkTypeModifiers = true;
-        } else {
-            this.checkTypeModifiers = false;
+        this.checkTypeModifiers = initCheckTypeModifiers();
+        if (!checkTypeModifiers) {
             BootstrapLogger.LOG.usingOldJandexVersion();
```
mkouba: I would move the log message inside the catch block on line 139.

impl/src/main/java/org/jboss/weld/bootstrap/FastAnnotatedTypeLoader.java
```diff
@@ -128,12 +126,21 @@
...
-    private boolean isJandexSetFlagsAvailable() {
-        try {
-            Method setFlags = AccessController.doPrivileged(GetDeclaredMethodAction.of(ClassFileInfo.class, "setFlags", short.class));
-            return setFlags != null;
-        } catch (Exception e) {
-            return false;
+    // checking availability of ClassInfo.setFlags method is just workaround for JANDEX-37
+    private boolean initCheckTypeModifiers() {
+
+        Class<?> classInfoclass = Reflections.loadClass(CLASSINFO_CLASS_NAME, new ClassLoaderResourceLoader(classFileServices.getClass().getClassLoader()));
```
tremes: I am still not really sure how this will work in Servlet, SE environment but I don't want to think about it during weekend.:-) Will check at Monday.

tremes: It works ok in SE environment so I think it's ok in general.

mkouba: Great!

mkouba:
Merged, thanks!

PR #1480, opened 2016-09-25
fixed to set the newly created session on the thread local var

mstoeckel:


mkouba:
@mstoeckel This fix shouldn't be necessary because `CURRENT_SESSION.set()` is called within `request.getSession(true)` invocation, see also `org.jboss.weld.servlet.WeldInitialListener.sessionCreated(HttpSessionEvent)`. In which environment do you run into troubles?

mkouba:
I'm closing this pull request for now. Feel free to create a new issue (https://issues.jboss.org/projects/WELD) or contact the community with some concise problem description (http://weld.cdi-spec.org/community).

mstoeckel:
@mkouba 
I am experimenting with embedded Jetty and Weld where I add WeldInitialListener to the org.eclipse.jetty.servlet.ServletContextHandler.addEventListener(), and it is blowing up at org.jboss.weld.context.beanstore.http.AbstractSessionBeanStore:116 (lockStore = (LockStore) session.getAttribute(SESSION_KEY)) because the attempt to get session with getSession(true) from line 111 is returning null because of the logic error in SessionHolder.
Furthermore, purely from the api semantic perspective, SessionHolder.getSession(request, true) should always return a valid HttpSession object irrespective of how and from where that method is called. the following is an obvious logic error:
HttpSession session = CURRENT_SESSION.get();
if (create && session == null) {
            request.getSession(true);
            session = CURRENT_SESSION.get();
        }
If CURRENT_SESSION thread local var didn't contain session, you cannot possibly expect it to contain the session after merely calling request.getSession(true). 
Please reconsider my PR.

Thanks,
Michael

mstoeckel:
@mkouba 
BTW, I fixed my problem by the following:
```
WeldInitialListener weldListener = new WeldInitialListener(this.beanManager);
handler.addEventListener(weldListener);//this should have been sufficient
handler.getSessionHandler().addEventListener(weldListener);//adding this fixed the problem
```

mkouba:
@mstoeckel If you're using `ServletContextHandler` then `ContextHandler.addEventListener(EventListener)` is not sufficient, the javadoc is only talking about `ServletContextListener`, `ServletContextAttributeListener`, `ServletRequestListener` and `ServletRequestAttributeListener`. So from my point of view your fix is correct.

> If CURRENT_SESSION thread local var didn't contain session, you cannot possibly expect it to contain the session after merely calling request.getSession(true). 

Well, we expect a fully functional servlet container environment.

PR #1485, opened 2016-10-05
WELD-2238 configuration should log keys intead of enums.

manovotn:


mkouba:
Looks good. Just two points. `StringJoiner` is 1.8, so it might be better to use `StringBuilder` instead (the same code could be used in 2.4). We should also improve the  `ConfigurationLogger.configurationInitialized(Object)` message so that it's clear that only non-default/modified values are shown.

We could also improve `ConfigurationLogger.configurationPropertyTypeMismatch(Object, Object)` message so that it also contains the key.

manovotn:
Thanks for review.
I deliberately used `StringJoiner` in master and have a `StringBuidler` version for 2.4 (which is a bit more verbose).

As for the other improvements - I'll look and see what I can come up with.

manovotn:
@mkouba updated PR with log msgs.

mkouba:
Looks perfect now!

mkouba:
Merged, thanks!

PR #1488, opened 2016-10-13
WELD-708, fix passivation test.

manovotn:


manovotn:
BTW there is another ignored passivation test - `org.jboss.weld.tests.interceptors.passivation.PassivationActivationTest`
But that one cannot be fixed easily, if at all.  It tests `@Pre/PostPasivate` interceptors which require real passivation and not just a dummy one.

manovotn:
Actually, looking closely at the `PassivationActivationTest` I think we might want to remove it.
There is probably no reliable way to make it work - e.g. to trigger real server-side passivation. Best effort would require to **somehow** set EJB module in Wildfly to minimal value in order to enforce passivation as soon as possible.

mkouba:
Merged, thanks!

PR #1489, opened 2016-10-13
WELD-708, fix passivation test.

manovotn:
Master branch equivalent of core/pull/1488
Here the tests wasn't ignored but was false negative (incorrectly passed).


mkouba:
Merged, thanks!

PR #1495, opened 2016-10-20
WELD-2249 Use default discovery mode for 1.1 or later

mkouba:


mkouba:
We decided not to include this into 2.4 and 3.0 branches as it might cause insidious migration issues.

PR #1496, opened 2016-10-20
WELD-2249 Use default discovery mode for 1.1 or later

mkouba:


mkouba:
retest this please

tremes:
Following beans.xml file is resolved with bdm `annotated` (which is wrong):

`<beans` xmlns="http://xmlns.jcp.org/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd">`

Line https://github.com/weld/core/pull/1496/files#diff-9f1fcb30bd31380c331cf9c6cb81048dR310 gets value `1.1` (from xsd file probably) even in this case.

mkouba:
We decided not to include this into 2.4 and 3.0 branches as it might cause insidious migration issues.

PR #1497, opened 2016-10-26
WELD-2251 check whether SeContainer is running when getting BeanManager.

tremes:
This log message is a little bit questionable (we don't know if container was running before or not). Maybe we can introduce new logging message

### DIFF
--- a/impl/src/main/java/org/jboss/weld/AbstractCDI.java
+++ b/impl/src/main/java/org/jboss/weld/AbstractCDI.java
@@ -63,41 +63,49 @@ public abstract class AbstractCDI<T> extends CDI<T> {
 
     @Override
     public Iterator<T> iterator() {
+        checkState();
         return getInstance().iterator();
     }
 
     @Override
     public T get() {
+        checkState();
         return getInstance().get();
     }
 
     @Override
     public Instance<T> select(Annotation... qualifiers) {
+        checkState();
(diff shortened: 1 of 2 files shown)


mkouba:
Well, if we have a container instance, it must have been running before, or?

WRT implementation - what about public methods from `AbstractCDI`? It might make sense to introduce a protected `AbstractCDI.checkState()` method with default NO-OP impl, use it to guard methods like `select()` and override with `isRunning()` check in `WeldContainer`.

impl/src/main/java/org/jboss/weld/AbstractCDI.java
```diff
@@ -128,4 +136,11 @@ protected String getCallingClassName() {
         return instanceCache.getValue(BeanManagerProxy.unwrap(getBeanManager()));
     }
 
+    /**
+     * Check whether container is running
+     */
+    protected void checkState(){
```
manovotn: Why not just make this `abstract`? So that subclasses (WeldContainer) have to implement this.
`protected abstract void checkState();`

tremes: We don't want to force {{SimpleCDI}} to implement this method AFAIK.

mkouba:
Merged, thanks!

PR #1500, opened 2016-11-03
Enhance Utils.deserialize, fix EE producer test and re-enable it.

manovotn:
Will **not pass** until we run it against WFLY 11 as it contains fix for WFLY-835.

### DIFF
--- a/tests-common/src/main/java/org/jboss/weld/test/util/Utils.java
+++ b/tests-common/src/main/java/org/jboss/weld/test/util/Utils.java
@@ -107,6 +107,12 @@ public class Utils {
         }
     }
 
+    public static <T> T deserialize(byte[] bytes, ClassLoader cl) throws IOException, ClassNotFoundException {
+        try (TCCLObjectInputStream in = new TCCLObjectInputStream(new ByteArrayInputStream(bytes), cl)) {
+            return Reflections.<T>cast(in.readObject());
+        }
+    }
+
     public static boolean isExceptionInHierarchy(Throwable exception, Class<? extends Throwable> expectedException) {
         while (exception != null) {
             if (exception.getClass().equals(expectedException)) {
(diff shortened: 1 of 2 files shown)


tests-common/src/main/java/org/jboss/weld/test/util/Utils.java
```diff
@@ -177,12 +183,19 @@ public static boolean isProxy(Object proxy) {
     private static class TCCLObjectInputStream extends ObjectInputStream {
 
         private final ClassLoader classLoader;
+        private ClassLoader lastResortClassLoader = null;
```
mkouba: `lastResortClassLoader`? Choose a better name pls :D The original `classLoader` could be renamed to `tccl` or something like that.Also the field should be final.

tests-common/src/main/java/org/jboss/weld/test/util/Utils.java
```diff
@@ -177,12 +183,19 @@ public static boolean isProxy(Object proxy) {
...
 
         public TCCLObjectInputStream(InputStream in) throws IOException {
             super(in);
             this.classLoader = Thread.currentThread().getContextClassLoader();
```
mkouba: Replace with `this(in, null)`

tests-common/src/main/java/org/jboss/weld/test/util/Utils.java
```diff
@@ -192,7 +205,17 @@ public TCCLObjectInputStream(InputStream in) throws IOException {
...
-                    return BeanManagerImpl.class.getClassLoader().loadClass(desc.getName());
+                    try {
+                        return BeanManagerImpl.class.getClassLoader().loadClass(desc.getName());
+                    } catch (ClassNotFoundException cnfe) {
+                        if (lastResortClassLoader != null) {
+                            // should all else fail, try the alternatively supplied CL
+                            return lastResortClassLoader.loadClass(desc.getName());
+                        } else {
+                            // rethrow the exception, we cannot handle this
+                            throw new ClassNotFoundException(cnfe.getMessage(), cnfe.getCause());
```
mkouba: Why not simply `throw cnfe`?

manovotn:
Adjusted accordingly to comments, I also put `@Ignore` on the test with a comment.

mkouba:
Merged, thanks!

PR #1501, opened 2016-11-04
Enhance Utils.deserialize, fix EE producer test but keep it ignored, …

manovotn:
…until we use WFLY 11.


mkouba:
Merged, thanks!

PR #1502, opened 2016-11-04
Deserialization in tests should use TCCLObjectInputStream. Fix EE pro…

manovotn:
…ducer test.

This should stay on hold until EAP 6.4.12 is released and our osprey executors switch to it.
Until then, the test will fail.


manovotn:
@mkouba I actually realized we can merge this right away. The test passes even now (we push in our proxy so it is serializable). Removing hold label.

mkouba:
Merged, thanks!

PR #1503, opened 2016-11-04
WELD-2255 GroovyMethodFilter - filter methods only from groovy objects

mkouba:


impl/src/main/java/org/jboss/weld/bean/proxy/GroovyMethodFilter.java
```diff
@@ -54,15 +55,29 @@ public boolean isEnabled() {
...
     @Override
-    public boolean accept(Method method) {
+    public boolean accept(Method method, Class<?> proxySuperclass) {
         if (GROOVY_OBJECT.equals(method.getDeclaringClass().getName())) {
             return false;
         }
-        for (MethodSignature groovyMethod : METHODS) {
-            if (groovyMethod.matches(method)) {
-                return false;
+        if (isGroovyObject(proxySuperclass)) {
```
leifbladt: What's the difference between `isGroovyObject(proxySuperclass)` and `GROOVY_OBJECT.equals(method.getDeclaringClass().getName())` (three lines before)?

mkouba: `isGroovyObject()` is used to detect any Groovy object (must implement `groovy.lang.GroovyObject`). `GROOVY_OBJECT.equals(method.getDeclaringClass().getName())` is perhaps unnecessary because it only matches methods from `GroovyObject` interface and any "groovy object" implements all these methods but I'm no Groovy expert so I think it would be safer to leave it here.

mkouba:
Merged.

PR #1512, opened 2016-12-04
WELD-2260 When the WeldManager has been located, track its id as a co…

johnament:
…ntext attribute.

Later on in `WeldInitialListener`, the ID is used to locate the underlying bean manager.  This was failing in 2.4.1 due to the changes in how IDs are used (when not specified).  Specifically this line https://github.com/weld/core/blob/2.4/impl/src/main/java/org/jboss/weld/servlet/WeldInitialListener.java#L85 that is looking it up.  The failure was happening on line #94.

### DIFF
--- a/environments/servlet/tests/jetty/pom.xml
+++ b/environments/servlet/tests/jetty/pom.xml
@@ -314,9 +314,10 @@
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration>
-                            <excludes>
-                                <exclude>%regex[.*servlet/test/(?!se).*]</exclude>
-                            </excludes>
+                            <includes>
+                                <include>org.jboss.weld.environment.servlet.test.se.coop.BootstrapNotNeededTest</include>
+                                <include>org.jboss.weld.environment.servlet.test.se.coop.builder.WeldSeBuilderTest</include>
+                            </includes>
                         </configuration>
                     </plugin>
                 </plugins>
--- a/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/WeldServletLifecycle.java
+++ b/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/WeldServletLifecycle.java
@@ -70,6 +70,7 @@ import org.jboss.weld.environment.undertow.UndertowContainer;
 import org.jboss.weld.environment.util.DevelopmentMode;
 import org.jboss.weld.environment.util.Reflections;
 import org.jboss.weld.injection.spi.ResourceInjectionServices;
+import org.jboss.weld.manager.BeanManagerImpl;
 import org.jboss.weld.manager.api.WeldManager;
 import org.jboss.weld.resources.ManagerObjectFactory;
 import org.jboss.weld.resources.WeldClassLoaderResourceLoader;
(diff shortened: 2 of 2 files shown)


environments/servlet/tests/jetty/pom.xml
```diff
@@ -314,9 +314,9 @@
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration>
-                            <excludes>
-                                <exclude>%regex[.*servlet/test/(?!se).*]</exclude>
-                            </excludes>
+                            <includes>
```
johnament: So here's one interesting thing.  I was trying to figure out how to run the tests, these tests weren't working due to bad regex.  I don't think this is quite right, but let me know.  

One other thing - even with this change, I'm still seeing failures (but it works fine when I test with a test app).  For some reason, the underlying `WeldManager`'s ID is based on a path, not the UUID.

mkouba:
@johnament Thanks for the PR!

I've tested your PR with Hammock's `UndertowBootTest` with no success. Then I realized that we would have to replace `context.setAttribute(org.jboss.weld.Container.CONTEXT_ID_KEY, manager.getId())` with:

    context.setInitParameter(org.jboss.weld.Container.CONTEXT_ID_KEY, BeanManagerProxy.unwrap(manager).getContextId());

* `setInitParameter()` because `WeldInitialListener` is using `ServletContext.getInitParameter()`
* `BeanManagerImpl..getContextId()` because `WeldManager.getId()` returns a bean deployment archive ID, not the ID of the current container instance

Also we have to set the init param inside `if (container instanceof ContainerInstance) {}` block below because Weld also supports to hand over `Weld` and `WeldContainer` instances.    

WRT regex - hmm, you're right the regex fails on 2.4 but works on 2.3 (with older surefire plugin). Looks like a surefire bug - build fails if a regex contains exclamation mark: `Exclamation mark not expected in 'exclusion'` (e.g. when using a negative lookahead like here). Since we only have two tests in here I think that it's ok to have two specific includes:

    <includes>
        <include>org.jboss.weld.environment.servlet.test.se.coop.BootstrapNotNeededTest</include>
        <include>org.jboss.weld.environment.servlet.test.se.coop.builder.WeldSeBuilderTest</include>
     </includes>

johnament:
Ok, then I'm not sure how I got this to pass locally.  You're 100% correct.  Let me push this up.

environments/servlet/tests/jetty/pom.xml
johnament: I'm not going to include `WeldSeBuilderTest` quite yet - it's failing locally for other reasons which probably warrants its own PR.

mkouba: It's failing because we're still missing `context.setInitParameter()` on line 154 - see also my last comment:

> Also we have to set the init param inside if (container instanceof ContainerInstance) {} block below because Weld allows to hand over Weld and WeldContainer instances.

mkouba: Weld/WeldContainer are not broken - it's the same problem, but instead of `BeanManager`, either `Weld` (builder) or `WeldInstance` (container instance) is passed from Weld SE. Just add `context.setInitParameter(org.jboss.weld.Container.CONTEXT_ID_KEY, contextId)` to the _if_ block mentioned above.

johnament:
@mkouba let me know if you need any more changes here or if its good.

mkouba:
Looks good. Thanks!

mkouba:
Merged, thanks!

PR #1515, opened 2016-12-09
WELD-2265 introduce ProcessSynthethicBean impl.

tremes:


tremes:
My first idea was to put this PSB firing to `org.jboss.weld.bootstrap.events.AfterBeanDiscoveryImpl#processBeanRegistration`. Not sure what's better TBH.

impl/src/main/java/org/jboss/weld/bootstrap/events/ContainerLifecycleEvents.java
```diff
@@ -241,6 +252,9 @@ public void fireProcessBean(BeanManagerImpl beanManager, Bean<?> bean) {
             } else if (bean instanceof ProducerMethod<?, ?>) {
                 ProcessProducerMethodImpl.fire(beanManager, (ProducerMethod<?, ?>) bean);
             } else {
+                if (isProcessSynthethicBeanObserved()) {
```
mkouba: I wouldn't detect `processSynthethicBeanObserved` at all (we don't detect `ProcessSyntheticAnnotatedType` neither). Instead, just check the extension param:

    if (extension != null) {

mkouba: In fact, it might make sense to create a simple wrapper for `Bean<?>` and `Extension`, i.e. `BeanContext` like we do for `SlimAnnotatedType` (see `SlimAnnotatedTypeContext`).

mkouba:
Merged, thanks!

PR #1516, opened 2016-12-09
WELD-2267 add interceptor which can activate Request Context

manovotn:


tests/src/test/java/org/jboss/weld/tests/unit/bootstrap/xml/BeansXmlTest.java
```diff
@@ -97,7 +97,7 @@ public void testInterceptorsEnabled() throws Exception {
 
             public void run(WeldManager beanManager) {
                 ModuleEnablement enabled = ((BeanManagerImpl) beanManager).getEnabled();
-                assertEquals(2, enabled.getInterceptors().size());
+                assertEquals(3, enabled.getInterceptors().size());
```
manovotn: Had to change unit tests here - from what I gathered, this tests all enabled interceptors in given module. In this case one interceptor is from test itself, 2 others come from weld core.

mkouba: Hm, ok. It might be better to test whether a concrete interceptor classes are present...

impl/src/main/java/org/jboss/weld/context/activator/ActivateRequestContextInterceptor.java
```diff
@@ -51,20 +50,10 @@ public ActivateRequestContextInterceptor(@Unbound RequestContext requestContext,
...
-        } else {
-            try {
-                requestContext.activate();
-                return ctx.proceed();
-            } finally {
-                requestContext.invalidate();
-                requestContext.deactivate();
-            }
-        }
+        return RequestContextActivationUtils.activateContext(ctx, requestContext, beanManager);
```
mkouba: I would try to use abstract superclass instead of util methods...

manovotn:
Introduced the abstract class instead of utils.
As for the test, I would keep it as is - it's purpose is to test module-wide enabled CDI interceptors/decorators/...

mkouba:
Merged, thanks!

PR #1536, opened 2016-12-20
WELD-2291 throw ISE also when data for InterceptionFactory are cached.

tremes:


impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
```diff
@@ -108,6 +108,7 @@ public T createInterceptedInstance(T instance) {
                 .getInterceptionFactoryData(configurator != null ? configurator.complete() : annotatedType);
 
         if (!cached.isPresent()) {
+            used = true;
```
mkouba: I think we should move `used = true;` right after `throw InterceptorLogger.LOG.interceptionFactoryNotReusable()` and also remove L125.

mkouba:
Merged, thanks!

PR #1547, opened 2017-01-04
WELD-2299 Add logging for InterceptionFactory.

manovotn:


mkouba:
I would also add some debug message when interceptors are not required (and proxy is not generated), ie.:
    
    if (!cached.isPresent()) {
           // log...  
           return instance;
    }

manovotn:
@mkouba not sure I am following here. What do you mean by 'not required'? That interceptor proxy already exists? In what situation would that happen?

mkouba:
We don't have to create a proxy if there are no `@AroundInvoke` interceptors associated. I'm not sure we have a test for this and whether it works as expected though...

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
```diff
@@ -65,6 +66,13 @@ public void testFooClassLevelBinding(@Produced("classLevel") Foo foo) {
     public void testFooEjbInterceptors(@Produced("ejbInterceptors") Foo foo) {
         assertEquals("Hello pong", foo.ping());
     }
+    
+    @Test
+    public void testFooNoInterceptors(@Produced("empty") Foo foo) {
+        // Note that the produced Foo is @Dependent
+        assertEquals(Foo.class, foo.getClass());;
```
manovotn: There is double `;` on this line.

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
```diff
@@ -65,6 +66,13 @@ public void testFooClassLevelBinding(@Produced("classLevel") Foo foo) {
...
+        assertEquals("pong", foo.ping());
```
manovotn: @mkouba I don't see how this verifies there is no proxy created?
The fact that you get `pong` string back is derived from the mere absence of **any** interceptor binding. It does not mean you don't have the proxy created, right?

mkouba: `assertEquals(Foo.class, foo.getClass())`. The produced `Foo` is `@Dependent` and so there is no client proxy and we can be sure the instance is the result of the producer method.

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
mkouba: Oops

mkouba:
Merged, thanks!

PR #1551, opened 2017-01-09
WELD-2277 Remove useless TODO comments

mkouba:
From now on, a TODO comment should only be used if it contains a key of an existing Weld issue.


mkouba:
Merged.

PR #1552, opened 2017-01-10
WELD-2213 Upgrade jboss-classfilewriter to 1.2.1.Final

mkouba:
- Weld still compiles against jboss-classfilewriter 1.1.x


mkouba:
Merged.

PR #1553, opened 2017-01-10
WELD-2307 Allow reflection calls for jboss-classfilewriter in tests.

manovotn:
Note that scope of WELD-2307 might be wider, e.g. more PRs might be coming due to reflection changes.
This one ensures that whenever jboss-classfilewriter is used, it is "granted reflection access" to JDK internals.

### DIFF
--- a/tests-arquillian/pom.xml
+++ b/tests-arquillian/pom.xml
@@ -227,7 +227,7 @@
                             </systemProperties>
                             <test>${test}</test>
                             <!-- Travis CI build workaround -->
-                            <argLine>${travis.surefire.argLine}</argLine>
+                            <argLine>${surefire.plugin.jdk9.args} ${travis.surefire.argLine}</argLine>
                         </configuration>
                     </plugin>
                 </plugins>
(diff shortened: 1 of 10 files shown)


manovotn:
Retest this please

manovotn:
Not sure why the osprey machine fails to start managed container for intergration tests. It works for me locally and the PR only adds JVM args which are in fact empty string on Java 8. Will need to look further into this.

tests-arquillian/pom.xml
```diff
@@ -272,7 +272,7 @@
                                 <arquillian.xml>wildfly-arquillian.xml</arquillian.xml>
                                 <jacoco.agent>${jacoco.agent}</jacoco.agent>
                                 <node.address>${node.address}</node.address>
-                                <additional.vm.args>${additional.vm.args}</additional.vm.args>
+                                <additional.vm.args>${surefire.plugin.jdk9.args}${additional.vm.args}</additional.vm.args>
```
manovotn: BTW, this one is more a paranoia than anything else. We are not yet testing this on WildFly, but I presume that as soon as we will, the same issue will pop up. Therefore, I added it right away.

mkouba:
Merged, thanks!

PR #1554, opened 2017-01-10
WELD-2308 Default scope for beans created from BeanConfigurator

mkouba:


tremes:
Actually I found out one little discrepancy here and it's the case when you call `javax.enterprise.inject.spi.configurator.BeanConfigurator#read(javax.enterprise.inject.spi.AnnotatedType<U>)`. Because in this case the scope of added bean is copied from the original AT and thus {{https://github.com/weld/core/pull/1554/files#diff-b16b9e99c5a67cdbceb32d229a0049d8R253}} will be always true. 

I think this is worth of CDI spec issue as well since this behavior makes sense I think. Btw. currently spec doesn't talk about {{BeanConfigurator.read}} methods at all AFAIK.

mkouba:
@tremes I don't understand. Could you elaborate?

tremes:
I tried to sum it up in https://issues.jboss.org/browse/CDI-681 and you can check following commit https://github.com/tremes/cdi-tck/commit/14f593b7cdfcea8d3cca2c880b303d8bb5a4cb19 which adds test for this. 
E.g the question could be: What will happen when I read some AnnotatedType by BeanConfigurator in ABD and add some stereotype defining scope? Which scope does this new bean have?

mkouba:
@tremes Yes, that makes sense.

mkouba:
Merged.

PR #1558, opened 2017-01-11
WELD-2305 JDK 9 package clash fixes

manovotn:
Split into two commits. First addresses clash between API and Core by changing core package name.
The other changes packages for all Weld modules as there were multiple clashes with both, API and Core (see stack in WELD-2305).
Technically, we could change just some of the packages, but then it would look weird/inconsistent. So I went ahead and changes all module packages to `org.jboss.weld.module.${module_name}`.

Along with that I also had to change module's META-INF/services in order to load them properly.

**NOTE:** Will require changes in WFLY (looking at it now).


manovotn:
@mkouba @tremes please let me know what you think  of this approach or whether you have any better ideas :)

manovotn:
Fixed one last glitch in servlet core to make it work on Tomcat.
@mkouba this is a complete PR, with it whole build and tests passed (using WFLY with https://github.com/weld/wildfly/pull/10)

impl/src/main/java/org/jboss/weld/contexts/AbstractBoundContext.java
```diff
@@ -14,9 +14,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.weld.context;
+package org.jboss.weld.contexts;
 
-import org.jboss.weld.context.beanstore.BoundBeanStore;
+import org.jboss.weld.context.BoundContext;
```
tremes: It's not so important, but is this import necessary?

manovotn: Not sure I see what you mean... `AbstractBoundContext implements BoundContext`?

mkouba:
Merged, thanks!

PR #1559, opened 2017-01-11
WELD-2306 enable EJBTest.

tremes:
This is basic MDB test. I think it's also tested in TCK but still we can keep it.

### DIFF
--- a/tests-arquillian/src/test/java/org/jboss/weld/tests/ejb/mdb/EJBTest.java
+++ b/tests-arquillian/src/test/java/org/jboss/weld/tests/ejb/mdb/EJBTest.java
@@ -24,8 +24,6 @@ import javax.jms.QueueSession;
 import javax.jms.Session;
 import javax.naming.InitialContext;
 
-import org.junit.Assert;
-
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.shrinkwrap.api.Archive;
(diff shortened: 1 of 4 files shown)


tests-arquillian/src/test/java/org/jboss/weld/tests/ejb/mdb/EJBTest.java
```diff
@@ -51,20 +48,16 @@
...
     @Test
-    @Ignore
-    // TODO Need a way to deploy test-destinations-service.xml to JBoss AS
     public void testMdbUsable(Control control) throws Exception {
         InitialContext ctx = new InitialContext();
         QueueConnectionFactory factory = (QueueConnectionFactory) ctx.lookup("ConnectionFactory");
         QueueConnection connection = factory.createQueueConnection();
         QueueSession session = connection.createQueueSession(false, Session.AUTO_ACKNOWLEDGE);
-        Queue queue = (Queue) ctx.lookup("queue/testQueue");
+        Queue queue = (Queue) ctx.lookup("java:jboss/exported/jms/queue/test");
```
mkouba: Is it safe to use a hard-coded JNDI name? I mean the name of the test queue could change in the next WildFly version and so might not be able to run the tests against multiple WildFly versions...

tremes: Well I think yes because https://github.com/weld/core/blob/master/tests-arquillian/src/test/java/org/jboss/weld/tests/util/WeldExtension.java handles this in https://github.com/weld/core/blob/master/tests-arquillian/src/test/java/org/jboss/weld/tests/util/WildFly8EEResourceManager.java#L40

We can export this string as a constant e.g in https://github.com/weld/core/blob/master/tests-arquillian/src/test/java/org/jboss/weld/tests/util/WildFlyMessaging.java

mkouba: @tremes You're right.
> We can export this string as a constant...

That's a good idea.

mkouba:
Merged, thanks!

PR #1563, opened 2017-01-16
WELD-2312 fix JDK 9 module clashes between Weld Core and API.

manovotn:


manovotn:
**NOTE:** CI cannot pass this as we need to build WFLY with newest Weld version to have the packages correct.
@mkouba also, WildFly will again need to add [`<module name="org.jboss.weld.api"/>`](https://github.com/weld/wildfly/commit/92acc2c6f2c72b7733d4ce1629aea46c0d6318b8#diff-2a1e1b801961cfa8b6689c2c07c6971dR45) into its jsf-injection module.

However, I did the testing locally, and after applying the above change, all tests passed for me.

manovotn:
We decided to restrain from applying such vast changes to a bugfix release version of Weld. Note that Weld 2.4+ will still be executable with JDK 9 for as long as one doesn't use module-info. Closing the PR.

PR #1572, opened 2017-01-26
WELD-2322 Fix performance issue with WeldManager.instance()

mkouba:


mkouba:
Note to myself: we should also align the relevant parts in 2.4 branch with the proposed solution.

mkouba:
Merged.

PR #1573, opened 2017-01-26
WELD-2323 InstanceImpl optimization

mkouba:
- do not prefetch all resolved beans for Instance<Object> with no
qualifiers


mkouba:
Merged.

PR #1575, opened 2017-01-28
WELD-2262 Introduce CompositeCDIProvider

mkouba:


impl/src/main/java/org/jboss/weld/CompositeCDIProvider.java
```diff
@@ -0,0 +1,71 @@
...
+ * Looks up {@link WeldCDIProvider} service providers and sorts them using {@link WeldCDIProvider#getPriority()}. Providers with higher priority are taken
+ * first.
+ * <p>
+ * See also <a href= "https://issues.jboss.org/browse/WELD-2262">https://issues.jboss.org/browse/WELD-2262</a>.
+ *
+ * @author Martin Kouba
+ */
+public class CompositeCDIProvider implements CDIProvider {
+
+    private volatile List<WeldCDIProvider> providers;
```
manovotn: Why do you use `volatile` here? It seems redundant given the synchronized block in `providers()` method. I guess I am missing something though.

mkouba: It's not reundant - see L52. If a thread T1 writes the `providers` the sync block ensures the value is propagated to the main memory. However, if `providers` was not volatile a thread T2 might read the value from "cache", ie. `null` and attempt to find them again. It wouldn't be a big problem but I think the current solution is better :-)

mkouba:
@tremes Weld SE/Servlet cooperation tests are not run by default. AFAIK there was a special job for this...

mkouba:
Merged.

PR #1579, opened 2017-01-31
Weld 2319

tremes:


tremes:
Feel free to suggest better `scanning information` UI formatting or displaying.

probe/core/src/main/client/probe.html
```diff
@@ -380,6 +389,44 @@
                         </div>
                     </div>
                     {{/if}}
+                    <div class="form-group">
+                        <label class="col-sm-2 control-label">Scanning information:</label>
```
manovotn: Maybe we should only display this section if there is some content? Or have default message when there is nothing to display (`No scanning information available`)?

mkouba: I'm working on it ;-)

mkouba:
@tremes @manovotn I've added few things... pls review.

mkouba:
Merged, thanks!

PR #1582, opened 2017-02-02
Increase container connection timeout for SE tests. Windows machines …

manovotn:
…were failing.

I came across a (random) timeout on Win machine during release testing. I think it will do no harm to increase the timeout here.

### DIFF
--- a/environments/se/tests/src/test/resources/arquillian.xml
+++ b/environments/se/tests/src/test/resources/arquillian.xml
@@ -15,7 +15,7 @@
          <property name="logLevel">INFO</property>
          <property name="keepDeploymentArchives">false</property>
          <property name="additionalJavaOpts">${surefire.plugin.jdk9.args} ${jacoco.agent}</property>
-         <property name="waitTime">15</property>
+         <property name="waitTime">${connection.wait.time:15}</property>
       </configuration>
    </container>
 


tremes:
I would rather introduce configurable variable so that you can pass value only for related jobs.

manovotn:
Hmm, that's also an option, but that would require modifications to jenkins job. I mean add OS specific bash code. While it's easy to do, I don't like that approach for maintenance.
Extending total time doesn't necessarily mean the jobs will take longer, only that it will wait if need be. What is it you don't like about it?

mkouba:
I think it's fine to use the increased hardcoded value here.

tremes:
I am not saying I don't like it. I just think the variable is more universal solution. What if 25 is still not enough?

mkouba:
@tremes @manovotn `arquillian.xml` does support properties with default values, or?

tremes:
Yes I think so. We have something like that in `jboss-tck-runner/src/test/wildfly8/arquillian.xml`

mkouba:
+1 Yep, that's the way to go. I think.

manovotn:
Didn't know it supports default. Fair enough then, I will change it to variable.

mkouba:
Merged, thanks!

PR #1585, opened 2017-02-06
Allow to pass jbossArguments to Arquillian properties in integration …

tremes:
…tests.


mkouba:
It seems WildFly Arquillian adapter does not like empty strings in configuration...

mkouba:
Merged, thanks!

PR #1587, opened 2017-02-07
WELD-2330 Improve logging in DeferredEventNotification.

manovotn:


manovotn:
If this looks alright, I'll make a PR for 2.4 as well.

modules/jta/src/main/java/org/jboss/weld/module/jta/DeferredEventNotification.java
```diff
@@ -82,7 +82,13 @@ protected void execute() {
             }.run();
 
         } catch (Exception e) {
-            EventLogger.LOG.asyncObserverFailure(metadata);
+            String stackTrace;
+            if (e.getCause().getMessage() != null) {
```
mkouba: `e.getCause()` might return null.

modules/jta/src/main/java/org/jboss/weld/module/jta/DeferredEventNotification.java
```diff
@@ -82,7 +82,13 @@ protected void execute() {
...
+                stackTrace = "StackTrace: " + e.getCause().getMessage();
+            } else {
+                stackTrace = "No stack trace to show, the exception's message was empty.";
```
mkouba: It's actually not a stack trace but a exception/detail message.

mkouba:
Merged, thanks!

PR #1593, opened 2017-02-15
WELD-2302 Wro4j maven plugin does not work with JDK 9

antoinesd:
upgrade plugin and Mockito dependency to have a Mockito version supporting Java 9

### DIFF
--- a/probe/core/pom.xml
+++ b/probe/core/pom.xml
@@ -106,7 +106,15 @@
                     <plugin>
                         <groupId>ro.isdc.wro4j</groupId>
                         <artifactId>wro4j-maven-plugin</artifactId>
-                        <version>1.7.7</version>
+                        <version>1.8.0</version>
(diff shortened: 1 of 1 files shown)


mkouba:
ok to test

mkouba:
add to whitelist

manovotn:
Looks OK, though we cannot be sure this will make it into wro plugin itself (they are using prehistoric `2.0.42-beta` there). Therefore, we cannot be sure this won't effect the plugin's functionality in some weird way.
Although as far as tests are concerned, it looks alright. 

@antoinesd please create similar PR for 2.4 branch

antoinesd:
Well @manovotn, Mockito is used only in 2 methods in the code so it should be easy to test a possible issue.
Will backport to 2.4 as well.

mkouba:
Merged, thanks!

PR #1597, opened 2017-02-17
WELD-2338 modify the condition recognizing OM as container lifecycle …

manovotn:
…one.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/util/Observers.java
+++ b/impl/src/main/java/org/jboss/weld/util/Observers.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Method;
 import java.security.AccessController;
 import java.util.Set;
 
+import javax.enterprise.inject.Any;
 import javax.enterprise.inject.spi.AfterBeanDiscovery;
 import javax.enterprise.inject.spi.AfterDeploymentValidation;
 import javax.enterprise.inject.spi.AfterTypeDiscovery;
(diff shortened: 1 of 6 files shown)


impl/src/main/java/org/jboss/weld/util/Observers.java
```diff
@@ -84,7 +85,8 @@ private Observers() {
 
     public static boolean isContainerLifecycleObserverMethod(ObserverMethod<?> method) {
         return CONTAINER_LIFECYCLE_EVENT_TYPES.contains(Reflections.getRawType(method.getObservedType()))
-                || (Object.class.equals(method.getObservedType()) && method instanceof ContainerLifecycleEventObserverMethod);
+                || (Object.class.equals(method.getObservedType()) && method.getObservedQualifiers().size() == 1
```
mkouba: `void observe(@Observes Object event, Foo foo)` has no observed qualifiers so it would NOT be recognized as a container lifecycle event - that's not correct. Also `org.jboss.weld.tests.extensions.validation.WildcardObserverTest` accidentally passes (the deployment fails with "_org.jboss.weld.exceptions.IllegalArgumentException: WELD-001456: Argument resolvedBean must not be null_").

manovotn: I see. Well, when I adjusted the condition to accept this observer as well, I am getting a `DefinitionException` saying that CLE observer cannot inject anything else than BM. I'll take a closer look into it.

manovotn: @mkouba I am aware that the above behaviour is per spec, however I am not confused if we want to support this in a non-portable fashion or not? The earlier conversation around this issue gave me the impression that we do?

manovotn:
Alright, think I finally understood it, pushed some changes which should do the trick.

manovotn:
@mkouba how do you like the condition now? :-P

mkouba:
@manovotn Perfect!

mkouba:
Merged, thanks!

PR #1599, opened 2017-02-20
Weld 2335 Having different types between InterceptionFactory and instance to intercept should fail

antoinesd:
As the idea behind InterceptionFactory is to customise the underlying AnnotatedType to add interceptor binding on class or method, it doesn't make sense to pass an instance of a different type than the parameter type used to defined the InterceptionFactory.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
+++ b/impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
@@ -112,6 +112,10 @@ public class InterceptionFactoryImpl<T> implements InterceptionFactory<T> {
         }
         used = true;
 
+        if (annotatedType.getJavaClass().isInterface()) {
+            throw InterceptorLogger.LOG.interceptionFactoryNotOnInstance(annotatedType.getJavaClass().getCanonicalName());
+        }
+
         Optional<InterceptionFactoryData<T>> cached = beanManager.getServices().get(InterceptionFactoryDataCache.class)
                 .getInterceptionFactoryData(configurator != null ? configurator.complete() : annotatedType);
 
--- a/tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
+++ b/tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
@@ -18,8 +18,10 @@
 package org.jboss.weld.tests.interceptors.producer;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 
 import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 
 import javax.enterprise.inject.Instance;
@@ -29,6 +31,7 @@ import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.BeanArchive;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.weld.exceptions.IllegalStateException;
 import org.jboss.weld.test.util.Utils;
 import org.jboss.weld.tests.interceptors.producer.Producer.Bar;
 import org.jboss.weld.tests.interceptors.producer.Producer.Foo;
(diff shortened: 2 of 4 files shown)


mkouba:
Well, sometimes it might make sense to pass a subtype instance - consider this dummy producer:
```
@Produces
void Number produceNumber(InterceptionFactory<Number> factory) {
   // MyInterceprotBinding has @Inherited
   factory.configure().add(MyInterceprotBinding.Literal.INSTANCE);
   return factory.createInterceptedInstance(isLong ? Long.valueOf("1") : Short.valueOf());
}
```

manovotn:
First of all, I have to say I feel like this functionality is very limiting. Not being able to use interfaces sounds like a bad practice. Not to mention it's completely counter-intuitive, so we should at the very least point this out somewhere in docs.

Anyhow, [here is another test case ](https://github.com/manovotn/core/blob/weld2335_anotherCase/tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/Producer.java#L93)in which the graceful failure fails - you get `NoSuchElementException` there. It is a plain bean implementing an interface and I tried using InterceptionFactory on the interface.

antoinesd:
@manovotn this feature is here to apply interceptor binding on an instance dynamically.
When you use interceptors on the standard way, applying an interceptor binding to an interface doesn't produce anything, so failing here is somehow compliant with interceptor spec.
Will look at your new test.

I agree that the question of supporting parent class is open, even if interceptor binding are only applied on subclass if the annotation is `@Inherited` and applied on method if they are not overloaded in subclass...

antoinesd:
@mkouba I update the PR to support creation of intercepted instance on child class instance.

Technically we could support interfaces (by creating a new AnnotatedType and copying annotations), but is it worth?

mkouba:
@antoinesd I think it is more a spec issue/question for EG. My opinion is that interfaces should not be supported. And I don't think it a bad practice or counter-intuitive - `InterceptionFactory` is a helper for a produced instance - implementing class. The bean type could be an interface.

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
```diff
@@ -88,4 +91,22 @@ public void testMapPut(@Produced Map<String, Object> map) {
         assertEquals(1, Producer.INVOCATIONS.size());
         assertEquals(Arrays.toString(params), Producer.INVOCATIONS.get(0));
     }
+    
+    @Test(expected = IllegalStateException.class)
+    public void testListAdd(BeanManager bm) {
+        // resolving via Instance just to make the NPE exception human-readable (direct method injection will blow up with Arq. stack)
+        // Invalid producer using an InterceptionFactory for List (interface) and applying it to ArrayList
+        List<Object> list = bm.createInstance().select(new TypeLiteral<List<Object>>() {}, Produced.Literal.INSTANCE).get();
```
mkouba: Just a minor thing... why no use `@Produced Instance<List<Object>> isntance` parameter instead of `BeanManager bm`?

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
```diff
@@ -88,4 +91,22 @@ public void testMapPut(@Produced Map<String, Object> map) {
...
+        Producer.reset();
```
mkouba: These lines should be replaced with `Assert.fail()`...

impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
```diff
@@ -112,6 +112,10 @@ public T createInterceptedInstance(T instance) {
         }
         used = true;
 
+        if(annotatedType.getJavaClass().isInterface()) {
```
mkouba: Please use the predefined formatting style...

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
manovotn: Because it was the example test I had on my branch and I did it this is way to make the exception human-readable (as the comments suggest). Otherwise you got an NPE coming from arquillian.
I agree it should be either changed now, or at least the comment is now useless.

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
manovotn: +1, since this is now deeded as expected failure

mkouba:
@antoinesd could you please squash the last 3 commits please?

mkouba:
Merged, thanks!

PR #1600, opened 2017-02-21
WELD-2340 Fire ContainerInitialized event after ShutdownHook registra…

manovotn:
…tion.

This change should be sufficient to fix the behaviour, yet I am still struggling to write an automated test for this.
Putting on hold till we have some test.

### DIFF
--- a/environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
+++ b/environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
@@ -166,6 +166,7 @@ public class WeldContainer extends AbstractCDI<Object> implements AutoCloseable,
                 }
             }
         }
+        container.fireContainerInitializedEvent();
     }
 
     /**
@@ -209,8 +210,11 @@ public class WeldContainer extends AbstractCDI<Object> implements AutoCloseable,
         this.creationalContext = beanManager().createCreationalContext(null);
         this.instance = beanManager().getInstance(creationalContext);
         this.event = beanManager().event();
-        beanManager().fireEvent(new ContainerInitialized(id), Initialized.Literal.APPLICATION);
+    }
+
+    private void fireContainerInitializedEvent() {
         WeldSELogger.LOG.weldContainerInitialized(id);
+        beanManager().fireEvent(new ContainerInitialized(id), Initialized.Literal.APPLICATION);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
```diff
@@ -209,6 +210,9 @@ private void complete() {
         this.creationalContext = beanManager().createCreationalContext(null);
         this.instance = beanManager().getInstance(creationalContext);
         this.event = beanManager().event();
+    }
+
+    private void fireContainerInitializedEvent() {
         beanManager().fireEvent(new ContainerInitialized(id), Initialized.Literal.APPLICATION);
         WeldSELogger.LOG.weldContainerInitialized(id);
```
mkouba: Hm, don't move the message - it should be logged before the event, otherwise it will never appear if a `ContainerInitialized` observer blocks...

manovotn: I didn't move it. The logging [was **after**](https://github.com/weld/core/blob/master/environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java#L213)  the event.
But you are right, that it makes sense to swap those lines.

mkouba: Well yes, I meant: only `beanManager().fireEvent()...` should have been moved after shutdown hook ;-)

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/shutdown/hook/ShutdownHookRegisteredBeforeInitEventTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * This will throw a nasty exception because we used System.exit() to simulate forcible shutdown
+     */
+    @Override
+    @InSequence(2)
+    @Test(expected = UndeclaredThrowableException.class)
+    public void testFooPing() {
+        // change logging level, does NOT work
+        Logger logger = Logger.getLogger(JMXMethodExecutor.class.getName());
```
mkouba: The logging actually happens on the "client" side. So we'll need to setup the log level there, maybe using `@BeforeClass` and `@AfterClass`? I tried to move this code to the `AbstractShutdownHookTest.deploy()` method and it seemed to work.

manovotn:
In this state, it should have even (a bit hardcore, log surpressing). Please review :)

mkouba:
Merged, thanks!

PR #1603, opened 2017-02-24
WELD-2341 fix injection of InjectionPoint metadata to singleton session

tremes:
bean.


modules/ejb/src/main/java/org/jboss/weld/module/ejb/InjectionPointPropagatingEnterpriseTargetBeanInstance.java
```diff
@@ -41,12 +41,12 @@
 
     private final InjectionPointHolder injectionPointHolder;
     private final String contextId;
-    private transient SLSBInvocationInjectionPoint slsbInvocationInjectionPoint;
+    private transient CurrentInvocationInjectionPoint nonSFSBInvocationInjectionPoint;
```
mkouba: `currentInvocationInjectionPoint` would be a probably better field name...

mkouba:
Merged, thanks!

PR #1604, opened 2017-02-24
WELD-2337 OSGI bundle exports need to export modules

antoinesd:
correct modified packages for OSGi manifest generation


mkouba:
Merged, thanks!

PR #1606, opened 2017-02-27
WELD-2331 Misleading error message when Extension observer method is (wrongly) static

antoinesd:
Added specific deployment error and test for it
Activated WeldExceptionTransformer to test specific error messages at deployment

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/ExtensionBeanDeployer.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/ExtensionBeanDeployer.java
@@ -44,7 +44,6 @@ import org.jboss.weld.util.BeanMethods;
 import org.jboss.weld.util.DeploymentStructures;
 import org.jboss.weld.util.Observers;
 import org.jboss.weld.util.reflection.Formats;
-import org.jboss.weld.util.reflection.Reflections;
 
 /**
  * @author pmuir
@@ -149,8 +148,13 @@ public class ExtensionBeanDeployer {
             Set<ObserverInitializationContext<?, ?>> observerMethodInitializers, boolean isAsync) {
         ObserverMethodImpl<T, X> observer = ObserverFactory.create(method, declaringBean, beanManager, isAsync);
         ObserverInitializationContext<T, X> observerMethodInitializer = ObserverInitializationContext.of(observer, method);
-        if (isAsync && Observers.CONTAINER_LIFECYCLE_EVENT_TYPES.contains(Reflections.getRawType(observer.getObservedType()))) {
-            throw EventLogger.LOG.asyncContainerLifecycleEventObserver(observer, Formats.formatAsStackTraceElement(method.getJavaMember()));
+        if (Observers.isContainerLifecycleObserverMethod(observer)) {
+            if (isAsync) {
+                throw EventLogger.LOG.asyncContainerLifecycleEventObserver(observer, Formats.formatAsStackTraceElement(method.getJavaMember()));
+            }
+            if (method.isStatic()) {
+                throw EventLogger.LOG.staticContainerLifecycleEventObserver(observer, Formats.formatAsStackTraceElement(method.getJavaMember()));
+            }
         }
         observerMethodInitializers.add(observerMethodInitializer);
     }
--- a/impl/src/main/java/org/jboss/weld/logging/EventLogger.java
+++ b/impl/src/main/java/org/jboss/weld/logging/EventLogger.java
@@ -83,7 +83,7 @@ public interface EventLogger extends WeldLogger {
     @Message(id = 413, value = "{0} cannot be replaced by an observer method with a different bean class {1}", format = Format.MESSAGE_FORMAT)
     DefinitionException beanClassMismatch(ObserverMethod<?> originalObserverMethod, ObserverMethod<?> observerMethod);
 
-    @Message(id = 414, value = "Observer method for container lifecycle events cannot be asynchronous. {0}\n\tat {1}\n  StackTrace:", format = Format.MESSAGE_FORMAT)
+    @Message(id = 414, value = "Observer method for container lifecycle event cannot be asynchronous. {0}\n\tat {1}\n  StackTrace:", format = Format.MESSAGE_FORMAT)
     DefinitionException asyncContainerLifecycleEventObserver(ObserverMethod<?> observer, Object stackElement);
 
     @Message(id = 415, value = "Custom implementation of observer method does not override either notify(T) or notify(EventContext<T>): {0}", format = Format.MESSAGE_FORMAT)
@@ -95,5 +95,8 @@ public interface EventLogger extends WeldLogger {
     @LogMessage(level = Level.WARN)
     @Message(id = 417, value = "The original observed type {0} is not assignable from {1} set by extension {2} - the observer method invocation may result in runtime exception!", format = Format.MESSAGE_FORMAT)
     void originalObservedTypeIsNotAssignableFrom(Object originalObservedType, Object observedType, Object extension);
+
+    @Message(id = 418, value = "Observer method for container lifecycle event cannot be static. {0}\n\tat {1}\n  StackTrace:", format = Format.MESSAGE_FORMAT)
+    DefinitionException staticContainerLifecycleEventObserver(ObserverMethod<?> observer, Object stackElement);
 
 }
--- a/tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/ExtensionWithStaticLifecycleEventObserver.java
+++ b/tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/ExtensionWithStaticLifecycleEventObserver.java
@@ -1,0 +1,32 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.weld.tests.extensions.staticObserver;
+
+import javax.enterprise.event.Observes;
+import javax.enterprise.inject.spi.Extension;
+import javax.enterprise.inject.spi.ProcessAnnotatedType;
+
+/**
+ * @author Antoine Sabot-Durand
+ */
+public class ExtensionWithStaticLifecycleEventObserver implements Extension {
+
+    static void processAnnotatedType(@Observes ProcessAnnotatedType<?> pat) {
+
+    }
+}
--- a/tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
+++ b/tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
@@ -1,0 +1,53 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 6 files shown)


impl/src/main/java/org/jboss/weld/bootstrap/ExtensionBeanDeployer.java
```diff
@@ -149,8 +149,13 @@ public void addExtension(Metadata<? extends Extension> extension) {
             Set<ObserverInitializationContext<?, ?>> observerMethodInitializers, boolean isAsync) {
         ObserverMethodImpl<T, X> observer = ObserverFactory.create(method, declaringBean, beanManager, isAsync);
         ObserverInitializationContext<T, X> observerMethodInitializer = ObserverInitializationContext.of(observer, method);
-        if (isAsync && Observers.CONTAINER_LIFECYCLE_EVENT_TYPES.contains(Reflections.getRawType(observer.getObservedType()))) {
-            throw EventLogger.LOG.asyncContainerLifecycleEventObserver(observer, Formats.formatAsStackTraceElement(method.getJavaMember()));
+        if (Observers.CONTAINER_LIFECYCLE_EVENT_TYPES.contains(Reflections.getRawType(observer.getObservedType()))) {
```
mkouba: Actually, I think we should replace `CONTAINER_LIFECYCLE_EVENT_TYPES.contains()` with `Observers.isContainerLifecycleObserverMethod()`...

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
```diff
@@ -0,0 +1,51 @@
...
+public class StaticLifecycleEventObserverTest {
+    @Deployment
+    @ShouldThrowException(StaticLifecycleEventObserverForTestException.class)
+    public static Archive<?> deploy() {
+        return ShrinkWrap.create(BeanArchive.class, Utils.getDeploymentNameAsHash(StaticLifecycleEventObserverTest.class))
+                .addClass(ExtensionWithStaticLifecycleEventObserver.class).addAsServiceProvider(Extension.class, ExtensionWithStaticLifecycleEventObserver.class);
+    }
+
+    @Test
+    public void testDummy() {
```
manovotn: The name of the test method should reflect what it tests or what is the excepted outcome. E.g.  something like `testStaticObserverFailsDeployment` or `testStaticLifecycleObserver`.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Testing that a deployment error occurs extension has static lifecycle event observer method
+ *
+ * @author Antoine Sabot-Durand
+ */
+@RunWith(Arquillian.class)
+public class StaticLifecycleEventObserverTest {
+    @Deployment
+    @ShouldThrowException(StaticLifecycleEventObserverForTestException.class)
```
mkouba: I think this wouldn't work for `incontainer` tests (`WildFly8DeploymentExceptionTransformer`). We should only check a `DefinitionException` is thrown.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
```diff
@@ -0,0 +1,51 @@
...
+    public static Archive<?> deploy() {
+        return ShrinkWrap.create(BeanArchive.class, Utils.getDeploymentNameAsHash(StaticLifecycleEventObserverTest.class))
+                .addClass(ExtensionWithStaticLifecycleEventObserver.class).addAsServiceProvider(Extension.class, ExtensionWithStaticLifecycleEventObserver.class);
+    }
+
+    @Test
+    public void testDummy() {
+        Assert.assertTrue(true);
```
manovotn: What is the point of this line? You might as well leave it blank, or, better still, only leave a comment `// should throw exception`.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverForTestException.java
```diff
@@ -0,0 +1,28 @@
...
+package org.jboss.weld.tests.extensions.staticObserver;
+
+/**
+ *
+ * Exception to distinguish {@link javax.enterprise.inject.spi.DeploymentException} related to Static Lifecycle event observer
+ * method in test.
+ *
+ * @author Antoine Sabot-Durand
+ */
+public class StaticLifecycleEventObserverForTestException extends RuntimeException {
```
manovotn: Thinking about this. it is maybe an overkill to test it like this? I think it would be sufficient to have a test asserting that:

1. Static container lifecycle observer throws definition exception
2. Other static observers in extensions work as expected

@tremes WDYT?

impl/src/main/java/org/jboss/weld/logging/EventLogger.java
```diff
@@ -96,4 +96,7 @@
     @Message(id = 417, value = "The original observed type {0} is not assignable from {1} set by extension {2} - the observer method invocation may result in runtime exception!", format = Format.MESSAGE_FORMAT)
     void originalObservedTypeIsNotAssignableFrom(Object originalObservedType, Object observedType, Object extension);
 
+    @Message(id = 418, value = "Observer method for container lifecycle events cannot be static. {0}\n\tat {1}\n  StackTrace:", format = Format.MESSAGE_FORMAT)
```
manovotn: Just a minor thing here:
`events` -> `event`

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverForTestException.java
mkouba: +1, see also https://github.com/weld/core/pull/1606#pullrequestreview-23955400 

I would completely remove this exception and revert `WeldExceptionTransformer` modifications. Actually, `WeldExceptionTransformer` could be possibly removed - it's not used.

tremes: Yes it's not necessary. Only `DefinitionException` should be checked.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
antoinesd: Then test would be totally useless since we are creating it to test the content of `DeploymentException` message (which is already thrown with misleading message). I would rather update `WildFly8DeploymentExceptionTransformer`

impl/src/main/java/org/jboss/weld/bootstrap/ExtensionBeanDeployer.java
antoinesd: ok

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
antoinesd: This method will never be called and is only here to have JUnit pick up the test. `willNeverBeCalled becauseDeploymentWillFail` is probably a more accurate name.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
antoinesd: +1 for a blank line, `// Should throw exception` is misleading.

impl/src/main/java/org/jboss/weld/logging/EventLogger.java
mkouba: Hm, the same wording is used in `EventLogger.asyncContainerLifecycleEventObserver()`. We should fix the message there as well ;-)

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverForTestException.java
antoinesd: Yes I hesitated to write it. I thought about false positive.

antoinesd: guys, could you then explain me the interest of the test if we don't check the exception message? `DefinitionException` is already thrown without this PR.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
mkouba: I don't think so. The "misleading" exception is `IllegalStateException` whereas the new one is `DefinitionException`. Anyway, we don't test exception messages because it's not possible to create a reliable `DeploymentExceptionTransformer` for all container adapters. Even `WildFly8DeploymentExceptionTransformer` contains some hacks just to support `DefinitionException` and `DeploymentException`.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
manovotn: I know it won't be invoked, but the point is to make it obvious what the test does by just glancing at the code. And `testDummy` does very little in that regard.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverForTestException.java
manovotn: From what I gather, it was throwing `IllegalStateException` before, not `DefinitionException`.
Anyhow, the point is to have a pair of tests proving that CLE observers cannot be static and any other observers in extension can.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
antoinesd: Without ExtensionBeanDeployer modification Stack trace contains `IllegalStateException` but root exception is:

`org.jboss.weld.exceptions.DefinitionException: WELD-000143: Container lifecycle event method invoked outside of extension observer method invocation.`

So having detection on `DefinitionException` will have the test pass with and without code modification.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverForTestException.java
antoinesd: During default maven build `IllegalStateException` is the root cause but `DefinitionException` is at the top of stack trace see https://github.com/weld/core/pull/1606#discussion_r103184460.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverTest.java
mkouba: Hm, the difference is that the original exception (invoked outside of extension...) is only thrown if a container lifecycle event method is invoked (e.g. `ProcessAnnotatedType.getAnnotatedType()`). After this modification, an exception must be thrown even if no method is invoked - having a static observer should be enough.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/StaticLifecycleEventObserverForTestException.java
mkouba: Antoine, we're not going to test the message contents because of aforementioned problems (Arquillian-related) and the fact, that it would imply additional maintanance effort.

antoinesd: ok, your the boss @mkouba.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/staticObserver/ExtensionWithStaticLifecycleEventObserver.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.enterprise.inject.spi.Extension;
+import javax.enterprise.inject.spi.ProcessAnnotatedType;
+
+/**
+ * @author Antoine Sabot-Durand
+ */
+public class ExtensionWithStaticLifecycleEventObserver implements Extension {
+
+    static void processAnnotatedType(@Observes ProcessAnnotatedType<?> pat) {
+        pat.veto();
```
mkouba: Please remove this invocation. Right now, the test would pass even without your `ExtensionBeanDeployer` modification. If you remove the invocation, the test would fail without the modification.

Otherwise looks good now.

mkouba:
Looks good now! Please create a PR for 2.4 branch.

antoinesd:
done with #1608

mkouba:
Merged, thanks!

PR #1607, opened 2017-02-27
WELD-2342 Probe - fix unused beans

mkouba:
- do not mark a bean as unused if only injected into an observer method
or disposer method


mkouba:
Merged.

PR #1608, opened 2017-02-28
WELD-2331 for 2.4 Misleading error message when Extension observer method is (wrongly) static

antoinesd:
Added specific deployment error and test for it


mkouba:
Merged, thanks!

PR #1610, opened 2017-03-01
WELD-2343 Fix bean identifier index

mkouba:
- do not include built-in session and conversation scoped beans


mkouba:
We should run some replication/failover tests before merging.

mkouba:
Merged.

PR #1614, opened 2017-03-02
WELD-2346 Fix HttpSessionBean

mkouba:


mkouba:
> or have you done that already?

No. Hopefully, it will work :-)

mkouba:
> Shouldn't we try to create some tests to validate it ?

Good point. I will try to add a test for eager `HttpSession` initialization. There are some replication tests in the WildFly test suite.

mkouba:
When writing a test I realized that an invocation of a injected `HttpSession` actually triggers an HTTP session creation because weld creates a new session after a bean instance of a `@SessionScoped` bean is created (and the built-in bean for `HttpSession` is `@SessionScoped`). So I will add the test but it would pass even without the fix from this PR.

antoinesd:
I'm still curious about reproducing the case where HttpSession is null.

mkouba:
@antoinesd Ok, I will try to add a test for this as well ;-)

mkouba:
Actually, the problem with null session is also not valid. For the same reason: https://github.com/weld/core/pull/1614#issuecomment-283956896

manovotn:
Verified manually - works like a charm.

PR #1616, opened 2017-03-08
WELD-2349 throw definition exception if there is a stereotype on obse…

manovotn:
…rver method.


impl/src/main/java/org/jboss/weld/logging/EventLogger.java
```diff
@@ -99,4 +99,6 @@
     @Message(id = 418, value = "Observer method for container lifecycle event cannot be static. {0}\n\tat {1}\n  StackTrace:", format = Format.MESSAGE_FORMAT)
     DefinitionException staticContainerLifecycleEventObserver(ObserverMethod<?> observer, Object stackElement);
 
+    @Message(id = 419, value = "Observer method cannot be annotated with a stereotype: {0}\n\tat {1}\n  StackTrace:", format = Format.MESSAGE_FORMAT)
```
tremes: I would use `with a stereotype annotation`.

mkouba:
See also https://issues.jboss.org/browse/WELD-2349?focusedCommentId=13376522&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13376522

PR #1618, opened 2017-03-14
WELD-2348 Correct the information about default thread pool used.

manovotn:


docs/reference/src/main/asciidoc/configure.asciidoc
```diff
@@ -84,7 +84,7 @@ might be ignored. An example of such integrator is WildFly.
 [cols=",,",options="header",]
 |=======================================================================
 |Configuration key |Default value |Description
-|`org.jboss.weld.executor.threadPoolType` |`FIXED` |The type of the thread pool. Possible values
+|`org.jboss.weld.executor.threadPoolType` |`COMMON` |The type of the thread pool. Possible values
```
mkouba: That's not correct - `COMMON` is the default for Weld SE only. It should be something like:
> FIXED (COMMON in Weld SE)

docs/reference/src/main/asciidoc/configure.asciidoc
```diff
@@ -72,7 +72,7 @@ First of all, let's see what types of thread pools are available:
 |`FIXED_TIMEOUT`|Uses a fixed number of threads. A thread will be stopped after a configured period of inactivity.
 |`SINGLE_THREAD`|A single-threaded thread pool
 |`NONE`|No executor is used by Weld
-|`COMMON`|The default ForkJoinPool.commonPool() is used by Weld. See https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html#commonPool--[link] for more details
+|`COMMON` (default)|The default ForkJoinPool.commonPool() is used by Weld. See https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html#commonPool--[link] for more details
```
johnament: Same here - default in SE.

PR #1620, opened 2017-03-15
WELD-2140 log a warning when there's lifecycle callback method on tar…

tremes:
…get class with one param.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/logging/ValidatorLogger.java
+++ b/impl/src/main/java/org/jboss/weld/logging/ValidatorLogger.java
@@ -273,8 +273,9 @@ public interface ValidatorLogger extends WeldLogger {
     @Message(id = 1468, value = "Method {0} defined on class {1} is not defined according to the specification. It is annotated with @{2} but it does not have a {3} return type.\n\tat {4}\n  StackTrace", format = Format.MESSAGE_FORMAT)
     DefinitionException interceptorMethodDoesNotHaveVoidReturnType(Object param1, Object param2, Object param3, Object param4, Object stackElement);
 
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/jboss/weld/logging/ValidatorLogger.java
```diff
@@ -273,8 +273,9 @@ public void log(Object... params) {
     @Message(id = 1468, value = "Method {0} defined on class {1} is not defined according to the specification. It is annotated with @{2} but it does not have a {3} return type.\n\tat {4}\n  StackTrace", format = Format.MESSAGE_FORMAT)
     DefinitionException interceptorMethodDoesNotHaveVoidReturnType(Object param1, Object param2, Object param3, Object param4, Object stackElement);
 
-    @Message(id = 1469, value = "Method {0} defined on class {1} is not defined according to the specification. It is annotated with @{2} but it does not have zero parameters.\n\tat {3}\n  StackTrace", format = Format.MESSAGE_FORMAT)
-    DefinitionException interceptorMethodDoesNotHaveZeroParameters(Object param1, Object param2, Object param3, Object stackElement);
+    @LogMessage(level = Level.WARN)
+    @Message(id = 1469, value = "Method {0} defined on class {1} is not defined according to the specification. It is annotated with @{2} but it does not have zero parameters.\n", format = Format.MESSAGE_FORMAT)
```
mkouba: Why did you remove the stackElement part?

mkouba: Ah, it's just a warning. Then ok ;-)

tremes: There is no stackTrace. Do you think it's worth of telling the line number of the corresponding method? I thought the name of the method is sufficient.

mkouba: Yep, it is. https://github.com/weld/core/pull/1620#discussion_r106180968

PR #1625, opened 2017-03-20
WELD-2350 Add findbugs check to Probe.

manovotn:


probe/core/src/main/java/org/jboss/weld/probe/InvocationMonitor.java
```diff
@@ -86,7 +86,7 @@
 
     @Intercepted
     @Inject
-    private Bean<?> interceptedBean;
+    private transient Bean<?> interceptedBean;
```
mkouba: -1. A built-in bean which is a passivation capable dependency (as defined by the spec) is injected.

probe/core/src/main/java/org/jboss/weld/probe/JsonObjects.java
```diff
@@ -373,7 +373,7 @@ public int compare(ConfigurationKey o1, ConfigurationKey o2) {
                     }
                     desc = description.value();
                 }
-            } catch (Exception e) {
+            } catch (NoSuchFieldException e) {
```
mkouba: I think the original intent was to ignore any exception thrown, incl. runtime exceptions...

probe/core/src/main/java/org/jboss/weld/probe/Resource.java
```diff
@@ -251,7 +251,7 @@ protected void handle(HttpMethod method, JsonDataProvider jsonDataProvider, Stri
 
     private final String[] parts;
 
-    private final Handler handler;
+    private final transient Handler handler;
```
mkouba: `Handler` instances do not hold any state so it might be better to `Handler implements Serializable` instead?

probe/core/src/main/java/org/jboss/weld/probe/InvocationMonitor.java
manovotn: How can I know that the injected instance is always serializable? 
The `Bean` interface does not require this and the serializability (if that word exists) depends on the type parameter. Or not?

probe/core/src/main/java/org/jboss/weld/probe/JsonObjects.java
manovotn: Well, I have no recollection of original intents :) 
I'll revert it and use a `@SuppressWarnings` then.

probe/core/src/main/java/org/jboss/weld/probe/InvocationMonitor.java
mkouba: It is injected by the container and the spec requires the beans to be passivation capable deps, see also http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#bean_metadata and `org.jboss.weld.bean.builtin.BeanMetadataBean.newInstance(InjectionPoint, CreationalContext<Bean<?>>)`.

probe/core/src/main/java/org/jboss/weld/probe/Resource.java
manovotn: Fair enough, will do.

probe/core/src/main/java/org/jboss/weld/probe/JsonObjects.java
mkouba: I know :-). My point is that we should not probably change the original functionality...

probe/core/src/main/java/org/jboss/weld/probe/InvocationMonitor.java
manovotn: Ah, I see. Thanks for pointing that bit out. Fix incoming.

manovotn:
Alright, added a commit which updated FB annotations dependency. Therefore, I changed all occurrences of FB's `SuppressWarnings` to `SuppressFBWarnings` (both would work, though the former one is considered deprecated due to name clash with `java.lang.SuppressWarnings`)

PR #1629, opened 2017-03-23
WELD-2354 Provide a toString() for ContainerInitialized/Shutdown even…

manovotn:
…ts with container ID.

### DIFF
--- a/environments/se/core/src/main/java/org/jboss/weld/environment/se/events/WeldContainerEvent.java
+++ b/environments/se/core/src/main/java/org/jboss/weld/environment/se/events/WeldContainerEvent.java
@@ -33,4 +33,8 @@ public abstract class WeldContainerEvent {
         return containerId;
     }
 
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + " event for container with ID: " + getContainerId();
+    }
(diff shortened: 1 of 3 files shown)


environments/se/core/src/main/java/org/jboss/weld/environment/se/events/ContainerInitialized.java
```diff
@@ -37,4 +37,8 @@ public ContainerInitialized(String containerId) {
         super(containerId);
     }
 
+    @Override
+    public String toString() {
+        return ContainerInitialized.class.getSimpleName() + " event for container with ID: " + getContainerId();
```
mkouba: We can probably declare `toString()` only for `WeldContainerEvent`...

manovotn: Depends what we want to return. If it's only container ID, then I agree.

mkouba: `return getClass().getSimpleName() + " event for container with ID: " + getContainerId();`

PR #1631, opened 2017-03-24
Fix numberguess Tomcat profile

tremes:


examples/jsf/numberguess/pom.xml
```diff
@@ -124,12 +124,6 @@
             <id>tomcat</id>
             <dependencies>
                 <dependency>
-                    <groupId>org.jboss.spec.javax.annotation</groupId>
-                    <artifactId>jboss-annotations-api_1.2_spec</artifactId>
-                    <scope>compile</scope>
-                </dependency>
-
-                <dependency>
```
tremes: We should probably add `javax.annotation` dependency instead

tremes:
Closed in favor of https://github.com/weld/core/pull/1641

PR #1633, opened 2017-03-24
WELD-2357 fallback in JandexDiscoveryStrategy when there is no ClassI…

tremes:
…nfo available.

### DIFF
--- a/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/jandex/JandexDiscoveryStrategy.java
+++ b/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/jandex/JandexDiscoveryStrategy.java
@@ -38,10 +38,10 @@ import org.jboss.weld.environment.deployment.WeldBeanDeploymentArchive;
 import org.jboss.weld.environment.deployment.discovery.AbstractDiscoveryStrategy;
 import org.jboss.weld.environment.deployment.discovery.BeanArchiveBuilder;
 import org.jboss.weld.environment.deployment.discovery.DiscoveryStrategy;
+import org.jboss.weld.environment.util.Reflections;
 import org.jboss.weld.resources.spi.ClassFileServices;
 import org.jboss.weld.resources.spi.ResourceLoader;
 import org.jboss.weld.util.collections.ImmutableSet;
-import org.jboss.weld.util.reflection.Reflections;
 
 /**
  * An implementation of {@link DiscoveryStrategy} that is used when the jandex is available.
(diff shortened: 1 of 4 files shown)


impl/src/main/java/org/jboss/weld/util/reflection/Reflections.java
```diff
@@ -493,4 +499,36 @@ public static boolean isDefault(Method method) {
         return ((method.getModifiers() & (Modifier.ABSTRACT | Modifier.PUBLIC | Modifier.STATIC)) == Modifier.PUBLIC)
                 && method.getDeclaringClass().isInterface();
     }
+
+    public static boolean hasBeanDefiningAnnotation(Class<?> clazz, Set<Class<? extends Annotation>> initialBeanDefiningAnnotations) {
```
mkouba: Wrong `Reflections` class ;-) It should be `org.jboss.weld.environment.util.Reflections`.

environments/common/src/main/java/org/jboss/weld/environment/util/Reflections.java
```diff
@@ -33,6 +40,8 @@
  */
 public final class Reflections {
 
+    public static final List<Class<? extends Annotation>> metaAnnotations = ImmutableList.of(Stereotype.class, NormalScope.class);
```
mkouba: The constant name should be `META_ANNOTATIONS`

environments/common/src/main/java/org/jboss/weld/environment/util/Reflections.java
```diff
@@ -125,6 +134,8 @@ public static boolean isClassLoadable(ResourceLoader resourceLoader, String clas
             return cast(Class.forName(className));
         } catch (Exception e) {
```
mkouba: What about `catch (Exception | LinkageError e) {...`?

environments/common/src/main/java/org/jboss/weld/environment/util/Reflections.java
```diff
@@ -123,8 +132,8 @@ public static boolean isClassLoadable(ResourceLoader resourceLoader, String clas
         }
         try {
             return cast(Class.forName(className));
-        } catch (Exception e) {
-            throw CommonLogger.LOG.cannotLoadClass(className, e);
+        } catch (Exception | LinkageError e) {
+            throw new ResourceLoadingException("Error loading class " + className, e);
```
mkouba: I think we should use `CommonLogger.LOG.cannotLoadClass()` here. `org.jboss.weld.resources.spi.ResourceLoader` is related to `ResourceLoader`. Also the `cannotLoadClass()` method is only used here...

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/jandex/JandexDiscoveryStrategy.java
```diff
@@ -90,8 +90,16 @@ protected WeldBeanDeploymentArchive processAnnotatedDiscovery(BeanArchiveBuilder
...
-            if (!containsBeanDefiningAnnotation(cinfo, className)) {
-                classIterator.remove();
+            if (cinfo != null) {
+                if (!containsBeanDefiningAnnotation(cinfo, className)) {
+                    classIterator.remove();
+                }
+            } else {
+                //if ClassInfo is not available (e.g for WEB-INF/lib/jars) then fallback to reflection
+                Class<?> clazz = Reflections.loadClass(resourceLoader, className);
+                if (!Reflections.hasBeanDefiningAnnotation(clazz, initialBeanDefiningAnnotations)) {
```
mkouba: One last thing - we should add `if (clazz != null)` to avoid possible NPE...

tremes: Then it would make more sense to define it as:
`clazz == null || !Reflections.hasBeanDefiningAnnotation(clazz, initialBeanDefiningAnnotations)`
or not?

mkouba: Actually, we should follow `ReflectionDiscoveryStrategy.java `, ie.:
````
if (clazz == null || !Reflections.hasBeanDefiningAnnotation(clazz, initialBeanDefiningAnnotations)) {
   classIterator.remove();
}
````

PR #1637, opened 2017-03-27
WELD-2352 fireAsync with timeout. Requires API upgrade.

manovotn:
Will need API upgrade to compile successfully (https://github.com/weld/api/pull/58).

### DIFF
--- a/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/event/options/timeout/NotificationTimeoutTest.java
+++ b/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/event/options/timeout/NotificationTimeoutTest.java
@@ -1,0 +1,196 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.weld.environment.se.test.event.options.timeout;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import javax.enterprise.event.NotificationOptions;
+
+import org.jboss.arquillian.container.se.api.ClassPath;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.BeanArchive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.weld.bootstrap.api.CDI11Bootstrap;
+import org.jboss.weld.environment.se.Weld;
+import org.jboss.weld.environment.se.WeldContainer;
+import org.jboss.weld.events.WeldNotificationOptions;
+import org.jboss.weld.manager.api.ExecutorServices;
+import org.jboss.weld.resources.spi.ResourceLoader;
+import org.jboss.weld.test.util.Utils;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ *
+ * @author <a href="mailto:manovotn@redhat.com">Matej Novotny</a>
+ */
+@RunWith(Arquillian.class)
+public class NotificationTimeoutTest {
+
+    // has to be CLEARED in the beginning of each test method where you use it
+    public static final List<String> SUCCESSION_OF_EVENTS = new CopyOnWriteArrayList<String>();
+
+    @Deployment
+    public static Archive<?> createTestArchive() {
+        return ClassPath.builder().add(ShrinkWrap.create(BeanArchive.class, Utils.getDeploymentNameAsHash(NotificationTimeoutTest.class))
+            .addPackage(NotificationTimeoutTest.class.getPackage()).addClass(IncompleteCustomExecutorServices.class)).build();
+    }
+
+    @Test
+    public void testMultipleObserverNotificationTimeout() throws InterruptedException {
+        SUCCESSION_OF_EVENTS.clear();
+        try (WeldContainer container = new Weld().initialize()) {
+            BlockingQueue<Throwable> synchronizer = new LinkedBlockingQueue<>();
+            CountDownLatch countdown = new CountDownLatch(2);
+            container.event().select(CountDownLatch.class)
+                .fireAsync(countdown, NotificationOptions.of(WeldNotificationOptions.TIMEOUT, 2000l))
+                .exceptionally((Throwable t) -> {
+                    // handle TimeoutException we got here
+                    SUCCESSION_OF_EVENTS.add("Timeout");
+                    synchronizer.add(t);
+                    return null;
+                });
+            Throwable fromSynchronizer = synchronizer.poll(5, TimeUnit.SECONDS);
+            Assert.assertNotNull(fromSynchronizer);
+            Assert.assertTrue(fromSynchronizer.getCause().toString().contains(TimeoutException.class.getSimpleName()));
+            // wait for the observers to finish their jobs
+            countdown.await();
+            // assert contents AND order of events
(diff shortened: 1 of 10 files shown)


impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -347,9 +354,35 @@ public RuntimeException apply(Type eventType) {
             }), executor);
         }
 
-        // TODO if timeout set then wrap the completableFuture
+        // If NotificationOptionKeys.TIMEOUT is set, we will trigger the counter and use CompletableFuture.anyOf()
+        if (options.get(NotificationOptionKeys.TIMEOUT) != null) {
+            // TODO - any beter way to do the cast?
+            return new AsyncEventDeliveryStage<>((CompletableFuture<U>)CompletableFuture.anyOf(completableFuture, initiateTimer(options)), executor);
```
manovotn: This is still `TODO` but I don't know how to improve it. 
The cast itself should be harmless though - it only results to `CompletableFuture<Object>` because of the return value of `CompletableFuture.anyOf()`

impl/src/main/java/org/jboss/weld/executor/AbstractExecutorServices.java
```diff
@@ -38,6 +40,21 @@
...
+
+    /**
+     * Returns a previously created instance of ScheduledExecutorService.
+     * If there is none, a new one is created, stored and returned.
+     * @return A managed instance of ScheduledExecutorService
+     */
+    @Override
+    public ScheduledExecutorService getTimerExecutor() {
+        if (timerExecutor == null) {
+            timerExecutor = Executors.newScheduledThreadPool(1, new DaemonThreadFactory(new ThreadGroup("weld-workers"), "weld-timer-"));
```
mkouba: I would extract "weld-workers" to a constant on the `DaemonThreadFactory`.

impl/src/main/java/org/jboss/weld/logging/EventLogger.java
```diff
@@ -102,4 +102,10 @@
     @LogMessage(level = Level.WARN)
     @Message(id = 419, value = "{0} is not a valid notification mode, using the default: {1}", format = Format.MESSAGE_FORMAT)
     void invalidNotificationMode(Object mode, Object defaultMode);
+
+    @Message(id = 420, value = "The implementation of ExecutorServices has to provide an instance of SchedulesExecutorServices!", format = Format.MESSAGE_FORMAT)
+    UnsupportedOperationException noScheduledExecutorServicesProvided();
+
+    @Message(id = 421, value = "Invalid input value for observer notification timeout. Has to be parseable String. Original exception: {0}", format = Format.MESSAGE_FORMAT)
```
mkouba: Actually, `java.lang.Long` would be fine too ;-)

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -347,9 +354,35 @@ public RuntimeException apply(Type eventType) {
...
 
-        return new AsyncEventDeliveryStage<>(completableFuture, executor);
+    /**
+     * Checks if the ScheduledExecutorServices was provided.
+     * Verifies the input value for timeout.
+     * Starts the timer thread and returns it as CompletableFuture.
+     */
+    private <T> CompletableFuture<T> initiateTimer(NotificationOptions options) {
+        // throw exception if we don't have scheduled executor available
+        if (timerExecutor == null) {
```
mkouba: I would move this check right after `if (options.get(NotificationOptionKeys.TIMEOUT) != null) {`

impl/src/main/java/org/jboss/weld/logging/EventLogger.java
```diff
@@ -102,4 +102,10 @@
     @LogMessage(level = Level.WARN)
     @Message(id = 419, value = "{0} is not a valid notification mode, using the default: {1}", format = Format.MESSAGE_FORMAT)
     void invalidNotificationMode(Object mode, Object defaultMode);
+
+    @Message(id = 420, value = "The implementation of ExecutorServices has to provide an instance of SchedulesExecutorServices!", format = Format.MESSAGE_FORMAT)
```
mkouba: The message does not mention async notification timeout at all.

impl/src/main/java/org/jboss/weld/logging/EventLogger.java
manovotn: True, I was rather unsure if we should wrap the exception in `IllegalArgumentException` in the first place. The original NFE is pretty much self-explanatory in this case.

impl/src/main/java/org/jboss/weld/event/NotificationOptionKeys.java
```diff
@@ -29,6 +29,7 @@
 public final class NotificationOptionKeys {
 
     public static final String MODE = "org.jboss.weld.notification.mode";
+    public static final String TIMEOUT = "org.jboss.weld.notifications.timeout";
```
tremes: Minor - we should use unified name - probably `org.jboss.weld.notification.timeout`

mkouba:
Looks good now!

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/event/options/timeout/NotificationTimeoutTest.java
```diff
@@ -0,0 +1,197 @@
...
+                    synchronizer.add(t);
+                    return null;
+                });
+            Throwable fromSynchronizer = synchronizer.poll(5, TimeUnit.SECONDS);
+            Assert.assertNotNull(fromSynchronizer);
+            Assert.assertTrue(fromSynchronizer.getCause().toString().contains(TimeoutException.class.getSimpleName()));
+            // wait for the observer to finish the jobs
+            countdown.await();
+            // assert that two events happened and timeout came first
+            Assert.assertTrue(SUCCESSION_OF_EVENTS.size() == 2);
```
manovotn: This assertion fails when you execute the whole test class, but passes when you execute just this method.
And I am blind and cannot see why, will take a look at it tomorrow.

mkouba: What's the actual value?

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -352,11 +358,35 @@ public RuntimeException apply(Type eventType) {
...
+            completableFuture = CompletableFuture.anyOf(completableFuture, startTimer(timeout)).thenApply((ignoredObject) -> event);
+        }
         return new AsyncEventDeliveryStage<>(completableFuture, executor);
     }
 
+    /**
+     * Verifies that, if timeout options was set, the executor is available and input value for timeout can be interpreted as Long.
+     * Returns the timeout value if all is alright, null if this option was not requested.
+     */
+    private Long verifyNotificationTimeoutOption(NotificationOptions options) {
```
mkouba: If you pass the timeout option value you can save one `options.get()` invocation. I know it's a supermicro-optimization but I would like to keep the init/validation methods as similar as possible.

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -324,7 +330,7 @@ public RuntimeException apply(Type eventType) {
         }
         // We should always initialize and validate all notification options first
         final NotificationMode mode = initMode(options.get(WeldNotificationOptions.MODE));
-
+        final Long timeout = verifyNotificationTimeoutOption(options);
```
mkouba: We should unify the init/validation method names. If you don't like the `initX` template we could change it:
* `initModeOption()`/`initTimeoutOption()`
* `initializeMode()`/`initializeTimeout()`
* `verifyModeOption()`/`verifyTimeoutOption()`
* ...

mkouba:
Please update the commit message - API upgrade is not required anymore.

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
manovotn: Good point

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
manovotn: Yes, wanted to ask about this.
I didn't like the `init` one because I needed to do not only option check, but also check for the executor presence.
If you don't mind, I will switch both methods to `verify` format.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/event/options/timeout/NotificationTimeoutTest.java
```diff
@@ -0,0 +1,204 @@
...
+            Assert.assertTrue(SUCCESSION_OF_EVENTS.get(0).equals("Timeout"));
+        }
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void testCustomExecutorServiceNotImplementingTimerExecutor() {
+        Weld weld = new Weld() {
+            @Override
+            protected org.jboss.weld.bootstrap.spi.Deployment createDeployment(ResourceLoader resourceLoader, CDI11Bootstrap bootstrap) {
```
manovotn: Note to self - refactor to `Weld.addServices`

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
mkouba: Well, then I think `initializeX()` is the best name because initialization usually covers things like validation and verification before the initialized thing is put into service. Whereas `verifyX()` methods usually test whether something has a correct value.

PR #1639, opened 2017-03-29
WELD-2360 Weld SE - allow to register additional services

mkouba:


environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -644,6 +671,8 @@ public Weld reset() {
         enabledInterceptors.clear();
         enabledDecorators.clear();
         extensions.clear();
+        containerLifecycleObservers.clear();
```
manovotn: Was this an oversight from previous changes? Can't see any relation to this issue.

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -632,7 +637,29 @@ public Weld setProperties(Map<String, Object> propertiesMap) {
...
+     * </p>
+     * <p>
+     * Service implementation may specify their priority using {@link Priority}. Services with higher priority have precedence. Services that do not specify
+     * priority have the default priority of 4500.
+     * </p>
+     *
+     * @param service
+     * @return self
+     */
+    public Weld addService(Service service) {
```
manovotn: How about also a method to register several of them at once?
`addServices(Service... service)`

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
mkouba: Yes.

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
mkouba: Well, adding an SPI service is not supposed to be a frequent operation. So I would like to add a single method only. But you're right that varargs variant is more practical. I can change the method signature.

manovotn: +1 for varargs variant only

PR #1641, opened 2017-03-29
WELD-2356 clean up examples

tremes:


mkouba:
The commit message should contain some brief info about the changes.

manovotn:
**NOTE:** Once we agree on what should be deleted, we must no forget to change Weld docs accordingly. I suppose some of those examples are mentioned there.

mkouba:
http://docs.jboss.org/weld/reference/latest/en-US/html/weldexamples.html

tremes:
I dont' see anything problematic in this part of doc but need to check other parts.

PR #1646, opened 2017-03-30
WELD-2364 review and update "Loose coupling.." doc part.

tremes:
This is first proposal. The description of async events/observers is quite superficial but maybe it's sufficient...

### DIFF
--- a/docs/reference/src/main/asciidoc/decorators.asciidoc
+++ b/docs/reference/src/main/asciidoc/decorators.asciidoc
@@ -180,9 +180,5 @@ ensuring deterministic behavior, and
 Decorators enabled using `@Priority` are called before decorators
 enabled using `beans.xml`, the lower priority values are called first.
 
-NOTE: Having a decorator enabled by `@Priority` and in the same time listed
-in `beans.xml` leads to a non-portable behaviour! This combination of enablement 
-should therefore be avoided in order to maintain consistent behaviour across 
-different CDI implementations.
-As for Weld, such decorator is only invoked once in the `@Priority part of the 
+NOTE: A decorator enabled by `@Priority` and in the same time listed in beans.xml is only invoked once in the `@Priority` part of the
 invocation chain. E.g. the enablement via `beans.xml` will be ignored.
--- a/docs/reference/src/main/asciidoc/events.asciidoc
+++ b/docs/reference/src/main/asciidoc/events.asciidoc
@@ -44,11 +44,18 @@ public @interface Updated {}
 === Event observers
 
 An _observer method_ is a method of a bean with a parameter annotated
-`@Observes`.
+`@Observes` or `@ObservesAsync`.
 
 [source.JAVA, java]
 -------------------------------------------------------------------
 public void onAnyDocumentEvent(@Observes Document document) { ... }
+-------------------------------------------------------------------
+
+or in asynchronous version
+
+[source.JAVA, java]
+-------------------------------------------------------------------
+public void onAnyDocumentEvent(@ObservesAsync Document document) { ... }
 -------------------------------------------------------------------
 
 The annotated parameter is called the _event parameter_. The type of the
@@ -84,7 +91,7 @@ public void afterDocumentUpdate(@Observes @Updated Document document, User user)
 
 === Event producers
 
-Event producers fire events using an instance of the parameterized
+Event producers fire events either synchronously or asynchronously using an instance of the parameterized
 `Event` interface. An instance of this interface is obtained by
 injection:
 
@@ -93,7 +100,8 @@ injection:
 @Inject @Any Event<Document> documentEvent;
 -------------------------------------------
 
-A producer raises events by calling the `fire()` method of the `Event`
+==== Synchronous event producers
+A producer raises synchronous events by calling the `fire()` method of the `Event`
 interface, passing the event object:
 
 [source.JAVA, java]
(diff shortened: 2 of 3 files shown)


docs/reference/src/main/asciidoc/events.asciidoc
```diff
@@ -93,25 +100,47 @@ injection:
...
 assignable, and
 * specifies no qualifiers.
 
-The container simply calls all the observer methods, passing the event
+The container simply calls all the synchronous observer methods, passing the event
 object as the value of the event parameter. If any observer method
 throws an exception, the container stops calling observer methods, and
 the exception is rethrown by the `fire()` method.
 
+==== Asynchronous event producers
```
mkouba: I think it would make sense to move the "Notification options" section under this one.

docs/reference/src/main/asciidoc/events.asciidoc
```diff
@@ -93,25 +100,47 @@ injection:
...
+* has an event parameter to which the event object (the `Document`) is
+assignable, and
+* specifies no qualifiers.
+
+The container simply calls all the asynchronous observer methods, passing the event
+object as the value of the event parameter. If any observer method
+throws an exception, the container doesn't stop calling observer methods and
+the `CompletionException` contains all exceptions thrown by observers as suppressed exceptions.
+
+===== Applying qualifiers to event
```
mkouba: This should be probably only `====` section - it's applicable to both sync and async events.

docs/reference/src/main/asciidoc/events.asciidoc
tremes: Yes good point

docs/reference/src/main/asciidoc/events.asciidoc
```diff
@@ -93,25 +100,66 @@ injection:
...
+A producer raises synchronous events by calling the `fire()` method of the `Event`
 interface, passing the event object:
 
 [source.JAVA, java]
 -----------------------------
 documentEvent.fire(document);
 -----------------------------
 
-This particular event will be delivered to every observer method that:
+This particular event will be delivered only to synchronous observer method that:
```
manovotn: I think the position of `only` in the sentence isn't correct. How about:
`This particular event will only be delivered to synchronous observer method that:`
If you decide to change it, please do the same in the async part of this chapter.

docs/reference/src/main/asciidoc/decorators.asciidoc
```diff
@@ -181,8 +181,5 @@ Decorators enabled using `@Priority` are called before decorators
 enabled using `beans.xml`, the lower priority values are called first.
 
 NOTE: Having a decorator enabled by `@Priority` and in the same time listed
-in `beans.xml` leads to a non-portable behaviour! This combination of enablement 
-should therefore be avoided in order to maintain consistent behaviour across 
-different CDI implementations.
-As for Weld, such decorator is only invoked once in the `@Priority part of the 
+in `beans.xml`, such decorator is only invoked once in the `@Priority part of the
```
manovotn: This sentence sounds awkward to me (`...in the same time listed in beans.xml, such decorator ...`). How about:
`NOTE: A decorator enabled by @Priority and in the same time listed in beans.xml is only invoked once in the @Priority ...`
If you change this, please change the one in interceptors section as well.

docs/reference/src/main/asciidoc/events.asciidoc
```diff
@@ -93,25 +100,66 @@ injection:
...
 * specifies no qualifiers.
 
-The container simply calls all the observer methods, passing the event
+The container simply calls all the synchronous observer methods, passing the event
 object as the value of the event parameter. If any observer method
 throws an exception, the container stops calling observer methods, and
 the exception is rethrown by the `fire()` method.
 
+==== Asynchronous event producers
+A producer raises asynchronous events by calling the `fireAsync()` method of the `Event`
```
manovotn: Just a thought, wouldn't it be better to `fire event` instead of `raise events`? I mean both are correct but perhaps we should unify it. Though I haven't checked how many times we should one or the other in the docs.

docs/reference/src/main/asciidoc/events.asciidoc
```diff
@@ -93,25 +100,66 @@ injection:
...
+
+This particular event will be delivered only to asynchronous observer method that:
+
+* has an event parameter to which the event object (the `Document`) is
+assignable, and
+* specifies no qualifiers.
+
+The container simply calls all the asynchronous observer methods, passing the event
+object as the value of the event parameter. If any observer method
+throws an exception, the container doesn't stop calling observer methods and
```
manovotn: You sentence is probably correct but it is very long and therefore a bit confusing. I suggest this:
`... throws an exception, the container will suppress it and notify remaining observers. The resulting CompletionStage will then finish exceptionally with CompletionException containing all previously suppressed exceptions.`

docs/reference/src/main/asciidoc/events.asciidoc
```diff
@@ -93,25 +100,66 @@ injection:
...
+the `CompletionException` contains all exceptions thrown by observers as suppressed exceptions.
+
+===== Notification options
+
+The `Event.fireAsync()` method may be called with a `NotificationOptions` parameter to configure the notification of asynchronous observer methods , e.g. to specify an `Executor` object to be used for asynchronous delivery.
+Weld defines the following non-portable notification options:
```
manovotn: Perhaps we should mention the newly added API class (`WeldNotificationOption`) which has these values pre-defined?

mkouba: A simple "see also Notification options" link would be fine.

mkouba: @tremes Pls rebase - this is not the current wording of this section!

docs/reference/src/main/asciidoc/events.asciidoc
tremes: I just used the original wording.

docs/reference/src/main/asciidoc/events.asciidoc
tremes: OK

docs/reference/src/main/asciidoc/events.asciidoc
tremes: Could be. Thanks for suggestion.

docs/reference/src/main/asciidoc/decorators.asciidoc
tremes: Yes sounds better.

docs/reference/src/main/asciidoc/events.asciidoc
```diff
@@ -93,25 +100,70 @@ injection:
...
+documentEvent.fireAsync(document);
+-----------------------------
+
+This particular event will only be delivered to asynchronous observer method that:
+
+* has an event parameter to which the event object (the `Document`) is
+assignable, and
+* specifies no qualifiers.
+
+The container simply calls all the asynchronous observer methods, passing the event
```
mkouba: I think this should be more clear here. Ie. the method returns immediately and the observers are notified _"in one or more different threads"_... or something like that?

tremes: OK

mkouba:
Looks good.

docs/reference/src/main/asciidoc/events.asciidoc
antoinesd: "raise" term comes from Seam 2.0 I think. Fire() would be better but perhaps we should treat that in another PR...

manovotn: Yea, I meant we might want to replace `raise` for `fire` throughout whole spec.

tremes: spec or doc? I would personally keep orignal wording in the doc.

PR #1647, opened 2017-03-31
WELD-2365 Docs - review "Getting started with Weld"

mkouba:
- also upgrade Maven WildFly plugin to 1.1.0.Final

### DIFF
--- a/docs/reference/src/main/asciidoc/gettingstarted.asciidoc
+++ b/docs/reference/src/main/asciidoc/gettingstarted.asciidoc
@@ -5,7 +5,7 @@ Weld comes with a number of examples. We recommend you start with
 `examples/jsf/numberguess` and `examples/jsf/translator`. Numberguess is
 a web (war) example containing only non-transactional managed beans.
 This example can be run on a wide range of servers, including WildFly ,
-GlassFish, Apache Tomcat, Jetty, Google App Engine, and any compliant
+GlassFish, Apache Tomcat, Jetty, and any compliant
 Java EE 7 container. Translator is an enterprise (ear) example that
 contains session beans. This example must be run on WildFly 8 or better,
 GlassFish 4 or better, or any compliant Java EE 7 container.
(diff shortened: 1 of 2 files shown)


docs/reference/src/main/asciidoc/gettingstarted.asciidoc
```diff
@@ -33,12 +33,24 @@ You can also deploy the generated artifact (war or ear) to any other
 container that supports Java EE 7, such as GlassFish 4.
 
 The sections below cover the steps for deploying with Maven in detail.
-Let's start with WildFly.
+
+=== First try
+
+If you simply want to run the example without the requirement of a specific runtime then start with the following commands.
```
tremes: It reads weird IMO. `the example` - but which one? It doesn't explicitly talk about numberguess before AFAIK.

tremes:
Otherwise it's ok

docs/reference/src/main/asciidoc/gettingstarted.asciidoc
```diff
@@ -20,8 +20,8 @@ following:
 
 * the latest release of Weld, which contains the examples
 * Maven 3, to build and deploy the examples
-* a supported runtime environment (minimum versions shown)
-** WildFly 8.0.0.Final,
+* optionally a supported runtime environment (minimum versions shown)
```
manovotn: Minor thing, there should be a comma after `optionally` -> `optionally,`

docs/reference/src/main/asciidoc/gettingstarted.asciidoc
```diff
@@ -20,8 +20,8 @@ following:
...
+** WildFly 8.0.0.Final (10.1.0.Final recommended),
```
manovotn: This doc will concern running Weld 3.x on WFLY - I am pretty sure you cannot run the numberguess as easily as doing `mvn wildfly:run`. And furthermore, not on WFLY 8 for sure. Or am I missing something obvious?

docs/reference/src/main/asciidoc/gettingstarted.asciidoc
mkouba: Well, it should work for both examples. But you're right, it will be better to be more specific.

docs/reference/src/main/asciidoc/gettingstarted.asciidoc
mkouba: Ok.

PR #1648, opened 2017-04-03
Move TestClassToHashMapper to src/main/java to not cause CNFE with ma…

tremes:
…ven.test.skip prop.


mkouba:
Please create a PR for 2.4 branch. Thanks.

PR #1649, opened 2017-04-03
WELD-2366 Review Weld 3 docs - I. Beans.

manovotn:
Most are minor changes.
I removed the sections about `@New` qualifier, as with Weld 3.x, people should know better :)

### DIFF
--- a/docs/reference/src/main/asciidoc/injection.asciidoc
+++ b/docs/reference/src/main/asciidoc/injection.asciidoc
@@ -234,8 +234,7 @@ Whenever a bean or injection point does not explicitly declare a
 qualifier, the container assumes the qualifier `@Default`. From time to
 time, you'll need to declare an injection point without specifying a
 qualifier. There's a qualifier for that too. All beans have the
-qualifier`
-                @Any`. Therefore, by explicitly specifying `@Any` at an
+qualifier `@Any`. Therefore, by explicitly specifying `@Any` at an
 injection point, you suppress the default qualifier, without otherwise
 restricting the beans that are eligible for injection.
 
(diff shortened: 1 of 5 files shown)


docs/reference/src/main/asciidoc/injection.asciidoc
```diff
@@ -551,6 +558,12 @@ Annotation qualifier = synchronously ?
 PaymentProcessor p = anyPaymentProcessor.select(qualifier).get().process(payment);
 ----------------------------------------------------------------------------------
 
+[NOTE]
```
tremes: I would move this note to `4.4. The built-in qualifiers @Default and @Any` chapter

docs/reference/src/main/asciidoc/injection.asciidoc
```diff
@@ -551,6 +558,12 @@ Annotation qualifier = synchronously ?
...
+====
+Since CDI 2.0, most annotations from `javax.enterprise` package have their `AnnotationLiteral` implementations.
+Therefore, in order to to programatically obtain (for instance) `@Any` annotation, you can simply do `Any.Literal.INSTANCE`.
```
tremes: Typo - double `to`.

manovotn: Thanks

docs/reference/src/main/asciidoc/injection.asciidoc
manovotn: Hmm, I disagree. 
In `4.4` we don't say a word about implementing a qualifier in order to use it programmatically.
Here, on the contrary, we do, so I think we should straight away say that you needn't implement all of them.

PR #1651, opened 2017-04-04
WELD-2363 Docs - review IV. CDI and the Java EE ecosystem

mkouba:


docs/reference/src/main/asciidoc/extend.asciidoc
```diff
@@ -138,54 +137,18 @@ An extension observer method is not allowed to inject any other object.
 
 === The `BeanManager` object
 
-The nerve center for extending CDI is the `BeanManager` object. The
-`BeanManager` interface lets us obtain beans, interceptors, decorators,
-observers and contexts programmatically.
+The nerve center for extending CDI is the `BeanManager` object.
+The `BeanManager` interface provides operations of use to portable extensions, e.g. lets us obtain beans, interceptors, decorators, observers and contexts programmatically.
```
tremes: `The BeanManager interface provides operations of use to portable extensions` sounds weird to me.

mkouba: Well, I tried to follow the spec wording. But you're right that it sounds odd to my Czech ears ;-) 
Maybe `The BeanManager interface provides operations useful for portable extensions`?

tremes: Sounds better to my Czech ears.:-)

mkouba:
> 1. Did we mentioned that they are there since 2.0? Sth like CDI 2.0 introduced a concept of

Good idea. Let's add a note about configurators.

> 2. Configurators might not always be sufficient solution. Shouldn't we retain some advanced example...

I don't think so. This would mean adding a new more complicated example. By the way, the `ProcessInjectionTarget` example which is using the "wrapper" approach stays untouched (because there is no configurator in PIT ;-)

manovotn:
> `ProcessInjectionTarget` example which is using the "wrapper" approach stays untouched

Ah, you're right, missed that one.

docs/reference/src/main/asciidoc/extend.asciidoc
```diff
@@ -137,22 +138,29 @@ The observer method may inject a `BeanManager`
 
 An extension observer method is not allowed to inject any other object.
 
+==== Configurators
+
+CDI 2.0 introduced the Configurators API - a new way to easily configure some parts of the SPI during container lifecycle event notification.
+E.g. to add a qualifier to a bean an extension can observe `ProcessBeanAttributes`, then obtain a configurator instance through `ProcessBeanAttributes.configureBeanAttributes()` and finally use `BeanAttributesConfigurator.addQualifier(Annotation)`.
+No need to wrap/delegate to the original `BeanAttributes`.
+See also chapter http://docs.jboss.org/cdi/spec/2.0-PRD/cdi-spec.html#configurators[Configurators interfaces] of the CDI specification.
```
manovotn: Shouldn't this link lead to PFD instead of PRD? Not that it would matter much, since we will need to replace it with final ones eventually.

mkouba: Exactly. PRD or PFD does not matter now. WELD-2370

PR #1653, opened 2017-04-04
WELD-2371 Weld SE - allow progr. lookup during AfterDeploymentValidation

mkouba:


environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
```diff
@@ -238,6 +242,13 @@ private void fireContainerInitializedEvent() {
      */
     public Event<Object> event() {
         checkState();
+        if (event == null) {
+            synchronized (this) {
```
manovotn: I am not following here - why do we want this in `sync` block?
Also, why don't we do it just once in sync block in the `complete` method?

environments/se/core/src/main/java/org/jboss/weld/environment/se/logging/WeldSELogger.java
```diff
@@ -73,7 +73,7 @@
 
     // 2013 used in Weld 2.3
 
-    @Message(id = 2014, value = "Weld SE container {0} not initialized completely", format = Format.MESSAGE_FORMAT)
-    IllegalStateException weldContainerNotInitializedCompletely(Object id);
+    @Message(id = 2014, value = "Weld SE container {0} deployment not validated yet - methods for programmatic lookup cannot be used", format = Format.MESSAGE_FORMAT)
```
manovotn: Sounds a bit clumsy, how about a twist:
`Weld SE container with id {0} has not yet validated the deployment - methods ....`

environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
mkouba: Visibility - `event` is volatile and may be initialized outside the `complete()` method, the same applies to `instance` (that is the point of WELD-2371).

environments/se/core/src/main/java/org/jboss/weld/environment/se/logging/WeldSELogger.java
mkouba: Sounds good.

environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
tremes: So this is basically check whether the Event built-in bean has been already initialized/used right? Still not quite sure why it's volatile. Usage within extension could mean execution in different thread?

mkouba: We want to be sure the `event` is only initialized once and visible to all threads, that's the reason why it's volatile and initialized in a synchronized block.

PR #1654, opened 2017-04-04
Review Weld 3 docs - Appendix.

manovotn:


docs/reference/src/main/asciidoc/ri-spi.asciidoc
```diff
@@ -637,9 +631,6 @@ Servlet deployment safely.
 
 NOTE: Weld only supports JSF 1.2 and above.
```
tremes: This just got my attention. Is that still true? I mean support for JSF 1.2. What about to update or remove this statement? Otherwise looks good.

mkouba: Hm, I think JSF 1.2 is not relevant anymore. Weld 3 depends on JSF 2.2 API. I'm not sure how to update this statement. Given that our Tomcat example is using JSF 2.2 impl we should probably state that Weld 3 supports JSF 2.2+?

mkouba: > OT: Do we have an issue for alignment to EE8 APIs?

Not yet. In fact, Weld 3.0.0.Final does not target Java EE 8.

PR #1657, opened 2017-04-07
WELD-2374 Add option to disable Jandex in SE and Servlet environment.

manovotn:


mkouba:
I think we could share `org.jboss.weld.discovery.disableJandexDiscovery` and the log message for both Weld SE and Servlet (i.e. into `weld-environment-common`).

Also I would add a boolean `isJandexDisabled` (name TBD) param to `org.jboss.weld.environment.deployment.discovery.DiscoveryStrategyFactory.create(ResourceLoader, Bootstrap, Set<Class<? extends Annotation>>)` and let the factory handle the param correctly.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/discovery/disableJandex/JandexDiscoveryStrategyDisabledTest.java
```diff
@@ -0,0 +1,58 @@
...
+        return ClassPath.builder().add(ShrinkWrap.create(BeanArchive.class).addPackage(JandexDiscoveryStrategyDisabledTest.class.getPackage()))
+            .add(oldJandex) //add prehistoric Jandex to CP
+            .addSystemProperty(Jandex.DISABLE_JANDEX_DISCOVERY_STRATEGY, "true") //disable jandex discovery atrategy
+            .build();
+    }
+
+    @Test
+    public void testDeploymentWorksBecauseJandexIsNotUsed() {
+        try (WeldContainer container = new Weld().initialize()) {
+            // no-op, if the option does not work, this will crash
```
tremes: What about to add some basic assertion like container is running or something like that.

manovotn: Can do that, although it makes no real difference. If the option does not work, you get an exception during bootstrap.

manovotn:
BTW apart from the SE test I did test it manually with numberguess on Tomcat 8.5.5.

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/DiscoveryStrategyFactory.java
```diff
@@ -37,17 +37,22 @@ private DiscoveryStrategyFactory() {
...
     public static DiscoveryStrategy create(ResourceLoader resourceLoader, Bootstrap bootstrap,
-            Set<Class<? extends Annotation>> initialBeanDefiningAnnotations) {
-        if (Jandex.isJandexAvailable(resourceLoader)) {
-            CommonLogger.LOG.usingJandex();
-            try {
-                return Jandex.createJandexDiscoveryStrategy(resourceLoader, bootstrap, initialBeanDefiningAnnotations);
-            } catch (Exception e) {
-                throw CommonLogger.LOG.unableToInstantiate(Jandex.JANDEX_DISCOVERY_STRATEGY_CLASS_NAME,
+        Set<Class<? extends Annotation>> initialBeanDefiningAnnotations, boolean jandexStrategyDisabled) {
+        if (jandexStrategyDisabled) {
```
mkouba: I think we could move this check inside the `if (Jandex.isJandexAvailable(resourceLoader))` block. It makes more sense to only show the log message if jandex is on the classpath. WDYT?

manovotn: Hmm, we could but I don't like it.
I specifically made it this way, so that we don't even need to look whether there is any Jandex on CP if this option is set.

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/DiscoveryStrategyFactory.java
```diff
@@ -37,17 +37,22 @@ private DiscoveryStrategyFactory() {
...
-            Set<Class<? extends Annotation>> initialBeanDefiningAnnotations) {
+        Set<Class<? extends Annotation>> initialBeanDefiningAnnotations, boolean jandexStrategyDisabled) {
         if (Jandex.isJandexAvailable(resourceLoader)) {
-            CommonLogger.LOG.usingJandex();
-            try {
-                return Jandex.createJandexDiscoveryStrategy(resourceLoader, bootstrap, initialBeanDefiningAnnotations);
-            } catch (Exception e) {
-                throw CommonLogger.LOG.unableToInstantiate(Jandex.JANDEX_DISCOVERY_STRATEGY_CLASS_NAME,
+            if (jandexStrategyDisabled) {
+                CommonLogger.LOG.jandexDiscoveryStrategyDisabled();
```
manovotn: @mkouba changed based on your suggestion

mkouba:
I think we're ready to merge now. @manovotn could you send a PR for 2.4 branch?

manovotn:
Done in https://github.com/weld/core/pull/1659

PR #1658, opened 2017-04-10
WELD-2372 fire @Initialied/@Destroyed events for postConstruct callback

tremes:
and async observer notification

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/ManagedBean.java
+++ b/impl/src/main/java/org/jboss/weld/bean/ManagedBean.java
@@ -32,9 +32,9 @@ import javax.enterprise.inject.spi.InjectionTarget;
 import org.jboss.weld.annotated.enhanced.EnhancedAnnotatedField;
 import org.jboss.weld.annotated.enhanced.EnhancedAnnotatedType;
 import org.jboss.weld.bootstrap.BeanDeployerEnvironment;
-import org.jboss.weld.contexts.CreationalContextImpl;
 import org.jboss.weld.context.RequestContext;
 import org.jboss.weld.context.unbound.UnboundLiteral;
+import org.jboss.weld.contexts.CreationalContextImpl;
 import org.jboss.weld.injection.producer.BasicInjectionTarget;
 import org.jboss.weld.interceptor.spi.metadata.InterceptorClassMetadata;
 import org.jboss.weld.interceptor.spi.model.InterceptionModel;
@@ -168,10 +168,13 @@ public class ManagedBean<T> extends AbstractClassBean<T> {
             RequestContext context = getUnboundRequestContext();
             try {
                 context.activate();
+                beanManager.fireRequestContextInitialized(getId());
                 getProducer().postConstruct(instance);
             } finally {
+                beanManager.fireRequestContextBeforeDestroyed(getId());
                 context.invalidate();
                 context.deactivate();
+                beanManager.fireRequestContextDestroyed(getId());
             }
         }
         return instance;
@@ -286,7 +289,7 @@ public class ManagedBean<T> extends AbstractClassBean<T> {
             BasicInjectionTarget<?> weldProducer = (BasicInjectionTarget<?>) producer;
             final InterceptionModel interceptors = getInterceptors();
             if (interceptors == null || interceptors.getInterceptors(InterceptionType.POST_CONSTRUCT, null).isEmpty()) {
-                if (!weldProducer.getLifecycleCallbackInvoker().hasPostConstructMethods()) {
+                if (!weldProducer.getLifecycleCallbackInvoker().hasPostConstructCallback()) {
                     return false;
                 }
             }
--- a/impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
+++ b/impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
@@ -370,7 +370,7 @@ public class ObserverNotifier {
      * Returns the timeout value if all is alright, null if this option was not requested.
      */
     private Long initTimeoutOption(Object timeoutOptionValue) {
-        if (timeoutOptionValue  != null) {
+        if (timeoutOptionValue != null) {
             // throw exception if we don't have scheduled executor available
             if (timerExecutor == null) {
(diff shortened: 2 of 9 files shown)


impl/src/main/java/org/jboss/weld/injection/producer/LifecycleCallbackInvoker.java
```diff
@@ -35,5 +35,12 @@
 
     boolean hasPreDestroyMethods();
 
-    boolean hasPostConstructMethods();
+    default boolean hasPostConstructMethods(){
```
mkouba: This method shouldn't be default...

impl/src/main/java/org/jboss/weld/injection/producer/LifecycleCallbackInvoker.java
```diff
@@ -35,5 +35,12 @@
...
+        return hasPostConstructCallback();
+    }
+
+    default boolean hasPostConstructCallback(){
+        return false;
```
mkouba: Should be `return hasPostConstructMethods();`.

impl/src/main/java/org/jboss/weld/injection/producer/NoopLifecycleCallbackInvoker.java
```diff
@@ -53,8 +53,4 @@ public boolean hasPreDestroyMethods() {
         return true; // we cannot say no for sure
     }
 
-    @Override
-    public boolean hasPostConstructMethods() {
```
mkouba: Don't remove this method but add `hasPostConstructCallback()` with `return false`.

impl/src/main/java/org/jboss/weld/bean/ManagedBean.java
```diff
@@ -168,10 +172,13 @@ public T create(CreationalContext<T> creationalContext) {
             RequestContext context = getUnboundRequestContext();
             try {
                 context.activate();
+                FastEvent.of(Object.class, beanManager, Initialized.Literal.REQUEST).fire(this.getId());
```
mkouba: We need to optimize here (cache the event instance or something like this) - this would probably degrade the performance significantly.

mkouba: `FastEvent` instances could be probably cached on `BeanManagerImpl`.

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -432,14 +443,20 @@ private NotificationMode initModeOption(Object value) {
         return () -> {
             final ThreadLocalStackReference<EventMetadata> stack = currentEventMetadata.pushIfNotNull(metadata);
             final RequestContext requestContext = requestContextHolder.get();
+            final FastEvent<Object> requestInitializedEvent = requestInitializedEventHolder.get();
+            final FastEvent<Object> requestBeforeDestroyedEvent = requestBeforeDestroyedEventHolder.get();
+            final FastEvent<Object> requestDestroyedEvent = requestDestroyedEventHolder.get();
             try {
                 securityContext.associate();
+                requestInitializedEvent.fire(toString());
```
mkouba: I think we should use a dedicated `ContextEvent` instance as a payload, similar to  `org.jboss.weld.event.ContextEvent.REQUEST_INITIALIZED_EJB`.

manovotn:
I find the following `if` condition suspicious -> https://github.com/weld/core/pull/1658/commits/8d5551fd9ef5b1ca88b808a842211ff7b8612461#diff-59db52822df42c8332bb69b0821facebL161

If I am not mistaken, no matter what branch (if-else), the `postConstruct` will be called all the same; the only difference is whether we want to activate req context beforehand. It seems that `!hasPostConstructCallback` check is redundant, but maybe I am missing something?

impl/src/main/java/org/jboss/weld/injection/producer/NoopLifecycleCallbackInvoker.java
```diff
@@ -55,6 +55,12 @@ public boolean hasPreDestroyMethods() {
 
     @Override
     public boolean hasPostConstructMethods() {
-        return true; // we cannot say no for sure
+        return hasPostConstructCallback();
```
manovotn: I still don't comprehend this approach - the class is used for interceptors, I get that much. We don't want/need post contruct callbacks there.
What I don't get, is why we still invoke them? Putting `System.out` in `postConstruct` will trigger it. That is IMO not intended

mkouba:
@manovotn I think it's ok. The request context should only be activated if a `@PostConstruct` callback exists. `hasPostConstructCallback` should be true if the bean itself defines a callback or any lifecycle callback interceptor is associated. The truth is that if there is no callback we can probably skip `getProducer().postConstruct()` completely but this would require further verification.

impl/src/main/java/org/jboss/weld/injection/producer/LifecycleCallbackInvoker.java
```diff
@@ -36,4 +36,9 @@
     boolean hasPreDestroyMethods();
 
     boolean hasPostConstructMethods();
+
+    default boolean hasPostConstructCallback(){
+        return hasPostConstructMethods();
+    };
```
mkouba: ...useless semicolon + formatting

impl/src/main/java/org/jboss/weld/bean/ManagedBean.java
```diff
@@ -168,10 +168,13 @@ public T create(CreationalContext<T> creationalContext) {
             RequestContext context = getUnboundRequestContext();
             try {
                 context.activate();
+                beanManager.fireRequestContextInitialized(this.getId());
```
mkouba: `this.getId()` -> `getId()`

impl/src/main/java/org/jboss/weld/injection/producer/NoopLifecycleCallbackInvoker.java
mkouba: This method should return `true` as before - we still cannot say for sure... Otherwise the new `hasPostConstructCallback()` would be useless and we could just override this one.

PR #1660, opened 2017-04-11
WELD-2373 BeanManager#createInstance can only be invoked after deploy…

manovotn:
…ment validation.

@mkouba I did not create a test here, because that should be covered in [TCK-571](https://issues.jboss.org/browse/CDITCK-571). No point in having that duplicated IMO.


mkouba:
> I did not create a test here...

+1

PR #1661, opened 2017-04-11
WELD-2376 Export HTML report if validation fails

mkouba:


probe/core/src/main/java/org/jboss/weld/probe/Reports.java
```diff
@@ -0,0 +1,369 @@
...
+        head.add(HtmlTag.style().add(IOUtils.getResourceAsString("/report.css")));
+
+        HtmlTag body = html.addAndGet(HtmlTag.body());
+
+        body.add(HtmlTag.h1(TITLE));
+        HtmlTag meta = body.addAndGet(HtmlTag.stripedTable());
+        meta.add(HtmlTag.tr().add(HtmlTag.td().add(HtmlTag.strong("Generated at:")), HtmlTag.td(LocalDateTime.now().toString())));
+        meta.add(HtmlTag.tr().add(HtmlTag.td().add(HtmlTag.strong("Weld Version:")), HtmlTag.td(Formats.getSimpleVersion())));
+        meta.add(HtmlTag.tr().add(HtmlTag.td().add(HtmlTag.strong("Weld Environment:")), HtmlTag.td(environment.toString())));
+        meta.add(HtmlTag.tr().add(HtmlTag.td().add(HtmlTag.strong("Java Version:")),
```
tremes: Just minor suggestion - displaying `java.vendor` could probably be helpful as well.

mkouba: Yep, maybe there are more interesting system properties (e.g. `os.name`) we could add.

probe/core/src/main/java/org/jboss/weld/probe/Reports.java
```diff
@@ -0,0 +1,373 @@
...
+            Field field = ConfigurationKey.class.getDeclaredField(key.toString());
+            if (field != null && field.isEnumConstant()) {
+                Description description = field.getAnnotation(Description.class);
+                if (description == null) {
+                    // Don't show config options without description
+                    return null;
+                }
+                return description.value();
+            }
+        } catch (Exception e) {
```
manovotn: FindBugs finds this line not to it's liking it seems:
`Exception is caught when Exception is not thrown in org.jboss.weld.probe.Reports.getDesc(ConfigurationKey) ["org.jboss.weld.probe.Reports"] At Reports.java:[lines 78-370]`

mkouba:
> I see the report is by default generated into...

That's  a good question. Right now, if `org.jboss.weld.probe.exportDataAfterDeployment` is not defined, a system property `user.dir` is used - the current working directory. We can also use the temp dir, ie. `java.io.tmpdir` or user home. I'm not sure what is better... 

> BTW I think it would later deserve a section...

+1

manovotn:
Probably leave it, as it is. Who knows where `user.dir` and `java.io.tmpdir` are on Windows, or worse still on MacFruit systems.

PR #1667, opened 2017-04-19
WELD-2377 add jacoco profile and its property also for SE core tests.

tremes:


tremes:
The current failure looks like intermitten failure in integration tests.

mkouba:
Hm, there should be no tests which fail intermittenly ;-)

mkouba:
retest this please

PR #1668, opened 2017-04-25
WELD-2380 ObserverNotifier should by default use fucntional SecurityS…

manovotn:
…ervice API method.


impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -427,22 +427,24 @@ private NotificationMode initModeOption(Object value) {
...
-                stack.pop();
-                requestContext.invalidate();
-                requestContext.deactivate();
-                securityContext.dissociate();
-                securityContext.close();
-            }
+            securityContextActionConsumer.accept(() -> {
+                try {
+                    requestContext.activate();
+                    System.out.println(Thread.currentThread().getId());
```
mkouba: ???

manovotn: Uh my bad, didn't realize I created a PR already. Just wanted to store it somewhere :-D
Will remove of course.

PR #1669, opened 2017-04-28
WELD-2273 Decorated bean with private constructors only cause NPE in SE  

antoinesd:
When Weld relaxed mode is true, some checks are skipped, which could trigger an NPE if a decorated bean has no non private constructor.

### DIFF
--- a/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/DecoratedWithPrivateConstructorTest.java
+++ b/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/DecoratedWithPrivateConstructorTest.java
@@ -42,15 +42,12 @@ public class DecoratedWithPrivateConstructorTest {
                 .build();
     }
 
-    @Test
+    @Test(expected = DeploymentException.class)
     public void privateConstructorTest() {
-        try {
-            WeldContainer container = new Weld().initialize();
-            Assert.fail();
-        } catch (DeploymentException e) {
-            Assert.assertTrue(e.getMessage().contains("Bean class which has decorators should declare at least one non private constructor"));
-        }
-
+        WeldContainer container = new Weld()
+                .property("org.jboss.weld.construction.relaxed", true)
+                .initialize();
+        Assert.fail();
     }
 }
 
--- a/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/DecoratedWithPrivateConstructorTest.java
+++ b/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/DecoratedWithPrivateConstructorTest.java
@@ -1,0 +1,55 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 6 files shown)


environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/DecoratedWithPrivateConstructorTest.java
```diff
@@ -0,0 +1,55 @@
...
+    public static Archive<?> createTestArchive() {
+        return ClassPath.builder().add(ShrinkWrap.create(BeanArchive.class)
+                                               .addPackage(Bravo.class.getPackage()))
+                .build();
+    }
+
+    @Test
+    public void privateConstructorTest() {
+        try {
+            WeldContainer container = new Weld().initialize();
```
tremes: I would add `Assert.fail()`

antoinesd: +!

tremes:
This proposal doesn't seem to me 100% correct. Just try to add `public BravoImpl(Foo foo){}` to `BravoImpl` and you will get NPE.

antoinesd:
You're right @tremes. I thought I had tested this use case but obviously not. Thx.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/DecoratedWithPrivateConstructorTest.java
```diff
@@ -0,0 +1,56 @@
...
```
mkouba: You should also set `org.jboss.weld.construction.relaxed` to true. It's currently enabled by default in SE but this might change in the future.

impl/src/main/java/org/jboss/weld/injection/producer/BeanInjectionTarget.java
```diff
@@ -138,10 +138,24 @@ protected void setupConstructorInterceptionInstantiator(InterceptionModel interc
...
+    protected void checkConstructors(EnhancedAnnotatedType<T> type) {
         if (!beanManager.getServices().get(ProxyInstantiator.class).isUsingConstructor()) {
+            checkNonPrivateConstructor(type);
+        } else {
+            checkNoArgsConstructor(type);
+        }
+    }
+
+    protected void checkNonPrivateConstructor(EnhancedAnnotatedType<T> type) {
+        if (type.getEnhancedConstructors().stream().anyMatch(c -> !c.isPrivate())) {
```
mkouba: We should avoid using Java8 for issues which could be backported in 2.4.

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
```diff
@@ -519,4 +519,7 @@
     @Message(id = 1569, value = "Cannot inject injection point metadata in a non @Dependent bean: {0}", format = Format.MESSAGE_FORMAT)
     IllegalArgumentException cannotInjectInjectionPointMetadataIntoNonDependent(Object bean);
 
+    @Message(id = 1570, value = "Bean class which has decorators should declare at least one non private constructor: {0}\n  StackTrace:", format = Format.MESSAGE_FORMAT)
```
mkouba: Well, the message should rather clarify that "relaxed construction" feature does not work for decorators. Maybe we should investigate first whether it wouldn't be possible to fix it, ie. make relaxed construction work with decorators?

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/DecoratedWithPrivateConstructorTest.java
```diff
@@ -0,0 +1,56 @@
...
+            Assert.fail();
+        } catch (DeploymentException e) {
+            Assert.assertTrue(e.getMessage().contains("Bean class which has decorators should declare at least one non private constructor"));
```
manovotn: Assertion on an exact `String` in exception is a bad idea.
You can either simply `catch (DeploymentException e)` and do nothing (because you `fail()` it otherwise and the point here is not to get NPE) or you could use the `@Test(expected=DeploymentException.class)`.

antoinesd:
made suggested changes.
@mkouba. In relaxed mode Decorated bean must have a non private constructor with no args or annotated with `@Inject`. Isn't it the expected behaviour ?

mkouba:
Closed in favor of https://github.com/weld/core/pull/1687.

PR #1670, opened 2017-05-02
WELD-2382 Bootstrap skip logging increased to WARN. Refactor Conversa…

manovotn:
…tionAwareViewHandler to allow for null value if the container is not available.


manovotn:
I did verify it with the reproducer mentioned in WELD-2382.
I haven't created an automated test as I think it's not worth it. Feel free to prove me wrong.

mkouba:
I agree that the test is not worth it. It's a very rare use case.

PR #1676, opened 2017-05-17
Add syntax for individual tests.

benjamin-confino:


mkouba:
@benjamin-confino Thanks for the contribution. However, I don't think it is necessary to add this info - it's a standard Maven Surefire Plugin feature - see also http://maven.apache.org/surefire/maven-surefire-plugin/examples/single-test.html. Also `-Dtest` works for TCK as well ([there is a workaround for SUREFIRE-569](https://github.com/weld/core/blob/master/jboss-tck-runner/pom.xml#L174)). Finally, the syntax to run  subset of the tests methods is `-Dtest=TestClass#testMethod` (for both `-Dtest` and `-DtckTest`).

benjamin-confino:
Even for standard features used across many projects I like to keep in mind the possibility that this will be the reader's first time using the feature. But if you insist I won't mind if you close the pull request.

As for the syntax, I'm just going by what worked when I ran it locally. I needed to use -DtckTest and if I run with `-DtckTest=TestClass.testMethod` it works as I would expect, but `-DtckTest=TestClass#testMethod` seems to ignore "#testMethod". If that's not expected behavior then I don't know what's causing it.

mkouba:
Ok, let's add some info to the readme. But we should be sure about the syntax and possibilitites ;-) @tremes @manovotn Could you verify my findings?

Hm, I've just tried:
* `mvn clean test -Dtest=FireEventTest#testBeanManagerFireEvent`, works, surefire pattern could be used
* `mvn clean test -DtckTest=FireEventTest#testBeanManagerFireEvent` - works, single method executed
* `mvn clean test -DtckTest=FireEventTest#testBeanManagerFireEvent+testTypeVariableEventTypeFails` - works, comma-separated list of methods executed
* `mvn clean test -DtckTest=FireEventTest.testBeanManagerFireEvent` - does not run any test method, dots should be used to qualifty the test class IMHO
```

benjamin-confino:
I'd suggest going with what worked on your machine and ignoring mine, I'm still ironing out this workspace.

tremes:
@mkouba I'll check that but I think your findings will likely by correct.

tremes:
Confirmed, updated and merged. Thank you.

PR #1677, opened 2017-05-18
WELD-2387 Configurator defining a custom bean with no callback will f…

manovotn:
…ail with ISE during deployment.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/BeanConfiguratorImpl.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/BeanConfiguratorImpl.java
@@ -311,6 +311,12 @@ public class BeanConfiguratorImpl<T> implements BeanConfigurator<T>, Configurato
     }
 
     public Bean<T> complete() {
+        if (createCallback == null) {
+            // not callback specified, Weld does not know how to instantiate this new custom bean
+            throw BeanLogger.LOG.noCallbackSpecifiedForCustomBean("bean [" + beanClass.toString()
+                + ", with types: " + Formats.formatTypes(attributes.types)
+                + ", and qualifiers: " + Formats.formatAnnotations(attributes.qualifiers) + "]");
+        }
         return new ImmutableBean<>(this);
     }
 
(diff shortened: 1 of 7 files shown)


impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/BeanConfiguratorImpl.java
```diff
@@ -311,7 +311,12 @@ public BeanConfiguratorImpl(Class<?> defaultBeanClass, BeanDeploymentFinder bean
     }
 
     public Bean<T> complete() {
-        return new ImmutableBean<>(this);
+        ImmutableBean<T> result = new ImmutableBean<>(this);
+        if (createCallback == null) {
```
manovotn: I am aware that creating bean and only then throwing exception isn't the best way, but without having `ImmutableBean`, there is no way to access information like types of bean or qualifiers.

manovotn:
The failure is TCK-aspected - there is a [test](https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/extensions/configurators/bean/LifecycleObserver.java#L165) creating a bean with no callback (but the testcase is to verify defaults scope, so we can fix that - already on it).

impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/BeanConfiguratorImpl.java
mkouba: Well, you could make `BeanAttributesConfiguratorImpl.qualifiers` and `types` package-private and use something like:

````
throw BeanLogger.LOG.noCallbackSpecifiedForCustomBean("Configurator [types: " + Formats.formatTypes(attributes.types) + ", qualifiers: " + Formats.formatAnnotations(attributes.qualifiers) + "]");
````

Hm, `BeanAttributesConfiguratorImpl.types` should be final - pls add this change too.

mkouba:
Please also update the exclude list: `jboss-tck-runner/src/test/tck20/tck-tests.xml`.

PR #1680, opened 2017-05-29
WELD-2390 update StereoTypeModel check.

tremes:


manovotn:
Looks like a compilation error. You are using `assertNotEquals` which is in JUnit since 4.11 ( I think) and 2.4 uses JUnit 4.8 (master uses newer version).
I think we might as well upgrade JUnit too?

mkouba:
+1 for upgrade

tremes:
Updated. @manovotn thanks for checking.

PR #1681, opened 2017-06-02
WELD-2386 allow to define bean discovery mode for synthetic archives …

manovotn:
…in SE environment.

### DIFF
--- a/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
+++ b/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
@@ -82,6 +82,7 @@ import org.jboss.weld.environment.se.logging.WeldSELogger;
 import org.jboss.weld.environment.util.BeanArchives;
 import org.jboss.weld.environment.util.DevelopmentMode;
 import org.jboss.weld.environment.util.Files;
+import org.jboss.weld.environment.util.Reflections;
 import org.jboss.weld.metadata.BeansXmlImpl;
 import org.jboss.weld.resources.ClassLoaderResourceLoader;
 import org.jboss.weld.resources.spi.ClassFileServices;
@@ -214,6 +215,8 @@ public class Weld implements ContainerInstanceFactory {
     private String containerId;
 
     private boolean discoveryEnabled = true;
+
+    private BeanDiscoveryMode beanDiscoveryMode = BeanDiscoveryMode.ALL;
 
     private final Set<String> beanClasses;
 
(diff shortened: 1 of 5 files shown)


environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -662,6 +664,17 @@ public Weld addServices(Service... services) {
     }
 
     /**
+     * Sets the bean discovery mode for synthetic bean archive. Default mode is ALL.
+     * @param mode Bean discovery mode as string, possible values are ALL, ANNOTATED, NONE
+     *
+     * @return  self
+     */
+    public Weld setSyntheticArchiveBeanDiscoveryMode(String mode) {
```
mkouba: I think `setBeanDiscoveryMode()` should be ok. We don't use `syntheticArchive` in other methods neither.

manovotn: Hmm, can do I just wanted to make it clear that this makes sense only once you `disableDiscovery()`

mkouba: Also `BeanDiscoveryMode` is part of SPI. So we could probably use the enum as the param.

tremes: +1 to enum

tremes:
It's just minor, but I would add new SE tests to `environments/se/tests`

manovotn:
@tremes I don' understand, that's where the tests are, isn't it?

tremes:
Ah sorry you're right. I thought it was in `environments/se/core/test/`.

manovotn:
Since weekend are long...a note to myself:
* Martin suggested a filter instead of mode as parameter - we could expand on this.
* I removed all notions of mode `NONE` as an option because it makes no sense in SE.
** Therefore I added an exception/test for IAE when you use this mode

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -662,6 +665,21 @@ public Weld addServices(Service... services) {
     }
 
     /**
+     * Sets the bean discovery mode for synthetic bean archive. Default mode is ALL.
+     * @param mode Bean discovery mode as string, possible values are ALL, ANNOTATED
```
mkouba: It's not a string anymore...

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -662,6 +665,21 @@ public Weld addServices(Service... services) {
...
+     *
+     * @return  self
+     */
```
mkouba: Add `@throws IllegalArgumentException If...`

mkouba:
The idea of a filter is probably an overkill.

PR #1682, opened 2017-06-05
WELD-2392 Log a warning when we detect bean module without EEModuleDescriptor but with servlet listener.

manovotn:


manovotn:
@mkouba I don't really know how to test this. Moreover, I am not convinced this is a bug at all as it only appears alongside incorrect SPI usage.

mkouba:
I think it is a bug since `EEModuleDescriptor` is optional.

impl/src/main/java/org/jboss/weld/bootstrap/BeanDeploymentModule.java
```diff
@@ -55,8 +55,13 @@
     private final boolean web;
     private final ObserverNotifier notifier;
     private final Set<BeanManagerImpl> managers;
+    private final boolean fallbackModule;
```
mkouba: I don't think we need a special flag to mark a fallback - the default/fallback BDM has a specific `id`.

impl/src/main/java/org/jboss/weld/bootstrap/WeldStartup.java
```diff
@@ -545,7 +545,7 @@ public void endInitialization() {
             // fire @Initialized(ApplicationScoped.class) for non-web modules
             // web modules are handled by HttpContextLifecycle
             for (BeanDeploymentModule module : modules) {
-                if (!module.isWebModule()) {
+                if (!module.isWebModule() && !module.isFallbackModule()) {
```
mkouba: So we wouldn't fire an event for any non-web module if `EEModuleDescriptor` is not provided? That does not look correct.

PR #1684, opened 2017-06-05
WELD-2396 Remove redundant BeanManagerImpl.childId.

manovotn:


mkouba:
The commit message is missing JIRA key.

manovotn:
Oops, overlooked that. Both the PR and the commit msg should now have the JIRA key.

PR #1687, opened 2017-06-07
WELD-2273 Decorating bean with private constructor is not allowed.

manovotn:
So, after quite some time digging in this, I found that we in fact even verify this [in code](https://github.com/weld/core/blob/3.0.0.Final/impl/src/main/java/org/jboss/weld/injection/producer/BeanInjectionTarget.java#L148) already, but the [check for relaxed mode](https://github.com/weld/core/blob/3.0.0.Final/impl/src/main/java/org/jboss/weld/injection/producer/BeanInjectionTarget.java#L142) incorrectly skips this verification. Incorrectly, because the mode has nothing to do with this - this is a Jav lang. limitation; we cannot instantiate decorated bean without calling constructor (e.g. with Unsafe) because of potential data-loss.

Feedback and comments are welcome.

### DIFF
--- a/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/invalid/DecoratorWithPrivateConstructorTest.java
+++ b/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/invalid/DecoratorWithPrivateConstructorTest.java
@@ -1,0 +1,67 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2017, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/decorator/invalid/DecoratorWithPrivateConstructorTest.java
```diff
@@ -0,0 +1,67 @@
...
+            .initialize()) {
+            container.select(ImplementingBean.class).get().ping();
+        }
+    }
+
+    @Test(expected = DeploymentException.class)
+    public void testExceptionIsThrownWithRelaxedMode() {
+        Weld weld = new Weld();
+        try (WeldContainer container = weld.disableDiscovery().addPackages(DecoratorWithPrivateConstructorTest.class.getPackage())
+            // Weld SE has by defualt org.jboss.weld.construction.relaxed = true, no need to set it
```
mkouba: It's not a good idea to rely on a default setting. I would recommend using `.property("org.jboss.weld.construction.relaxed", true)`.

mkouba:
Otherwise looks good!

mkouba:
retest this please

mkouba:
retest this please

PR #1688, opened 2017-06-07
WELD-2273 Decorating bean with private constructor is not allowed.

manovotn:


manovotn:
Oops, accidentally made double PR - closing.

PR #1690, opened 2017-06-08
WELD-2393 Fire PIT and PP before PBA.

manovotn:


tests-arquillian/src/test/java/org/jboss/weld/tests/event/lifecycle/ordering/PoorWorker.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Inject
+    MassiveJugCoffee coffee;
+
+    @Produces
+    public HighQualityAndLowCostProduct makeManagersHappy(String docs) {
+        return new HighQualityAndLowCostProduct();
+    }
+
+    @Produces
```
mkouba: It's a good idea to use a qualifier for any `String` bean type. Otherwise, we risk ambiguous dependencies. E.g. we had similar problems with beans produced by Batch properties before `@BatchProperty` was introduced.

manovotn: Not sure I follow...Arquillian specifically adds this package to the deployment. Where would the ambiguous dependency come from?

mkouba: Integration of other EE specs, other WildFly modules, etc.

PR #1693, opened 2017-06-13
WELD-2399 Add WeldContainer.current method.

manovotn:


environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
```diff
@@ -122,6 +122,17 @@ public static WeldContainer instance(String id) {
         }
     }
 
+    public static WeldContainer current() {
```
mkouba: Where's the contract? ;-)

mkouba: I mean javadoc - we should support autocompletion-based development...

manovotn: Yea I know what you mean, I was too hungry and pushed it without docs :-P

environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
```diff
@@ -123,6 +123,23 @@ public static WeldContainer instance(String id) {
     }
 
     /**
+     * A convenient method for CDI.current(). Returns current WeldContainer instance
+     * if there is exactly one instance running. Throws IllegalStateException in any other case.
+     * @return current WeldContainer instance iff exactly one instance exists
```
mkouba: `iff`typo and pls use `{@link }` for classes and methods.

manovotn: `iff` is not a typo, it's an abbreviation for `if and only if`

mkouba: Well, if it's abbreviation then use upper case please. But I would rather avoid using similar abbreviations in javadoc...

environments/se/core/src/main/java/org/jboss/weld/environment/se/WeldContainer.java
```diff
@@ -123,6 +123,23 @@ public static WeldContainer instance(String id) {
     }
 
     /**
+     * A convenient method for {@link CDI.current()}. Returns current {@link WeldContainer} instance
+     * if there is exactly one instance running. Throws {@link IllegalStateException} in any other case.
+     * @return Current {@link WeldContainer} instance if and only if exactly one instance exists
+     * @throws {@link IllegalStateException} if there is 0 or more than 1 instance running
+     */
+    public static WeldContainer current() {
+        List<String> ids = WeldContainer.getRunningContainerIds();
```
mkouba: Just a minor impl note - every time you call `getRunningContainerIds()` you get a new immutable copy of `WeldContainer.RUNNING_CONTAINER_IDS`. So you could probably use the list directly.

manovotn: I know, I used that as I think it's a safe way to handle lists.
But you are right that here we can simply use the original list; I will change the code.

mkouba: You're right it's more safe to use the copy here.

PR #1695, opened 2017-06-21
WELD-2401 Revert the behaviour - this cannot be fixed on WFLY side.

manovotn:
This does not seem fixable on WFLY side (so that Swarm stays untouched) and in the same time we cannot do much unless we want to break Weld API. Therefore we should probably revert the change and leave it like that (and think this through for 3.x).

We can leave this PR hanging here for a while and see if we can come up with anything better.


mkouba:
retest please

mkouba:
retest this please

manovotn:
retest this please

mkouba:
retest this please

manovotn:
Added a test which should cover this case in the future - an EAR with Lib and EJB, both of which have an observer listening to `@Initialized` for `AppScope`.

PR #1697, opened 2017-06-29
WELD-2397 Allow to create WildFly patch through Maven profile.

manovotn:
With this change you can run command such as the following one to generate a patch for WFLY 10.1.0.Final (assuming you have prepared the patched WFLY previously ofc):

`mvn clean install -Pwfly-patch-gen -Dwildfly.download.version=10.1.0.Final -DpatchConfigFile=/home/manovotn/GitRepo/build/wildfly/patch-config-wildfly-10-weld-2.4.xml -DwildflyPatched=/home/manovotn/GitRepo/wildfly/dist/target/wildfly-10.1.0.Final -Denforcer.skip=true`

There are two options, you can either let Maven download the original, pristine WildFly (which is what the above command does) or you can define a path to already existing one using `-DwildflyOriginal=...`

NOTE: The plugin is in `Alpha` version, therefore our Enforcer plugin will start complaining, hence the need to silence it.

Comments/suggestions are welcome.

### DIFF
--- a/jboss-as/pom.xml
+++ b/jboss-as/pom.xml
@@ -28,8 +28,26 @@
 
     <properties>
         <!-- We have to explicitly state the project version. We cannot use ${project.version}
-            as the release plugin won't deal with double evaluation -->
+        as the release plugin won't deal with double evaluation -->
         <weld.update.version>2.4.5-SNAPSHOT</weld.update.version>
+        
+        <!-- Following section is dedicated to WFLY download/update and patch creation -->
+        <!--Path to the patched WFLY server, if you chose to let Mavne download it, this default will be used-->
+        <wildflyPatched>${project.build.directory}/wildfly-patched</wildflyPatched>
+        <!--Path to patch file; only change if you have your own, else Maven downloads it-->
+        <patchConfigFile>${project.build.directory}/patch.xml</patchConfigFile>
+        <!--These patches are automatically downloaded from github.com/weld/build/tree/master/wildfly, specify the file name here to determine which one to use-->
+        <patch.file.name>patch-config-wildfly-10-weld-2.4.xml</patch.file.name>
+        <!--Use this property to set the path to and the name of the resulting patch zip file-->
+        <patchOutputFile>${project.build.directory}/patch.zip</patchOutputFile>
+        <!--Path to original (pristine) WFLY-->
+        <!--Running with -Pdownload-wfly will make Maven download the version specified below-->
+        <wildflyOriginal>${project.build.directory}/wildfly-original</wildflyOriginal>
+        <wildfly.download.version>10.1.0.Final</wildfly.download.version>
+        
+        <!--Plugin versions-->
+        <download.maven.plugin>1.3.0</download.maven.plugin>
+        <patch.gen.plugin>2.0.1.Alpha5</patch.gen.plugin>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


manovotn:
Notes to self:
* We should also download the script from our repo somehow (an alternative to settting the path to local directory)
* The patched WFLY should be created by:
  * Making a copy of original WFLY
  * Applying ant script to it, same one that we use in `update-jboss-as`

manovotn:
@mkouba @tremes current version should be able to:

- Download WFLY from mvn central with `-Pdownload-wfly`
  - You can specify version with `-Dwildfly.download.version` but it defaults to the one we support (10.1.0.Final)
  - This also creates a secondary copy of wfly in `/target/` so that you have one pristine and one to patch. Maven sets JBOSS_HOME to this new copy.
 - With `-Pupdate-jboss-as` you are patching whatever is in `JBOSS_HOME`, therefore typically the WFLY you just downloaded
 - With `-Pwfly-patch-gen` you trigger:
   - Firstly, download of patch file from out `build` repository. You can specify version but it again default to what we normally use.
   - Secondly, patch generation process which uses the two folders with WFLY you downloaded/patched and the patchfile you just grabbed.
   - This profile can be invoked as a standalone one, provided that you supply all arguments!

If you do:

- `mvn clean install -Pupdate-jboss-as,download-wfly,wfly-patch-gen`
  - In your `target` folder you get pristine WFLY, patched WFLY, patch file (xml) and a complete `patch.zip`
- `mvn clean install -Pupdate-jboss-as,download-wfly`
  - You will end up with pristine WFLY and a patched one in your `target` folder
- `mvn clean install -Pupdate-jboss-as`
  - You can patch your server the way we used it before

Since I am no Maven guru, I would appreciate if you could try this out a bit :)
Any feedback is welcome.

mkouba:
Yes, this is exactly what I meant! Some testing would be great. @tremes @antoinesd pls give it also a try!

tremes:
I've tried the first option (including all 3 profiles) and it seems to work fine! Verified downloaded WFLY (default version), patched WFLY and also the patch.zip file.

mkouba:
Hm, you cannot run `mvn install -Pwfly-patch-gen` separately, i.e. after `mvn install -Pupdate-jboss-as,download-wfly` (`wildflyOriginal` points to `${project.build.directory}/wildfly-original` by default). It's probably not a big deal, but it might be useful in rare cases.

Maybe we should also modify the ant script to include artifacts versions so that the generated patch is the same as if it was done manually?

Otherwise looks good.

manovotn:
> Hm, you cannot run mvn install -Pwfly-patch-gen separately, i.e. after mvn install -Pupdate-jboss-as,download-wfly

How would you do that? I think maven can only safely store these thing in `target` folder (which gets deleted). You don't want it hanging in the repository and any other directory might have persmissions problems.

> Maybe we should also modify the ant script to include artifacts versions so that the generated patch is the same as if it was done manually?

Hmm, interesting idea but I think this might get overly complex. We can track the Weld core versions used (have a property in pom for that). But we don't really know Weld API and CDI API versions which belong to it.

mkouba:
> How would you do that?...

It's not deleted unless you use `clean`.

> Hmm, interesting idea but I think this might get overly complex....

AFAIK `maven-dependency-plugin` does not have to strip the versions from the artifact names.

manovotn:
Hmm right, I kind of automatically use clean every time. Should be fixed now.

> AFAIK maven-dependency-plugin...

Yea, I know that. I didn't realize we can leverage versions from parent pom for APIs.

PR #1698, opened 2017-07-10
WELD-2402 use default proxy package for signed classes

tremes:


tremes:
I have no idea (yet) how to test this.

tremes:
Ahh I need to update the condition for sealed jars as well.

mkouba:
Hm, I'm not sure `org.jboss.weld.proxies.protected` is the best name. It reminds visibility modifiers and is completely unrelated at the same time.

tremes:
@mkouba would you prefer `org.jboss.weld.proxies.signed`? 

Or what about `org.jboss.weld.proxies.inaccessible` or `org.jboss.weld.proxies.closed`?

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -232,14 +234,18 @@ static String getProxyName(String contextId, Class<?> proxiedBeanType, Set<? ext
             if (superInterface == null) {
                 throw new IllegalArgumentException("Proxied bean type cannot be java.lang.Object without an interface");
             } else {
-                if (superInterface.getPackage() == null) {
+                if (superInterface.getSigners() != null || superInterface.getPackage().isSealed()) {
```
mkouba: This might throw NPE if the package is null.

mkouba:
Maybe we could use the default package for all cases (no package, signed, sealed) but log a warning and describe possible problems. WDYT?

tremes:
That was my original idea. Warning might be useful but on the other hand user will not be able to do anything about this I guess.

mkouba:
Yep, but at least a user would be aware of the problems.

tremes:
Ehm....forgot the warning

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -232,15 +235,19 @@ static String getProxyName(String contextId, Class<?> proxiedBeanType, Set<? ext
             if (superInterface == null) {
                 throw new IllegalArgumentException("Proxied bean type cannot be java.lang.Object without an interface");
             } else {
-                if (superInterface.getPackage() == null) {
+                if (superInterface.getPackage() == null || superInterface.getSigners() != null || superInterface.getPackage().isSealed()) {
```
mkouba: Hm, I would add a similar method:
```java
private String getDefaultPackageReason(Class<?> clazz) {
  if (clazz.getPackage() == null) {
    return NULL;
  }
  if (clazz.getSigners() != null) {
    return SIGNED;
  }
  if (clazz.getPackage().isSealed() == null) {
    return SEALED;
  }
  return null;
}
```
And then simply:
```java
String reason = getDefaultPackageReason(superInterface);
if (reason != null) {
  BeanLogger.LOG.generatingProxyToDefaultPackage(superInterface, DEFAULT_PROXY_PACKAGE, reason);
}

File: impl/src/main/java/org/jboss/weld/logging/BeanLogger.java [hunk 3]
```diff
@@ -522,4 +522,8 @@
     @Message(id = 1570, value = "Invalid BeanConfigurator setup - no callback was specified for {0}", format = Format.MESSAGE_FORMAT)
     IllegalStateException noCallbackSpecifiedForCustomBean(Object bean);
 
+    @LogMessage(level = Level.WARN)
+    @Message(id = 1571, value = "Proxy for {0} created in {1} because the class package is {2}.", format = Format.MESSAGE_FORMAT)
```
mkouba: Hm, I think it's the class which is signed so maybe this should be `Proxy for {0} created in {1} because {2}` and the reason should contain `the class...` etc.

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
```diff
@@ -522,4 +522,8 @@
     @Message(id = 1570, value = "Invalid BeanConfigurator setup - no callback was specified for {0}", format = Format.MESSAGE_FORMAT)
     IllegalStateException noCallbackSpecifiedForCustomBean(Object bean);
 
+    @LogMessage(level = Level.WARN)
```
manovotn: I was thinking if `INFO` wouldn't be more appropriate log level? Just because existing applications (which had, say, `null` package) will now get warning without changing anything. Also sealing a package/signing a class is not a bad practice and you shouldn't get warning for that.

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -911,4 +918,17 @@ protected void getMethodHandlerField(ClassFile file, CodeAttribute b) {
...
     }
+
+    private static String getDefaultPackageReason(Class<?> clazz) {
+        if (clazz.getPackage() == null) {
+            return NULL;
+        }
+        if (clazz.getSigners() != null) {
+            return SIGNED;
+        }
+        if (clazz.getPackage().isSealed()) {
```
manovotn: I think we need not tackle this for sealed packages. Last time I tried this was working just fine - I will try and add an automatic test for sealed packages (should be doable with Shrinkwrap).

The reason why I think it works is that sealed packages are checked once while checking the classpath (verifies that given package is present in at most one JAR) and you have no restrictions on adding stuff later on.

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
```diff
@@ -522,4 +522,8 @@
...
+    @Message(id = 1571, value = "Proxy for {0} created in {1} because {2}.", format = Format.MESSAGE_FORMAT)
```
manovotn: Just one thought here - user doesn't know/care where we store the proxies so it might not give him enough info.
E.g. they don't know we create proxies in the same package as was the original class.
Perhaps this msg would be more clear?
`Proxy for {0} created in package {1} which is different from the original one because {2}.`

tremes: The type is printed with FQCN so I think it should be obvious from the message.

manovotn: I know, I meant to underline that placing proxy elsewhere is exceptional behavior. But you are probably right that it should be obvious.

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
manovotn: Ok, did the test in a separate [branch](https://github.com/manovotn/core/blob/weld2402_test/tests-arquillian/src/test/java/org/jboss/weld/tests/proxy/sealed/ProxyCreationInSealedPackageTest.java) and it works without any code changes so I believe we should revert them in this PR. I will enrich this test to use some package private things and add a commit onto this PR.

manovotn:
I added a commit with a test using sealed package and package private access. This was working previously, but will now fail with the suggested changes (since we move the new proxies elsewhere).
Hence I believe we should not change the behaviour for sealed packages.

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
tremes: I think it could be INFO as well.

tremes:
OK thanks for the test @manovotn. I will remove the code handling sealed package.

mkouba:
Ok, looks good now. It seems that Weld is using the `java.security.ProtectionDomain` of the proxied bean type, and this domain provides the code source (through `ProtectionDomain.getCodeSource()`) which is most likely used to validate the sealed packages.

mkouba:
@tremes pls create a PR for 2.4 branch as well.

PR #1700, opened 2017-07-12
Test commmit - don't merge

tremes:
Trying to ignore JsfStressTest due to VM crash problem investigation


mkouba:
The root cause confirmed.

PR #1703, opened 2017-07-12
WELD-2405 : Ensure default method are called using invokeInterface

hypnoce:


mkouba:
test this please

impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedSubclassFactory.java
```diff
@@ -301,8 +301,8 @@ protected void invokeMethodHandler(ClassMethod method, MethodInformation methodI
 
         // this is a self invocation optimisation
         // test to see if this is a self invocation, and if so invokespecial the
-        // superclass method directly
-        if (addProceed) {
+        // superclass method directly or fallback to invokeinterface in case of default methods
```
mkouba: The comment should be rather something like: `Do not optimize...`

mkouba:
retest this please

mkouba:
Do not merge until we know the results of https://issues.jboss.org/browse/WELD-2407.

manovotn:
Just a side note - if I am not missing something, the test in this PR is probably redundant as we already have `InterfaceDefaultMethodInterceptedTest` (under `test/java8` folder though, so not run by default which we might want to change) and it does the same thing. Also, it passed even before, so the test here doesn't cover the underlaying change. Not that I would know how to test that yet :)

hypnoce:
The test passes maybe because it runs against class writer 1.2.x. As @mkouba suggested, I've moved the test here but it runs against classwriter 1.1 when in container mode. I can recommit the previous test that was failing.

manovotn:
@hypnoce Could you perhaps place the previous failing test on some other branch and link it here? I am just struggling to understand what exactly does this fix. Or rather with what test under which conditions it fail.

hypnoce:
@manovotn of course, here it is : https://github.com/hypnoce/core/commit/847aeda36f741c58df94c139237958eeded36b2f

manovotn:
Thanks for the test, I am finally getting a grasp of this!
To sum up - this fix addresses the the behaviour for classfilewriter 1.1.2 ( it manifests in unit tests [core/tests], while integration tests [core/tests-arquillian] already use version 1.2).

The fix IMO looks good I am just confused why you attached the [new test](https://github.com/weld/core/pull/1703/files#diff-5d7ec6806c3902eacde588dc23017c6b) instead of the [original one](https://github.com/hypnoce/core/commit/847aeda36f741c58df94c139237958eeded36b2f)?

hypnoce:
@manovotn I removed the test after a conversation with @mkouba advising me to put the test in the test-arquillian module. I can put back the original test and remove (or keep) the arquillian one. It's up to you.

manovotn:
Well, it's really up to Martin rather than me, but he's not in the office this week.
I just wanted to have a test which would verify this spot-on but it seems really difficult (my broken attempt with forced classfilewriter version is [here](https://github.com/manovotn/core/commit/f6ec771bc2495fe4222879f3cc5a07e236e214c3)).
So let's stick with what you did for now, we firstly need to fully understand and address WELD-2407 anyway.

mkouba:
Closing in favor of https://github.com/weld/core/pull/1712

PR #1706, opened 2017-07-18
WELD-2407 fix selfinvocation of non-intercepted default method and

tremes:
expand SelfInvocationTest.


hypnoce:
I had a local fix that looks the same. Good to be close to a working solution ! Thanks

mkouba:
@tremes pls create a PR for 2.4 branch.

PR #1707, opened 2017-07-18
WELD-2404 support interception for InjectionTarget created

tremes:
programmatically for a MDB.

### DIFF
--- a/modules/ejb/src/main/java/org/jboss/weld/module/ejb/EjbSupportImpl.java
+++ b/modules/ejb/src/main/java/org/jboss/weld/module/ejb/EjbSupportImpl.java
@@ -34,8 +34,11 @@ import org.jboss.weld.ejb.spi.EjbDescriptor;
 import org.jboss.weld.ejb.spi.EjbServices;
 import org.jboss.weld.injection.producer.AbstractInstantiator;
 import org.jboss.weld.injection.producer.BasicInjectionTarget;
+import org.jboss.weld.injection.producer.ConstructorInterceptionInstantiator;
 import org.jboss.weld.injection.producer.DefaultInstantiator;
+import org.jboss.weld.injection.producer.Instantiator;
 import org.jboss.weld.injection.producer.InterceptionModelInitializer;
+import org.jboss.weld.injection.producer.InterceptorApplyingInstantiator;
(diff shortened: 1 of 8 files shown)


tremes:
retest this please

modules/ejb/src/main/java/org/jboss/weld/module/ejb/EjbSupportImpl.java
```diff
@@ -70,13 +73,25 @@ public void cleanup() {
...
             instantiator = new DefaultInstantiator<T>(type, null, manager);
         } else {
             // Session bean subclassed by the EJB container
             instantiator = SubclassedComponentInstantiator.forSubclassedEjb(type, implementationClass, null, manager);
         }
+        InterceptionModel interceptionModel = manager.getInterceptorModelRegistry().get(type.slim());
+        if (interceptionModel != null) {
+            if (interceptionModel.hasExternalNonConstructorInterceptors()) {
+                instantiator = SubclassedComponentInstantiator
+                        .forInterceptedDecoratedBean(implementationClass, null, (AbstractInstantiator<T>) instantiator, manager);
```
manovotn: I think the `null` value here is OK as the MDB is not a bean but a mere injection target. Also a quick glance at the code shows it takes `null` into consideration and CI doesn't complain.

modules/ejb/src/main/java/org/jboss/weld/module/ejb/EjbSupportImpl.java
```diff
@@ -70,13 +73,25 @@ public void cleanup() {
...
+                instantiator = new InterceptorApplyingInstantiator<>(instantiator, interceptionModel, type.slim());
+
+            }
+            if (interceptionModel.hasExternalConstructorInterceptors()) {
```
manovotn: I wonder if `@AroundConstruct` makes any sense for MDB? We don't control it's lifecycle and don't really know what EJB container chooses to do with it and when. But perhaps I am missing something?

manovotn: Ok, we cleared this out - it makes sense but it's up to integrators which should use Weld SPI to enable this.

PR #1708, opened 2017-07-19
WELD-2408 Beans.xml with only comment will be treated as an empty bea…

manovotn:
…ns.xml

_Technically speaking, XML containing only comments is invalid_ (according to w3c scheme which we use) but so is empty XML file and CDI spec requires empty file to work.
The question is then, what does empty mean? No chars at all, or no meaningful elements?

I just played around with what we could do to make this work, not sure we should do it at all. Opinions are welcome.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
+++ b/impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
@@ -18,9 +18,12 @@ package org.jboss.weld.xml;
 
 import static org.jboss.weld.bootstrap.spi.BeansXml.EMPTY_BEANS_XML;
 
+import java.io.BufferedReader;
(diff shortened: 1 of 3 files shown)


manovotn:
retest this please

rdebusscher:
I agree that a file containing only a comment is strictly speaking not valid.

But since an empty file is supported, it seems logical to me that comments are ignored to determine if the file is empty or not (just like comments are ignored by the Java compiler)

mkouba:
I tend to disagree with this solution. An empty file is just a **marker**. Once you add a content it must be parsed and so I think it's not a problem when it fails for an invalid xml file. Using `<!-- empty beans xml --> <beans/>` is the way to go.

mkouba:
WELD-2408 was rejected.

PR #1709, opened 2017-07-20
WELD-2397 Semi-automate WFLY patch generation for Weld 3.

manovotn:


manovotn:
This **NOT** equal to the patching for 2.4 branch. It cannot be fully automated due to patched WFLY internals which we cannot simply replace. Therefore I went for this version which allows you to:

* Download WFLY from mvn central with `-Pdownload-wfly`
  * You can specify version with `-Dwildfly.download.version` but it defaults to the one we support (10.1.0.Final)
  * **WildFly 11 is not in the Central yet - this profile won't be able to download it**
    * We may use wget, but such WFLY won't be cached in your mvn repo and will be re-downloaded repeatedly
* With `-Pupdate-jboss-as` you are patching whatever is in `JBOSS_HOME`
  * **This is the different part - you have to supply WFLY patched to some version of Weld 3 yourself and set `JBOSS_HOME` to point at it**
  * This will still update such a WFLY with latest snapshot of Weld (assuming you are on master branch), but all it really does is replacement of Weld JARs
* With `-Pwfly-patch-gen` you:
  * Firstly, download patch file from our build repository. You can specify version but it again defaults to what we normally use (patch for WFLY 10 with Weld 3).
   * Secondly, you trigger patch generation process which uses the two folders with WFLY you downloaded/patched and the patchfile you just grabbed.
  * This profile can be invoked as a standalone one, provided that you supply all arguments!

If you do:

* `export JBOSS_HOME=/path/to/wfly; mvn clean install -Pupdate-jboss-as,download-wfly,wfly-patch-gen`
  * In your target folder you get pristine WFLY, patch file (xml) and a complete patch.zip
  * Your `JBOSS_HOME` will be patched
* `export JBOSS_HOME=/path/to/wfly; mvn clean install -Pupdate-jboss-as,download-wfly`
  * You will end up with pristine WFLYin your target folder
  * Your `JBOSS_HOME` will be patched
* `export JBOSS_HOME=/path/to/wfly; mvn clean install -Pupdate-jboss-as`
  * You can patch your server the way we used it before

@tremes Mind testing it out a bit? I OFC did that too, but just to be sure I haven't missed something obvious.

tremes:
Yes I'll experiment with it little bit.

manovotn:
Agreed, I was thinking along the same lines. Let's just wait what Martin says to this and if he agrees I can basically put my former comment into the README to reflect the options.

mkouba:
Looks good. @manovotn pls add some info to README.

PR #1711, opened 2017-07-25
WELD-2407 fix selfinvocation of non-intercepted default method and

tremes:
expand SelfInvocationTest.


tremes:
Just realized that we probably want to keep these related tests in dedicated `jdk8` directory or?

mkouba:
Yes, we also test 2.4 against Java7.

PR #1712, opened 2017-07-25
WELD-2405 Do not optimize self invocation for default methods

mkouba:
This PR is based on https://github.com/weld/core/pull/1703 from @hypnoce:
- redundant test removed
- we must obtain the current stack first otherwise
`DecoratedInteraceWithDefaultMethodTest` fails


hypnoce:
:+1:

PR #1713, opened 2017-07-25
WELD-1914 Allow to recognize Weld-generated classes and regain contex…

manovotn:
…tual instance from client proxies."

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyFactory.java
@@ -36,6 +36,8 @@ import org.jboss.classfilewriter.code.BranchEnd;
 import org.jboss.classfilewriter.code.CodeAttribute;
 import org.jboss.weld.Container;
 import org.jboss.weld.bean.proxy.util.SerializableClientProxy;
+import org.jboss.weld.exceptions.WeldException;
+import org.jboss.weld.proxy.WeldClientProxy;
 import org.jboss.weld.security.GetDeclaredFieldAction;
 import org.jboss.weld.security.SetAccessibleAction;
 import org.jboss.weld.serialization.spi.BeanIdentifier;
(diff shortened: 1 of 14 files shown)


tests-arquillian/src/test/java/org/jboss/weld/tests/internal/contructs/BeansImplementWeldInternalConstructsTest.java
```diff
@@ -0,0 +1,145 @@
...
+        // cast to WeldClientProxy and test the methods
+        WeldClientProxy wcp = (WeldClientProxy) clientProxyBean;
+        
+        Object contextualInstance = wcp.getContextualInstance();
+        // kind of indirect check that this is the actual contextual instance
+        Assert.assertTrue(contextualInstance instanceof ClientProxyBean);
+        Assert.assertFalse(contextualInstance instanceof WeldConstruct);
+        
+        // TODO how to verify this
+        Bean<?> bean = wcp.getBean();
```
mkouba: Just get the bean for `ClientProxyBean` from `BeanManager`, cast to `WeldBean` and compare the bean identifiers?

manovotn:
The API PR draft for this -> https://github.com/weld/api/pull/63

manovotn:
ATM there is a test failure with serialization, I have yet to explore why - `BeanMetadataPassivationCapableDependencyTest.testDefaultBeanMetadataPassivationCapableDependency`

manovotn:
To sum up - in the current state the tests pass on embedded container.
WFLY requires few more modules to explicitly depends on Weld API, then it passes as well. Those modules are namely:
 - org/hibernate/validator/main
 - com/sun/jsf-impl
 - org/jboss/resteasy/resteasy-cdi

The above list may or may not be complete, it is just what our testsuite discovered. Simply said, any module which wants to create some proxy objects will need to depends on Weld API so that it has access to new interface classes.

mkouba:
I think we could avoid the method names conflicts. See also how `org.jboss.weld.interceptor.util.proxy.TargetInstanceProxy` methods are implemented:
* https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java#L821
* https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyMethodHandler.java#L81

WRT WildFly - in fact, no module should depend on `weld-core` directly, but `weld-api` and `weld-spi` instead. Anyway, I believe adding weld-api dependency is not a problem.

manovotn:
@mkouba Changed both PRs to use `Metadata`. Corrected `ClientProxyFactory` to use `interfaces.add` and also fixed the order in which I generate methods on proxy  (`addMethods` - original ones first, then the new one).

I retested everything on embedded container and WFLY and it passed.

manovotn:
Closing in favor of https://github.com/weld/core/pull/1719

PR #1720, opened 2017-08-21
Fix test deployment in ApplicationContextInitializedEventFiredWithNoW…

manovotn:
…ebArchiveTest.


manovotn:
This caused troubles in our CI jobs but I couldn't reproduce it locally.
Either way, in accordance with EE "visibility" rules, the test class should reside in EJB archive.

PR #1724, opened 2017-09-18
WELD-2414 Expand current test suite to cover this case.

manovotn:


manovotn:
@mkouba since we already have this test in the testsuite, it seemed easier to just expand it instead of writing a new one.

manovotn:
Closing in favor of https://github.com/weld/core/pull/1725

PR #1725, opened 2017-09-18
WELD-2414 Fix InterceptedSubclassFactory

mkouba:
- also expand InterceptorBridgeMethodTest

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedSubclassFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedSubclassFactory.java
@@ -25,6 +25,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.security.AccessController;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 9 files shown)


impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedSubclassFactory.java
```diff
@@ -437,4 +450,61 @@ private void createDelegateMethod(ClassFile proxyClassType, Method method, Metho
         createDelegateToSuper(delegatingMethod, methodInformation);
     }
 
-}
+    private static class BridgeMethod {
```
manovotn: I know it more or less doesn't matter, but wouldn't it be a tad bit nicer to have this as a separate class file which `extends MethodSignature`?

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/bridgemethods/hierarchy/InterceptorBridgeMethodTest.java
```diff
@@ -10,43 +26,150 @@
...
+        reset();
+        special.getB();
+        verify(SpecialParent.class);
+        reset();
+        special.invokeDefault("foo");
+        verify(Base.class);
+    }
+
+    @Test
+    @Ignore("WELD-2424")
```
manovotn: Just a side note - once we merge this, we should link this test to WELD-2424 issue so that it's clear how the use case looks like. Don't know about you, but I sure won't remember there was a test added back here :)

impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedSubclassFactory.java
mkouba: We can make it separate once it is usable outside `InterceptedSubclassFactory`. Also extending `MethodSignature` does not bring any value here. And by the way, it's not a "method signature" as it contains the return type ;-)

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/bridgemethods/hierarchy/InterceptorBridgeMethodTest.java
mkouba: Yes, of course. I will add a link to this line once the PR is merged.

impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedSubclassFactory.java
manovotn: That's why it would be an 'extended' method signature, right? :-P

PR #1737, opened 2017-10-04
WELD-2429 Test and fix

manovotn:


mkouba:
Throw `CreationException` in `SessionBeanImpl.create(CreationalContext<T>)` if `SessionBeanImpl.initializeAfterBeanDiscovery()` was not called before.

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
```diff
@@ -526,4 +526,6 @@
     @Message(id = 1571, value = "Proxy for {0} created in {1} because {2}.", format = Format.MESSAGE_FORMAT)
     void generatingProxyToDefaultPackage(Object param1, Object param2, Object param3);
 
+    @Message(id = 1572, value = "Cannot create instance of session bean from Annotated Type {0} in AfterBeanDiscovery phase.", format = Format.MESSAGE_FORMAT)
```
mkouba: Strictly speaking it's not possible to create the instance **before `AfterDeploymentValidation`** not only during `AfterBeanDiscovery`.

PR #1739, opened 2017-10-09
WELD-2430 Throw EJBException when trying to invoke non-public method …

manovotn:
…on EJB bean.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/EnterpriseBeanProxyMethodHandler.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/EnterpriseBeanProxyMethodHandler.java
@@ -19,9 +19,12 @@ package org.jboss.weld.bean.proxy;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+
+import javax.ejb.EJBException;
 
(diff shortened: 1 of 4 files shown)


manovotn:
To keep the confusion to minimum I basically copied over the exception coming from WFLY. With the test (assuming you remove the catch block) you now get: `WELD-001573: Not a business method java.lang.String org.jboss.weld.tests.ejb.stateless.noInterfaceNonPublic.ProblematicBean.ping(). Do not call non-public methods on EJBs.`

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
```diff
@@ -505,4 +507,7 @@
 
     @Message(id = 1572, value = "Cannot create instance of session bean from Annotated Type {0} before AfterDeploymentValidation phase.", format = Format.MESSAGE_FORMAT)
     CreationException initABDnotInvoked(Object bean);
+
+    @Message(id = 1573, value = "Not a business method {0}. Do not call non-public methods on EJB's.", format = Format.MESSAGE_FORMAT)
+    EJBException nonPublicMethodInvoked(Object method);
```
mkouba: Note that EJB API is an optional dependency and `BeanLogger` is used in SE and other environments. So I think we should not import this class here.

manovotn: Agreed, moving this to `EnterpriseBeanProxyMethodHandler`

mkouba:
Looks good.

PR #1741, opened 2017-10-16
WELD-2432 Probe - fix wrongly recognized alternative bean NPE.

manovotn:


mkouba:
I believe we should extend https://github.com/weld/core/blob/master/probe/tests/src/test/java/org/jboss/weld/probe/tests/integration/ProbeDeploymentsTest.java to verify the behavior.

PR #1743, opened 2017-10-16
WELD-2433 + WELD-2436 Backport WeldInvocationContext and move key bindings to to API

manovotn:


manovotn:
@mkouba backported tests work for me. I also ran full test suite against patched wfly 10 - that one passed as well.

impl/src/main/java/org/jboss/weld/interceptor/proxy/InterceptorMethodHandler.java
```diff
@@ -27,13 +27,15 @@
 public class InterceptorMethodHandler implements StackAwareMethodHandler, Serializable {
 
     private static final long serialVersionUID = 1L;
+    public static final String INTERCEPTOR_BINDINGS_KEY = "org.jboss.weld.interceptor.bindings";
```
manovotn: This could be moved to API (if we do the API change for 2.4). In any case we should document this.

mkouba: https://issues.jboss.org/browse/WELD-2436

PR #1747, opened 2017-11-01
WELD-2436 Move interceptors binding key to API.

manovotn:


mkouba:
I think we should also modify [InvocationContextInterceptorBindingsTest](https://github.com/weld/core/blob/master/tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/context/bindings/InvocationContextInterceptorBindingsTest.java).

PR #1750, opened 2017-11-02
WELD-2427 WeldInstance#select and WeldEvent#select with Type.

manovotn:


impl/src/main/java/org/jboss/weld/resolution/ResolvableBuilder.java
```diff
@@ -50,7 +51,8 @@
 
 public class ResolvableBuilder {
 
-    private static final Class<?>[] FACADE_TYPES = new Class<?>[] { Event.class, Instance.class, WeldInstance.class, Provider.class, InterceptionFactory.class };
+    private static final Class<?>[] FACADE_TYPES = new Class<?>[] { Event.class, Instance.class, WeldEvent.class,
```
manovotn: Not really sure we need to change this?

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/event/enhanced/WeldEventTest.java
```diff
@@ -0,0 +1,172 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ *
+ * @author <a href="mailto:manovotn@redhat.com">Matej Novotny</a>
+ */
+@RunWith(Arquillian.class)
+public class WeldEventTest {
```
manovotn: The tests are verifying:
* ability to `@Inject WeldEvent<X>`
* known functionality when replacing `Event` with `WeldEvent`
* new selections based on `Type` and expected exceptions

I am accepting additional ideas for corner cases and/or possible problematic areas.

impl/src/main/java/org/jboss/weld/resolution/ResolvableBuilder.java
mkouba: I believe we need this.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/event/enhanced/WeldEventTest.java
```diff
@@ -0,0 +1,172 @@
...
+            try {
+                // using Type from List<Object> -> List<?>
+                bean.getEventListObject().select(listWildcardType);
+                Assert.fail();
+            } catch (IllegalStateException e) {
+                // expected
+            }
+
+            // verify valid behaviour, following selections should work
+            bean.getEventObject().select(firstType);
```
mkouba: We should probably also test observability here... i.e. that a child `Event` works as expected?

impl/src/main/java/org/jboss/weld/bean/builtin/InstanceImpl.java
```diff
@@ -151,9 +151,15 @@ public boolean isUnsatisfied() {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public <X> WeldInstance<X> select(Type subtype, Annotation... qualifiers) {
-        // TODO Auto-generated method stub
-        return null;
+        // verify if this was invoked on WeldInstance<Object>
+        if (!this.getType().getTypeName().equals(Object.class.getName())) {
```
mkouba: `!this.getType().equals(Object.class)` should be fine...

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/instance/enhanced/selectType/WeldInstanceSelectTypeTest.java
```diff
@@ -0,0 +1,84 @@
...
+    public static Archive<?> createTestArchive() {
+        return ClassPath.builder().add(ShrinkWrap.create(BeanArchive.class, Utils.getDeploymentNameAsHash(WeldInstanceSelectTypeTest.class))
+            .addPackage(WeldInstanceSelectTypeTest.class.getPackage())).build();
+    }
+
+    @Test
+    public void testValidSelect() {
+        try (WeldContainer container = new Weld().initialize()) {
+            // Class<T> implements Type, so let's make use of that
+            Type firstType = SomeInterface.class;
```
mkouba: Maybe we should add some tests for parameterized types...

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/instance/enhanced/selectType/WeldInstanceSelectTypeTest.java
```diff
@@ -0,0 +1,84 @@
...
+            try {
+                someInterfaceInstance.select(secondType);
+                Assert.fail("Selecting by type from any other type than Object should fail with IllegalStateException!");
+            } catch (IllegalStateException e) {
+                // expected
+            }
+        }
+    }
+    
+    // TODO add tests which will use a @Typed bean eliminating Object as bean type
```
mkouba: I would also add a test that `WeldInstance.destroy()` works if you create a new `@Dependent` bean instance with the new `select()` method.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/instance/enhanced/selectType/WeldInstanceSelectTypeTest.java
manovotn: Hmm, can add that too. Though there isn't much of a change here (as opposed to WeldEvent where I have such test). You could have done the same before with WeldInstance.

impl/src/main/java/org/jboss/weld/bean/builtin/InstanceImpl.java
manovotn: Hmm yep, it eluded me yet again that Class is in fact Type :)

impl/src/main/java/org/jboss/weld/resolution/ResolvableBuilder.java
manovotn: I _think_ it's for when you resolve bean via BM (why would you do that with Event though?), but I am not really sure. If I remove it `@Inject` will keep working.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/event/enhanced/WeldEventTest.java
manovotn: See `testEventObservability` method

mkouba: Ah, you're right ;-)

manovotn:
Added tests for `@PreDestroy` and `@Typed` and one simple valid select for `ParameterizedType`

PR #1752, opened 2017-11-03
 WELD-2258 InstanceImpl - do not cache priority comparator

mkouba:
I think there is no need to cache the priority comparator per `Instance`. We don't expect intensive use of this functionality.


manovotn:
Makes sense, but note that we could also store it on first usage. Don't think there is much difference between either approach.

mkouba:
Yet another lazy init ;-)

PR #1753, opened 2017-11-03
Fix API version in BOM.

manovotn:
I realized I forgot to upgrade API version in BOM.


mkouba:
No problem. This is only important during release and we will use 3.0.SP2 at that time...

PR #1755, opened 2017-11-08
WELD-2440 Document WeldInstance/WeldEvent#select.

manovotn:


docs/reference/src/main/asciidoc/events.asciidoc
```diff
@@ -419,3 +419,30 @@ public class Catalog {
...
+=== Enhanced version of `javax.enterprise.event.Event`
+
+Weld API offers slight advantage when dealing with events - `org.jboss.weld.events.WeldEvent`, an augmented version of `javax.enterprise.event.Event`.
+
+Currently there is only one additional method, `select(Type type, Annotation... qualifiers)`.
+This method allows to perform well-known selection with `java.lang.reflect.Type` as parameter which means things can get pretty generic.
+While there are no limitations to what you can select, there are limitation to the `WeldEvent` instance you perform selection on.
+In order to preserve type-safety, you have to invoke this method on `WeldInstance<Object>`.
+Using any other parameter will result in `IllegalStateException`.
+Here is a snippet borrowed over from Camel (and slightly modified):
```
manovotn: Can add link to that Camel code, or we can come up with different sample but this seemed to be justifying sample for this feature.

docs/reference/src/main/asciidoc/injection.asciidoc
```diff
@@ -614,10 +614,40 @@ class OrderService {
...
   }
-}
 ----------------------------------------------------------------------------------
 
+Third method is a twist on the `select()` method, but it accepts `java.lang.reflect.Type` as parameter and optionally qualifier(s).
+This allows for generic selection of instances which can be handy while dealing with third party beans through extensions.
+However, in order to stay type-safe, this method has a limitation - it can only be invoked on `WeldInstance<Object>`.
+Invocation on any other type than `Object` will result in an  `IllegalStateException`.
+Please note that the return value if such select will always be `WeldInstance<Object>` unless you specify it further using `<SomeType>` before invoking this `select()`.
+Let's look at actual code:
```
manovotn: @mkouba if you have any better sample to present, I am all ears. I was unable to come up with other simple-enough case for this use case.

docs/reference/src/main/asciidoc/events.asciidoc
mkouba: I don't like this example. It contains a lot of unknown classes. On the other hand a link to the Camel integration code seems reasonable. I think the snippet could be rather general, just a method which accepts `Type` and fires an event.

manovotn: The Camel code contains the old hacky version ATM, so just link won't do.
Also plain sample with `Type` doesn't really show what is this method good for. E.g. the case when you simply don't know or care for the actual class of event.

PR #1764, opened 2018-01-04
WELD-2445 Allow to recognize beans 1.1 xsd.

manovotn:
I don't like it much but it's the best idea we came up with ATM.

Also, I am still fishing for actual test for this.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
+++ b/impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
@@ -18,9 +18,14 @@ package org.jboss.weld.xml;
 
 import static org.jboss.weld.bootstrap.spi.BeansXml.EMPTY_BEANS_XML;
 
+import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
 import java.net.URL;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


manovotn:
BTW I tried the semi-test of using `beans.xml` with version 1.1 and using `<trim/>` at the same time. But Weld will simply log error and process the file anyway (handling it as trimmed archive). I am kinda out of ideas for tests.

impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
```diff
@@ -178,11 +196,12 @@ private boolean alreadyAdded(Metadata<String> metadata, List<Metadata<String>> l
...
     }
 
-    private static InputSource[] loadXsds() {
+    private static InputSource[] loadXsds(String beansXmlAsString) {
 
         List<InputSource> xsds = new ArrayList<InputSource>();
+        String beans11 = "beans_1_1.xsd";
 
-        for (XmlSchema schema : XmlSchema.values()) {
+        for (XmlSchema schema : XmlSchema.getSchemas(beansXmlAsString != null && beansXmlAsString.contains(beans11))) {
```
mkouba: The `beans11` variable is not needed - we could use `XmlSchema.CDI11.getFileName()` instead.

impl/src/main/java/org/jboss/weld/xml/XmlSchema.java
```diff
@@ -16,11 +16,14 @@
...
 
+import java.util.ArrayList;
+import java.util.List;
+
 import javax.enterprise.inject.spi.BeanManager;
 
 public enum XmlSchema {
 
-    CDI10("beans_1_0.xsd", BeanManager.class.getClassLoader()), CDI20("beans_2_0.xsd", BeanManager.class.getClassLoader()), WELD11("weld_1_1.xsd", BeansXmlParser.class.getClassLoader());
+    CDI10("beans_1_0.xsd", BeanManager.class.getClassLoader()), CDI11("beans_1_1.xsd", BeanManager.class.getClassLoader()), CDI20("beans_2_0.xsd", BeanManager.class.getClassLoader()), WELD11("weld_1_1.xsd", BeansXmlParser.class.getClassLoader());
```
mkouba: Hm, all the values use the same CL, maybe we could remove the arg and use `BeanManager.class.getClassLoader()` by default?

manovotn: `WELD11` uses different CL, so I kept it as is.

PR #1765, opened 2018-01-08
WELD-2447 Make proxy serialization container agnostic.

manovotn:
Still missing a test for this, not sure how to "easily" test it yet.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyFactory.java
@@ -69,10 +69,12 @@ public class ClientProxyFactory<T> extends ProxyFactory<T> {
      * field.
      */
     private static final String BEAN_ID_FIELD = "BEAN_ID_FIELD";
+    private static final String CONTEXT_ID_FIELD = "CONTEXT_ID_FIELD";
 
     private final BeanIdentifier beanId;
 
     private volatile Field beanIdField;
+    private volatile Field contextIdField;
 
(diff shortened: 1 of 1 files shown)


impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyFactory.java
```diff
@@ -69,14 +69,18 @@
      * field.
      */
     private static final String BEAN_ID_FIELD = "BEAN_ID_FIELD";
+    private static final String CONTEXT_ID_FIELD = "CONTEXT_ID_FIELD";
 
     private final BeanIdentifier beanId;
+    private final String contextId;
```
mkouba: No need to duplicate `org.jboss.weld.bean.proxy.ProxyFactory.contextId`...

impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyFactory.java
```diff
@@ -88,7 +92,13 @@ public T create(BeanInstance beanInstance) {
                 AccessController.doPrivileged(SetAccessibleAction.of(f));
                 beanIdField = f;
             }
+            if (contextIdField == null) {
+                final Field f = AccessController.doPrivileged(new GetDeclaredFieldAction(instance.getClass(), CONTEXT_ID_FIELD));
+                AccessController.doPrivileged(SetAccessibleAction.of(f));
+                contextIdField = f;
+            }
             beanIdField.set(instance, beanId);
+            contextIdField.set(instance, contextId);
```
mkouba: Use `ProxyFactory.getContextId()` instead of `contextId`...

PR #1770, opened 2018-01-15
WELD-2448: Fix shared proxy class instantiation

mkouba:
- if a shared proxy class was created with a nonmatching type of
ProxyInstantiator then try to use a different one

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyInstantiator.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyInstantiator.java
@@ -81,24 +81,14 @@ public interface ProxyInstantiator extends Service {
         }
 
         /**
-         * Obtains new ProxyInstantiator based on given {@link WeldConfiguration}.
+         * Obtains a ProxyInstantiator instance, {@link ConfigurationKey#PROXY_INSTANTIATOR} is not taken into account.
          *
          * @param configuration
-         * @return
+         * @return proxy instantiator
          */
-        @SuppressWarnings("deprecation")
-        public static ProxyInstantiator create(WeldConfiguration configuration) {
+        public static ProxyInstantiator create(boolean relaxedConstruction) {
             ProxyInstantiator result = DefaultProxyInstantiator.INSTANCE;
-            final String instantiator = configuration.getStringProperty(ConfigurationKey.PROXY_INSTANTIATOR);
-            if (!instantiator.isEmpty()) {
-                if (!DefaultProxyInstantiator.class.getName().equals(instantiator)) {
-                    try {
-                        result = newInstance(instantiator);
-                    } catch (Exception e) {
-                        throw new WeldException(e);
-                    }
-                }
-            } else if (configuration.getBooleanProperty(ConfigurationKey.RELAXED_CONSTRUCTION) || configuration.getBooleanProperty(ConfigurationKey.PROXY_UNSAFE)) {
+            if (relaxedConstruction) {
                 for (String implementation : IMPLEMENTATIONS) {
                     // use the first suitable implementation
                     try {
@@ -110,6 +100,33 @@ public interface ProxyInstantiator extends Service {
                         BootstrapLogger.LOG.catchingDebug(e);
                     }
                 }
+            }
+            return result;
+        }
+
+        /**
+         * Obtains a ProxyInstantiator based on given {@link WeldConfiguration}.
+         *
+         * @param configuration
+         * @return proxy instantiator
+         */
+        @SuppressWarnings("deprecation")
+        public static ProxyInstantiator create(WeldConfiguration configuration) {
+            ProxyInstantiator result = null;
+            String instantiator = configuration.getStringProperty(ConfigurationKey.PROXY_INSTANTIATOR);
+            if (!instantiator.isEmpty()) {
+                if (!DefaultProxyInstantiator.class.getName().equals(instantiator)) {
+                    try {
+                        result = newInstance(instantiator);
+                    } catch (Exception e) {
+                        throw new WeldException(e);
+                    }
+                } else {
+                    result = DefaultProxyInstantiator.INSTANCE;
+                }
+            } else {
+                result = create(configuration.getBooleanProperty(ConfigurationKey.RELAXED_CONSTRUCTION)
+                        || configuration.getBooleanProperty(ConfigurationKey.PROXY_UNSAFE));
             }
             BootstrapLogger.LOG.debugv("Using instantiator: {0}", result.getClass().getName());
             return result;
@@ -126,4 +143,5 @@ public interface ProxyInstantiator extends Service {
             return clazz.newInstance();
         }
     }
+
 }
--- a/impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
+++ b/impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
@@ -538,4 +538,8 @@ public interface BeanLogger extends WeldLogger {
 
     @Message(id = 1575, value = "WeldInstance.select(Type subtype, Annotation... qualifiers) can be invoked only on an instance of WeldInstance<Object>.", format = Format.MESSAGE_FORMAT)
     IllegalStateException selectByTypeOnlyWorksOnObject();
+
+    @LogMessage(level = Level.DEBUG)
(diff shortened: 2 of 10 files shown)


impl/src/main/java/org/jboss/weld/bean/proxy/ProxyInstantiator.java
```diff
@@ -81,24 +81,14 @@ private Factory() {
         }
 
         /**
-         * Obtains new ProxyInstantiator based on given {@link WeldConfiguration}.
+         * Obtains a ProxyInstantiator instance, {@link ConfigurationKey#PROXY_INSTANTIATOR} is not taken into account.
```
manovotn: > `ConfigurationKey#PROXY_INSTANTIATOR` is not taken into account

I'm not sure it's correct to completely disregard this option. I know it's for testing and such only but maybe that's why it should work as an "override all" option?

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
```diff
@@ -538,4 +538,8 @@
 
     @Message(id = 1575, value = "WeldInstance.select(Type subtype, Annotation... qualifiers) can be invoked only on an instance of WeldInstance<Object>.", format = Format.MESSAGE_FORMAT)
     IllegalStateException selectByTypeOnlyWorksOnObject();
+
+    @LogMessage(level = Level.DEBUG)
+    @Message(id = 1576, value = "Using {1} for a shared proxy class {0}; the current impl [{2}] does not match the instantiator the proxy was created with", format = Format.MESSAGE_FORMAT)
```
manovotn: Maybe a little twist?
`Using {1} to instantiate a shared proxy class {0}; former impl [{2}] does not...`

I was also thinking whether we couldn't be more precise and suggest that there are multiple deployments, some of which use relaxed construction while other don't. Or do you think there are other cases where this can happen?
While the log msg tells us something, I guess Average Joe won't have a clue what's happening there.

tests-arquillian/src/test/java/org/jboss/weld/tests/jsf/flow/FlowMultipleWarTest.java
```diff
@@ -0,0 +1,117 @@
...
+                        PropertiesBuilder.newBuilder()
+                                .set(ConfigurationKey.RELAXED_CONSTRUCTION.get(), "true").build(),
+                        "weld.properties");
+    }
+
+    @ArquillianResource
+    Deployer deployer;
+
+    /**
+     * This is not a real test method - we just need to deploy WAR1 first.
```
manovotn: I think you could just use `@Before` method.

tests-arquillian/src/test/java/org/jboss/weld/tests/jsf/flow/FlowMultipleWarTest.java
```diff
@@ -0,0 +1,117 @@
...
+        // In order to use @ArquillianResource URLs we need to deploy both test archives first
+        deployer.deploy(WAR1);
+        deployer.deploy(WAR2);
+    }
+
+    @Test
+    @RunAsClient
+    @InSequence(2)
+    public void testFlows(@ArquillianResource @OperateOnDeployment(WAR1) URL war1Context,
+            @ArquillianResource @OperateOnDeployment(WAR2) URL war2Context)
```
manovotn: Isn't this test missing the part where you invoke the same call from war2Context? Or how do you make sure the extension proxy get initiated from the other WAR as well?

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
mkouba: `former` is imho misleading in this use case, but +1 for `to instantiate a shared...`. I don't think we need to be more concrete here. First relaxed construction is an advanced feature. Also it's DEBUG level so I don't expect Average Joe to consume this message.

tests-arquillian/src/test/java/org/jboss/weld/tests/jsf/flow/FlowMultipleWarTest.java
mkouba: Probably, but only if you declare a single test method. This approach is imho more robust.

tests-arquillian/src/test/java/org/jboss/weld/tests/jsf/flow/FlowMultipleWarTest.java
mkouba: I think it should be initialized during bootstrap of a JSF app with flows.

manovotn: Hmm, makes sense.
If that's the case then the second method param is useless and can be removed?

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyInstantiator.java
mkouba: The problem with custom `ProxyInstantiator` is that it might not be available in all deployments. 

For sure this fix does not cover 100% of use cases but should work in most real scenarios.

tests-arquillian/src/test/java/org/jboss/weld/tests/jsf/flow/FlowMultipleWarTest.java
manovotn: `@Before` can be bound to each method, so it would work woth multiple methods as well.
BTW you might also want to call `deployer.undeploy()`. I think Arq. does it on its own, but it looks cleaner to handle it.

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
manovotn: ad `former` - how about `previous`? I just don't like the word `current` as it is current no longer if you see what I mean.

ad. Average Joe - I think even Advanced Joe would have hard times :)
But I am fine leaving it as it is.

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyInstantiator.java
manovotn: Actually, disregard this comment about disregarding the config option. If I think about it, there is no point in enabling the option in case of these "combined" deployments.

tests-arquillian/src/test/java/org/jboss/weld/tests/jsf/flow/FlowMultipleWarTest.java
mkouba: > @Before can be bound to each method, so it would work woth multiple methods as well.

Sure, but we only want to perform the deployment once.

> `deployer.undeploy()`

Good point.

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
mkouba: Hm, `current` means the "configured/default" `ProxyInstantiator` of the current deployment. `previous` is imho also misleading. We don't replace the original instantiator, just obtain a new one to service this particular proxy class.

tests-arquillian/src/test/java/org/jboss/weld/tests/jsf/flow/FlowMultipleWarTest.java
manovotn: I just wanted to avoid having this counted as a test method when it's actually a setup.

mkouba: I see. Actually, we can merge those two methods into one I think...

mkouba: Hm, probably not as `@ArquillianResource` would not work.

impl/src/main/java/org/jboss/weld/logging/BeanLogger.java
manovotn: I understand we don't replace it.
Then maybe `this deployment's impl...` or just `default`, `selected`?

tests-arquillian/src/test/java/org/jboss/weld/tests/jsf/flow/FlowMultipleWarTest.java
manovotn: You would need to know the URL yourself then, or extract it from somewhere.

PR #1773, opened 2018-01-17
WELD-2452 multiversion support for WildFly

manovotn:
This is a temporary re-introduction of some literals and code construct we need to support two CDI versions in WildFly. Once WFLY fully transfers to CDI 2.0 this PR is to be **reverted**.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/ContextHolder.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/ContextHolder.java
@@ -1,9 +1,8 @@
 package org.jboss.weld.bootstrap;
 
+import javax.enterprise.context.spi.Context;
 import java.lang.annotation.Annotation;
 import java.util.Set;
-
-import javax.enterprise.context.spi.Context;
 
 public class ContextHolder<T extends Context> {
 
--- a/impl/src/main/java/org/jboss/weld/bootstrap/events/ContainerLifecycleEvents.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/events/ContainerLifecycleEvents.java
@@ -200,7 +200,12 @@ public class ContainerLifecycleEvents extends AbstractBootstrapService {
             // FastProcessAnnotatedTypeResolver does not consider special scope inheritance rules (see CDI - section 4.1)
             if (checkScopeInheritanceRules(event.getOriginalAnnotatedType(), observer, beanManager)) {
                 try {
-                    observer.notify(event);
+                    //TODO: CDI 1.1 HACK remove once no longer required
+                    if(observer instanceof ObserverMethodImpl) {
+                        ((ObserverMethodImpl)observer).notify(event);
+                    } else {
+                        observer.notify(event);
+                    }
                 } catch (Throwable e) {
                     errors.add(e);
                 }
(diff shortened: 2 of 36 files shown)


manovotn:
retest this please

manovotn:
The PR should now contain all the changes needed.
In TCK tests, it automatically uses EE 8 mode for TCK 2.0 and disables it for TCK 1.2.
Our integration tests and probe tests use EE 8 mode only and by default enable it (there are no other tests starting WFLY I hope).

impl/src/main/java/org/jboss/weld/bootstrap/ContextHolder.java
```diff
@@ -14,7 +18,8 @@
     public ContextHolder(T context, Class<T> type, Set<Annotation> qualifiers) {
         this.context = context;
         this.type = type;
-        this.qualifiers = qualifiers;
+        this.qualifiers = asSet(DefaultLiteral.INSTANCE, AnyLiteral.INSTANCE);
```
mkouba: This is likely unnecessary. The set of qualifiers is expected to contain all elements. See also `org.jboss.weld.bootstrap.WeldStartup.createContexts(ServiceRegistry)` and other places where this constructor is used.

impl/src/main/java/org/jboss/weld/bootstrap/events/ContainerLifecycleEvents.java
```diff
@@ -200,7 +200,12 @@ private void fireProcessAnnotatedType(ProcessAnnotatedTypeImpl<?> event, Set<Con
             // FastProcessAnnotatedTypeResolver does not consider special scope inheritance rules (see CDI - section 4.1)
             if (checkScopeInheritanceRules(event.getOriginalAnnotatedType(), observer, beanManager)) {
                 try {
-                    observer.notify(event);
+                    //TODO: CDI 1.1 HACK remove once no longer required
```
mkouba: We should add the issue number (WELD-2452) instead of TODO and possibly add some brief description of what the hack actually does...

impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/ObserverMethodConfiguratorImpl.java
```diff
@@ -141,7 +142,10 @@ public ObserverMethodConfiguratorImpl(ObserverMethod<T> observerMethod, Extensio
         reception(observerMethod.getReception());
         transactionPhase(observerMethod.getTransactionPhase());
         priority(observerMethod.getPriority());
-        async(observerMethod.isAsync());
+        //TODO: CDI 1.1 HACK remove once no longer required
```
mkouba: The same as https://github.com/weld/core/pull/1773/files#r162912259

impl/src/main/java/org/jboss/weld/bootstrap/events/ContainerLifecycleEvents.java
manovotn: Good point, I actually just took Stuart's commit and added my stuff there so I missed this.

impl/src/main/java/org/jboss/weld/bootstrap/ContextHolder.java
manovotn: You are likely right, I just meant to be rather safe then sorry, I will remove it and we'll see how it goes.

impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/ObserverMethodConfiguratorImpl.java
```diff
@@ -38,6 +38,7 @@
 import javax.enterprise.inject.spi.EventContext;
 import javax.enterprise.inject.spi.Extension;
 import javax.enterprise.inject.spi.ObserverMethod;
+//import javax.enterprise.inject.spi.Prioritized;
```
mkouba: Should be removed??

PR #1776, opened 2018-01-18
WELD-2453: Use suppressedExceptions for deployment/definition problems

mkouba:
- but only if multiple problems occur


gsmet:
@mkouba looks good to me. Thanks!

PR #1777, opened 2018-01-18
WELD-2451: Weld SE builder - make some members protected

mkouba:
- allow subclasses to use members related to Weld.createDeployment()


mkouba:
TBH I don't have a suggestion neither...

manovotn:
Hmm, then just keep it as it is I suppose

PR #1781, opened 2018-01-23
WELD-2456: Drop container lifecycle observer methods metadata at runtime

mkouba:


mkouba:
Unfortunately, some container lifecycle events must be fired for Java EE component classes, and this might occur even after Weld initialization is finished :-(

manovotn:
You mean EJB, right?
I don't know a way around it yet but let's leave it hanging here for a moment and see if we can come up with something.

mkouba:
In theory, we could try to alter WildFly integration to invoke `Bootstrap.endInitialization()` after all EE components are installed.

manovotn:
Hmm but other integrators might depends on this behaviour...

mkouba:
You're right. In that case, we would have to make this configurable through `org.jboss.weld.configuration.spi.ExternalConfiguration`.

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -768,7 +768,9 @@ public WeldContainer initialize() {
                 // weld-se uses CommonForkJoinPoolExecutorServices by default
                 .add(EXECUTOR_THREAD_POOL_TYPE.get(), COMMON.toString())
                 // weld-se uses relaxed construction by default
-                .add(ConfigurationKey.RELAXED_CONSTRUCTION.get(), true);
+                .add(ConfigurationKey.RELAXED_CONSTRUCTION.get(), true)
+                 // We do not expect any EE components in SE
```
manovotn: I am not sure this is good idea. People do all kinds of combinations and customizations and this *might* break their current apps. I would leave it as opt-in.
Or maybe...wasn't there an option to detect EE services somehow? So we could make assumptions on EE component presence if this option is unset?

impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
```diff
@@ -257,6 +261,12 @@
     @Description("If a non-empty string and development mode is enabled, the Probe data will be automatically exported after deployment validation. The value represents a path of the directory where to export the data file.")
     PROBE_EXPORT_DATA_AFTER_DEPLOYMENT("org.jboss.weld.probe.exportDataAfterDeployment", ""),
 
+    /**
+     * If set to true, {@link Bootstrap#endInitialization()} must be called after all EE components which support injection are installed (that means all
+     * relevant {@link ProcessInjectionTarget} events were already fired). This allows to perform effective cleanup and further optimizations.
+     */
+    END_INIT_AFTER_EE_COMPONENTS("org.jboss.weld.bootstrap.endInitAfterEEComponents", false, true),
```
manovotn: This option basically means Weld will attempt a cleanup; shouldn't the name reflect it?
Something like `CLEANUP_AFTER_INIT` with javadoc being:
`If set to true, the integrator confirms that {@link Bootstrap#endInitialization()} is invoked after all EE components supporting injection were installed (e.g. all their ProcessInjectionTarget events were fired). This allows Weld to perform a cleanup on metadata which are no longer needed as well as other minor optimizations.`

And even if we don't go for this I would like to stress you *integrator's responsibility* for this. In a sense that this optimization should not be used unless you made sure it makes sense.

mkouba: In fact, weld already does a lot of cleanup during `endInitialization()` so this merely a hint that a more thorough cleanup can be performed.

> I would like to stress you integrator's responsibility for this...

This config property can only be set through `ExternalConfiguration`.

mkouba: Maybe something like `ALLOW_OPTIMIZED_CLEANUP`?

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
mkouba: You're probably right. For Weld SE there should be at least a possibility to disable this "feature". My concern is that optional features are rarely used.

impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
manovotn: Yes, sounds better.

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
manovotn: Ok, in that case we should figure out how significant is this optimization. If it's worth it, we can have it as default and provide a way to turn it off. If it's minor, we should do it the other way around. I agree with your concern though.

mkouba: It's probably minor but I'd like to have it enabled by default.

mkouba:
@manovotn Ok, so I added some tests and the possibility to disable the optimized cleanup in Weld SE. Let me know what you think.

PR #1786, opened 2018-02-06
WELD-2462 Migrate from FindBugs to SpotBugs.

manovotn:
Replace FindBugs with SpotBugs.

SpotBugs found few more details in code which I adapted as part of this PR, feel free to disagree.

I will keep the profile name and threshold properties (etc.) named "findbugs" as it is well-known and many CI jobs use it.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/config/WeldConfiguration.java
+++ b/impl/src/main/java/org/jboss/weld/config/WeldConfiguration.java
@@ -123,8 +123,8 @@ public class WeldConfiguration implements Service {
         this.proxyDumpFilePath = initProxyDumpFilePath();
         this.proxyIgnoreFinalMethodsPattern = initProxyIgnoreFinalMethodsPattern();
         StringJoiner logOutputBuilder = new StringJoiner(", ", "{", "}");
-        for (ConfigurationKey key : properties.keySet()) {
-            logOutputBuilder.add(key.get() + "=" + properties.get(key));
+        for (Entry<ConfigurationKey, Object> entry : properties.entrySet()) {
+            logOutputBuilder.add(entry.getKey().get() + "=" + entry.getValue());
         }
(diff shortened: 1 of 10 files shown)


impl/src/main/java/org/jboss/weld/config/WeldConfiguration.java
```diff
@@ -123,8 +123,8 @@ public WeldConfiguration(ServiceRegistry services, Deployment deployment) {
         this.proxyDumpFilePath = initProxyDumpFilePath();
         this.proxyIgnoreFinalMethodsPattern = initProxyIgnoreFinalMethodsPattern();
         StringJoiner logOutputBuilder = new StringJoiner(", ", "{", "}");
-        for (ConfigurationKey key : properties.keySet()) {
-            logOutputBuilder.add(key.get() + "=" + properties.get(key));
+        for (Entry<ConfigurationKey, Object> entry : properties.entrySet()) {
+            logOutputBuilder.add(entry.getKey().get() + "=" + properties.get(entry.getKey()));
```
mkouba: `properties.get(entry.getKey())` should be `entry.getValue()`, otherwise I see no improvement but just an attempt to make spotbugs happy ;-)

impl/src/main/java/org/jboss/weld/util/reflection/Formats.java
```diff
@@ -536,7 +536,7 @@ public static String version(@Deprecated Package pkg) {
     public static String getSimpleVersion() {
         Properties buildProperties = getBuildProperties();
         if(buildProperties != null) {
-            buildProperties.getProperty(BUILD_PROPERTIES_VERSION);
+            return buildProperties.getProperty(BUILD_PROPERTIES_VERSION);
```
mkouba: Good catch!

PR #1791, opened 2018-02-15
WELD-2464: Prevent NPE in ProtectionDomainCache

Azquelt:
Prevent NullPointerException in ProtectionDomainCache if
ProtectionDomain.getPermissions() returns null.


mkouba:
test this please

Azquelt:
@mkouba You asked me to test this but then merged the PR. Did you want me to open a new PR to add a unit test?

It should be possible to unit test except that `getProtectionDomainForProxy` is a package-scoped method so the unit test would have to be in the same package or use reflection to call the method and I couldn't see any existing unit tests which do that.

I did test that the change fixes the issue we were seeing in OpenLiberty before submitting it.

mkouba:
@Azquelt `test this please` is a phrase that triggers our CI builds (see also https://wiki.jenkins.io/display/JENKINS/GitHub+pull+request+builder+plugin ;-). I don't think a test is needed.

Azquelt:
Got it, thanks for clearing up my confusion 😃

PR #1795, opened 2018-02-19
JDK 10 test setup

manovotn:


manovotn:
Putting on `hold` because of [ARQ-2175](https://issues.jboss.org/browse/ARQ-2175).
After the fix is accepted and released, we will then need to introduce Arq. BOM dependency in order to enforce latest Arq version with that fix.

manovotn:
retest this please

manovotn:
I am sadly getting some more intermittent failures in tck tests after upgrading deps.
Will see what PR executor gets...

mkouba:
I've tried to run one of the failed test locally (`ProcessInjectionPointFiredTest`) and the generated `arquillian-testng.jar` is missing TestNG classes (including `org.testng.IHookable` implemented by `org.jboss.arquillian.testng.Arquillian`).

mkouba:
See also https://issues.jboss.org/browse/SHRINKWRAP-519 and https://issues.jboss.org/browse/ARQ-2144.

manovotn:
So to sum up on this....
By upgrading to Arq. 1.4.0.Final we solve the problem with [exception transformers](https://issues.jboss.org/browse/ARQ-2175) but in the same time we introduce a new problem, [this one TestNG-related](https://issues.jboss.org/browse/ARQ-2144).

This cannot be solved on Arquillian side as it is a [bug coming from TestNG](https://github.com/cbeust/testng/issues/1677) itself (where it should be fixed for version 6.14.3). However, TestNG uses JCommander (currently 7.2) which has the [same problem](https://github.com/cbeust/jcommander/issues/430). This was reportedly fixed in JCommander 7.3 but you now get a brand `NoClassDefFound` error because they have [wrong manifest](https://github.com/cbeust/jcommander/issues/434).

Therefore, we need new version of JCommander, then TestNG has to upgrade to that JCOmmander (else you need to exclude jcommander from testng dep in poms) and release it and then, perhaps, it will work.

A cherry on top - if you were to try all of the above, you will have to add [their bintray repository](https://github.com/dipak-pawar/testng-classloading/blob/master/pom.xml#L14-L20) as none of the above releases is (for unfathomable reasons) in Maven Central.

Conclusion/TL&DR -> TestNG is a steaming pile of poop and it blocks this PR.

mkouba:
I still think that directory entries are optional in zip files. For example zip linux command has `-D` option: "do not create entries in the zip archive for directories". So even if TestNG is "fixed" the bug in ShrinkWrap `URLPackageScanner` imho remains and may cause issues.

manovotn:
But fixing https://issues.jboss.org/browse/SHRINKWRAP-519 alone won't be enough, or am I missing something?

manovotn:
retest this please

manovotn:
Looks like JDK 8 failures were CCI instability for the most part.
Which means these workarounds at least don't break current state.
Now to figure out what is it JDK 10 doesn't like...

manovotn:
JDK 10 cannot run spotbugs ATM, see also https://github.com/spotbugs/spotbugs/issues/593
Once there is a release we can upgrade and re-enable that in the Jenkins job.

manovotn:
retest this please

manovotn:
retest this please

manovotn:
The list of changes should be complete now...
We still need to do sanity check in regards to servlet testing and I am experiencing weird crashes with JDK 10 on Tomcat (when it passes on Jenkins).

However, embedded and incontainer testing of Weld should work with this on both, JDK 8 and 10. I tried adding as many comments as possible so that we know when it's safe to remove additional dependency declarations and such.

PR #1796, opened 2018-02-19
Update WildFly Arquillian version, get rid of old workaround.

manovotn:


manovotn:
Travis fails because this update would actually require JDK 8+ which we likely cannot have on Weld 2.4 (since it implements EE 7 and CDI 1.2).

manovotn:
We needn't solve this in 2.4 - closing PR.

PR #1797, opened 2018-02-23
WELD-2465 Switch to JBoss Commons Annotation 1.3 version.

manovotn:


pom.xml
```diff
@@ -62,6 +62,7 @@
         <jandex.version>2.0.3.Final</jandex.version>
         <javax.activation.version>1.1</javax.activation.version>
         <javax.el.version>3.0.0</javax.el.version>
+        <jboss.spec.annotations-api.version>1.0.0.Final</jboss.spec.annotations-api.version>
```
mkouba: This should be in Weld API pom: https://github.com/weld/api/blob/master/pom.xml#L108-L110

manovotn: Ah, good point, thanks

manovotn: Hmm, it seems like we had this dep duplicated in core and API...supposedly to avoid having to release API?

manovotn:
Requires API upgrade - see https://github.com/weld/api/pull/75

manovotn:
@mkouba this should be ready for merging.

PR #1800, opened 2018-03-07
Improved beans.xml parsing performance

ljnelson:
Hello; thanks as always for your work on Weld.

(I am something of a Github rookie when it comes to pull requests and such so I hope I've done this right.)

This pull request improves the performance of parsing `beans.xml` files.  The general approach is to share the `SAXParserFactory` (carefully!) and to buffer certain reads to avoid too many method invocations.  All edits were performed after profiling to make sure they had some kind of measurable effect.

Happy to make edits etc. but might require some hand-holding where git is concerned.  😄

### DIFF
--- a/impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
+++ b/impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
@@ -18,13 +18,14 @@ package org.jboss.weld.xml;
 
 import static org.jboss.weld.bootstrap.spi.BeansXml.EMPTY_BEANS_XML;
 
+import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.io.StringReader;
 import java.net.URL;
-import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


manovotn:
test this please

impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
```diff
@@ -59,10 +60,21 @@
...
+    private static final int BEANS_XML_BUFFER_SIZE = 1024;
+
+    static {
+        PARSER_FACTORY.setValidating(!SystemPropertiesConfiguration.INSTANCE.isXmlValidationDisabled());
+        PARSER_FACTORY.setNamespaceAware(true);
+    }
+
     private Function<URL, BeansXml> URL_TO_BEANS_XML_FUNCTION = BeansXmlParser.this::parse;
 
+    private SAXParser parser;
```
mkouba: I'm afraid we cannot share - `SAXParser` is not guaranteed to be thread-safe and `BeansXmlParser.parse(URL)` can be used concurrently.

mkouba:
I'm not quite sure we should go this way (sharing `SAXParserFactory` and stuff) but in any case, I find this improvement worth investigation! Thanks @ljnelson for the tip.

impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
ljnelson: OK.  I was going off the class documentation, which reads: "This class is NOT threadsafe, and should only be called in a single thread".  Indeed, if this documentation is incorrect and the class is called from multiple threads, then sharing the parser won't work.

ljnelson:
OK.  Is there a reason we shouldn't include the buffered reads and sharing the `SAXParserFactory`?  (I get that we shouldn't share the `SAXParser` if, contrary to the documentation, the class is used concurrently by multiple threads.) I'd like to learn what I'm missing, if anything.

impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
mkouba: BTW `BeansXmlParser` itself is (incorrectly) marked as NOT threadsafe.

mkouba:
@ljnelson I was thinking about replacing `SAXParser` with something more efficient.

ljnelson:
Oh!  I see.  I wonder: can you do that (probably)?  Isn't JAXP (and selection of the parser by system properties, etc.) required in, say, a Java EE environment?  Not sure.  Anyway, if _that's_ on the table…. 😄

mkouba:
AFAIK there is no requirement to use a specific XML parser, i.e. we can use anything.

ljnelson:
Right; was more thinking about whether or not Java EE users expect to be able to affect the parsing of `beans.xml` files (and other descriptors) by using the [JAXP system properties](https://docs.oracle.com/javase/tutorial/jaxp/properties/usingProps.html).  It's mentioned in the tutorial, at any rate, as [being part of the platform](https://docs.oracle.com/javaee/7/tutorial/overview008.htm#BNACU) which means that it's possible that users might expect that these system properties are legal to use to affect `beans.xml` processing.  I don't know.

mkouba:
@ljnelson I've tried to replace our SAX parser with StAX: https://github.com/weld/core/pull/1805

It would be great if you could test my PR and let us know if and how the parsing performance improved for your particular use case.

Thanks!

mkouba:
Closing in favor of https://github.com/weld/core/pull/1805.

PR #1801, opened 2018-03-09
WELD-2470 Fix InterceptedSubclassFactory, add reproducer.

manovotn:


impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedSubclassFactory.java
```diff
@@ -240,8 +240,15 @@ void doReturn(CodeAttribute b, ClassMethod method) {
     private boolean bridgeMethodsContainsMethod(Set<BridgeMethod> processedBridgeMethods, MethodSignature signature, Type returnType) {
         for (BridgeMethod bridgeMethod : processedBridgeMethods) {
             if (bridgeMethod.signature.equals(signature)) {
+                // method signature is equal (name and params) but return type can still differ
                 if (returnType != null) {
-                    return bridgeMethod.returnType.equals(returnType) ? true : false;
+                    if (bridgeMethod.returnType.equals(Object.class)) {
+                        // bridge method with matching signature has parameterized type as return type, this is a match
```
mkouba: I think the return type is not a parameterized type but a type variable.

manovotn: Yea, that's what I meant. Thanks. :)

manovotn:
The previous solution wasn't good enough for when we had to deal with hierarchy (using `T extends Vehicle` instead of just `T`).

So I tried going a bit further and assuming that if the method we search for is an abstract one, then signature match (with non-null return value) is good enough. Matching bridge method then overrides this new one and we needn't add it.

I am not really sure this won't mess up with some decorator setups; though tests are passing...

PR #1803, opened 2018-03-21
WELD-2457: Make it possible to drop unused beans metadata 

mkouba:
This is a POC for ideas around [WELD-2457](https://issues.jboss.org/browse/WELD-2457).

I've tested the changes with a small application (slightly modified [JSF numberguess](https://github.com/weld/core/tree/master/examples/jsf/numberguess) example with DeltaSpike core dependency) deployed to WildFly 12. When measuring the cold start of the app server the memory consumption was reduced by ~ 10% (~ 300 KB) at the cost of increasing the bootstrap time by ~ 5% (~ 60ms).

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/WeldUnusedMetadataExtension.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/WeldUnusedMetadataExtension.java
@@ -1,0 +1,111 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2018, Red Hat, Inc. and/or its affiliates, and individual
+ * contributors by the @authors tag. See the copyright.txt in the
+ * distribution for a full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.weld.bootstrap;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.HashSet;
+import java.util.Set;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.Initialized;
+import javax.enterprise.event.Observes;
+import javax.enterprise.inject.Instance;
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.Extension;
+import javax.enterprise.inject.spi.InjectionPoint;
+import javax.enterprise.inject.spi.ProcessInjectionPoint;
+
+import org.jboss.weld.manager.BeanManagerImpl;
+import org.jboss.weld.resolution.Resolvable;
+import org.jboss.weld.resolution.ResolvableBuilder;
+import org.jboss.weld.util.reflection.Reflections;
+
+/**
+ * This optional extension collects all injection points
+ * <ul>
+ * <li>of Java EE components,</li>
+ * <li>with {@link Instance} required type,</li>
+ * </ul>
+ * so that Weld is able to identify unused beans better.
+ *
+ * @author Martin Kouba
+ */
+public class WeldUnusedMetadataExtension implements Extension {
+
+    private Set<InjectionPoint> componentInjectionPoints;
+
+    private Set<InjectionPoint> instanceInjectionPoints;
+
+    WeldUnusedMetadataExtension() {
+        this.componentInjectionPoints = new HashSet<>();
+        this.instanceInjectionPoints = new HashSet<>();
+    }
+
+    void processInjectionPoints(@Observes ProcessInjectionPoint<?, ?> event) {
+        if (event.getInjectionPoint().getBean() == null) {
+            componentInjectionPoints.add(event.getInjectionPoint());
+        }
+        if (Instance.class.equals(Reflections.getRawType(event.getInjectionPoint().getType()))) {
+            instanceInjectionPoints.add(event.getInjectionPoint());
+        }
+    }
+
+    void clear(@Observes @Initialized(ApplicationScoped.class) Object obj) {
+        componentInjectionPoints.clear();
+        instanceInjectionPoints.clear();
+    }
+
+    public boolean isInjectedByEEComponent(Bean<?> bean, BeanManagerImpl beanManager) {
+        if (componentInjectionPoints.isEmpty()) {
+            return false;
+        }
+        for (InjectionPoint injectionPoint : componentInjectionPoints) {
+            if (beanManager.getBeanResolver().resolve(new ResolvableBuilder(injectionPoint, beanManager).create(), false).contains(bean)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public boolean isInstanceResolvedBean(Bean<?> bean, BeanManagerImpl beanManager) {
+        if (instanceInjectionPoints.isEmpty()) {
+            return false;
+        }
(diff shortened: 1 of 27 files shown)


mkouba:
retest this please

mkouba:
retest this please

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
```diff
@@ -1559,4 +1569,160 @@ public void validationFailed(Exception failure, Environment environment) {
...
+                }
+                BootstrapLogger.LOG.dropUnusedBeanMetadata(bean);
+                removable.add(bean);
+            }
+
+            // Second pass to find beans which themself are unused and declare only unused producers
+            if (!unusedProducers.isEmpty()) {
+                for (Bean<?> bean : enabledBeans) {
+                    if (bean instanceof ForwardingBean) {
+                        // Unwrap the forwarding implementation
```
manovotn: The unwrapping is done twice in this class, maybe create a private helper method?

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
```diff
@@ -1559,4 +1569,160 @@ public void validationFailed(Exception failure, Environment environment) {
...
+
+            if (!removable.isEmpty()) {
+                // First remove unused beans from BeanManager
+                enabledBeans.removeAll(removable);
+                sharedBeans.removeAll(removable);
+                // Then perform additional cleanup
+                beanResolver.clear();
+                for (Bean<?> bean : removable) {
+                    if (bean instanceof RIBean<?>) {
+                        // This will be skipped in WeldStartup.endInitialization()
```
manovotn: Why is this going to be skipped in `WeldStartup`?

impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
```diff
@@ -271,6 +274,53 @@
...
+     * <li>does not have a name</li>
+     * <li>does not declare an observer</li>
+     * <li>is not eligible for injection to any injection point</li>
+     * <li>does not declare a producer which is eligible for injection to any injection point</li>
+     * <li>is not eligible for injection into any {@link Instance} injection point</li>
+     * </ul>
+     *
+     * @see ConfigurationKey#UNUSED_BEANS_EXCLUDE_ANNOTATION
+     */
+    UNUSED_BEANS_EXCLUDE_TYPE("org.jboss.weld.bootstrap.unusedBeans.excludeType", ".*"),
```
manovotn: So, if I get this correctly, we are basically going to be eager (if the cleanup is enabled) and these two options are safety latches which can be utilized to avoid cleaning some beans. And the reason is that we cannot detect, for instance, dynamic resolution.

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
```diff
@@ -1559,4 +1569,160 @@ public void validationFailed(Exception failure, Environment environment) {
...
+                }
+                // This bean is very likely an unused producer
+                if (bean instanceof AbstractProducerBean) {
+                    unusedProducers.add(bean);
+                }
+                BootstrapLogger.LOG.dropUnusedBeanMetadata(bean);
+                removable.add(bean);
+            }
+
+            // Second pass to find beans which themself are unused and declare only unused producers
```
manovotn: `themself` -> `themselves`

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
```diff
@@ -1559,4 +1569,160 @@ public void validationFailed(Exception failure, Environment environment) {
...
+                if (observerMethod instanceof ObserverMethodImpl) {
+                    ObserverMethodImpl<?, ?> observerMethodImpl = (ObserverMethodImpl<?, ?>) observerMethod;
+                    beanToDeclaredObservers.put(observerMethodImpl.getDeclaringBean(), observerMethod);
+                }
+            }
+            Set<Bean<?>> removable = new HashSet<>();
+            Set<Bean<?>> unusedProducers = new HashSet<>();
+            WeldUnusedMetadataExtension metadataExtension = getUnusedMetadataExtension();
+
+            for (Bean<?> bean : enabledBeans) {
```
manovotn: We are going to be looping three times over all enabled beans.
That sounds a but pricy in huge deployments (and I think we have cleanup auto-enabled in WFLY, don't we?).
Not that I would know how to make it better yet, just thinking aloud :)

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
```diff
@@ -1559,4 +1569,160 @@ public void validationFailed(Exception failure, Environment environment) {
...
+
+            // Second pass to find beans which themself are unused and declare only unused producers
+            if (!unusedProducers.isEmpty()) {
+                for (Bean<?> bean : enabledBeans) {
+                    if (bean instanceof ForwardingBean) {
+                        // Unwrap the forwarding implementation
+                        ForwardingBean<?> forwarding = (ForwardingBean<?>) bean;
+                        bean = forwarding.delegate();
+                    }
+                    if (!beanToDeclaredProducers.containsKey(bean)) {
```
manovotn: Couldn't we just loop straight over `beanToDeclaredProducers.keySet()`?

impl/src/main/java/org/jboss/weld/bootstrap/WeldUnusedMetadataExtension.java
```diff
@@ -0,0 +1,111 @@
...
+    private Set<InjectionPoint> componentInjectionPoints;
+
+    private Set<InjectionPoint> instanceInjectionPoints;
+
+    WeldUnusedMetadataExtension() {
+        this.componentInjectionPoints = new HashSet<>();
+        this.instanceInjectionPoints = new HashSet<>();
+    }
+
+    void processInjectionPoints(@Observes ProcessInjectionPoint<?, ?> event) {
```
manovotn: Any reason for this being package protected? Same for `clear()` method.

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
mkouba: Good point and it should go to `org.jboss.weld.util.Beans`.

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
mkouba: Because in `WeldStartup#endInitialization()` we only call `cleanupAfterBoot()` for "present" beans. `BeanManagerImpl#cleanupAfterBoot()` is invoked after bootstrap services cleanup but before `RIBean#cleanupAfterBoot()`.

impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
mkouba: More or less. By default, all types are excluded, i.e. the feature is de facto disabled. If you set this property to empty string (or reg exp which does not match anything) then no type is excluded. You can also specify a reg exp e.g. to exclude classes from your app. `UNUSED_BEANS_EXCLUDE_ANNOTATION` is an additional "safety latch" to exclude types considering annotations only (e.g. by default any JAX-RS annotation -> any JAX-RS resource).

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
mkouba: +10 ;-)

impl/src/main/java/org/jboss/weld/bootstrap/WeldUnusedMetadataExtension.java
mkouba: There is no need to have them public, right? `WeldUnusedMetadataExtension` itself must be public but we don't want to "tempt" passersby ;-) 

I think it's always good to minimize visibility.

manovotn: Yes, I was thinking they could just be private then.

mkouba:
@manovotn Good idea...

PR #1805, opened 2018-03-22
WELD-2469: Improve performance of parsing beans.xml

mkouba:
- replaced SAX with StAX
- move validation logic in a separate class
- keep old classes for backward compatibility and delegate to new impls

First results show that the new impl is ~ 2x - 4x faster (depends on beans.xml complexity). I've also created  [some microbenchmarks](https://github.com/weld/weld-core-benchmarks/tree/master/src/main/java/org/jboss/weld/benchmark/core/xml).


impl/src/main/java/org/jboss/weld/xml/BeansXmlStreamParser.java
```diff
@@ -0,0 +1,398 @@
...
+    public BeansXmlStreamParser(URL beansXml, Function<String, String> interpolator) {
+        this.beansXml = beansXml;
+        this.interpolator = interpolator;
+    }
+
+    public BeansXml parse() {
+        if (beansXml == null) {
+            throw XmlLogger.LOG.loadError("unknown", null);
+        }
+        try (InputStream in = beansXml.openStream()) {
```
ljnelson: Maybe buffer this?

impl/src/main/java/org/jboss/weld/xml/BeansXmlValidator.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+        return xsds.toArray(EMPTY_SOURCE_ARRAY);
+    }
+
+    private static StreamSource loadXsd(String name, ClassLoader classLoader) {
+        InputStream in = classLoader.getResourceAsStream(name);
+        if (in == null) {
+            return null;
+        } else {
+            return new StreamSource(in);
```
ljnelson: Maybe buffer this?

impl/src/main/java/org/jboss/weld/xml/BeansXmlStreamParser.java
mkouba: I'd leave this to `XMLInputFactory` implementation...

impl/src/main/java/org/jboss/weld/xml/BeansXmlStreamParser.java
```diff
@@ -0,0 +1,398 @@
...
+            if (in.available() == 0) {
+                // The file is just acting as a marker file
+                return EMPTY_BEANS_XML;
+            }
+            XMLInputFactory factory = XMLInputFactory.newInstance();
```
ljnelson: Maybe share this?  Repeated creation of the factory can be very expensive.

impl/src/main/java/org/jboss/weld/xml/BeansXmlValidator.java
mkouba: XSDs are only loaded once (ideally). So I'd consider this a premature optimization ;-) The other issue is that we should probably close the stream...

impl/src/main/java/org/jboss/weld/xml/BeansXmlStreamParser.java
mkouba: Well, the API does not explicitly declare whether it's ok to share `XMLInputFactory`. But it seems that it should be safe. I can try to run some tests and we can see if it's worth the risk.

impl/src/main/java/org/jboss/weld/xml/BeansXmlStreamParser.java
ljnelson: Minor point, but I disagree; the "creator" of an input stream determines how or if it's buffered.  The "wrapper" (the `XMLInputFactory` in this case) normally just uses it as supplied.

impl/src/main/java/org/jboss/weld/xml/BeansXmlStreamParser.java
ljnelson: No, it's not safe, but you're talking about a one-line `synchronized` section, e.g.:

    XMLEventReader reader;
    synchronized (XML_INPUT_FACTORY) {
        reader = XML_INPUT_FACTORY.createXMLEventReader(in);
    }

I'm pretty sure that in a multi-`beans.xml` environment you're going to see a pretty good performance improvement from this; I'll move on to profiling later today.  Thanks for your work on all this!

ljnelson: Oh, plus this will let us use Woodstox, which will _really_ speed things up!

impl/src/main/java/org/jboss/weld/xml/BeansXmlValidator.java
ljnelson: Up to you; I found during profiling that eliminating the many `read()` function calls in general was the best bang for the buck I could get.  XSDs like Java EE's tend to be huge, so….  But yes, I haven't profiled this.

I _believe_ that "sources" (like `StreamSource`) are guaranteed by contract to be closed for you on full consumption (just another bit of weirdness from the JAXP area).

ljnelson: Also selfishly we don't have validation on anyway so for me personally it doesn't matter.

mkouba: Yes, at least the javadoc is lacking proper contract description :-(

mkouba: > Also selfishly...

Fair enough ;-). Validation errors result only in log warnings anyway...

impl/src/main/java/org/jboss/weld/xml/BeansXmlStreamParser.java
mkouba: Except the contract is not defined. `beans.xml` are usually small so I don't think it's worth the effort.

impl/src/main/java/org/jboss/weld/xml/BeansXmlValidator.java
```diff
@@ -0,0 +1,158 @@
...
+            throw XmlLogger.LOG.loadError(beansXml, e);
+        }
+
+        if (schema == null) {
+            return;
+        }
+        Validator validator = schema.newValidator();
+        validator.setErrorHandler(errorHandler);
+        try {
+            validator.validate(new StreamSource(beansXml.toExternalForm()));
```
ljnelson: If you wanted to be able to close the stream here, maybe do:

    try (final InputStream in = new BufferedInputStream(beansXml.openStream()) {
        validator.validate(in);
    } catch (SAXException | IOException e) {
        // No-op - validation is optional
    }

ljnelson: …unless something magic™ happens 😄 by converting a URL to string form when you supply it to the `StreamSource` constructor?  If you already have the URL, I would think just open its stream again here.

impl/src/main/java/org/jboss/weld/xml/BeansXmlValidator.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    public void validate(URL beansXml, ErrorHandler errorHandler) {
+        if (beansXml == null) {
+            throw XmlLogger.LOG.loadError("unknown", null);
+        }
+        if (errorHandler == null) {
+            errorHandler = this;
+        }
+        Schema schema = cdi20Schema;
```
ljnelson: (Just a quick check: if there's no schema specified in the `beans.xml`, then CDI 2.0 will be used.  I assume that's right?)

mkouba: Yep.

impl/src/main/java/org/jboss/weld/xml/BeansXmlValidator.java
ljnelson: Ah; there's this: https://bugs.openjdk.java.net/browse/JDK-8139262

So probably want to pass in `InputStream`s explicitly to the `StreamSource` constructor wherever possible and then close them explicitly afterwards.  Bleagh.

impl/src/main/java/org/jboss/weld/xml/BeansXmlValidator.java
mkouba: Hm, `javax.xml.validation.Validator` only accepts `javax.xml.transform.Source` so something like this looks ok:
````
try (final InputStream in = new BufferedInputStream(beansXml.openStream()) {
    validator.validate(new StreamSource(in));
} catch (SAXException | IOException e) {
    // No-op - validation is optional
}
````

File: impl/src/main/java/org/jboss/weld/xml/BeansXmlValidator.java (hunk 2 above)
Comment (mkouba) [reply impl/src/main/java/org/jboss/weld/xml/BeansXmlValidator.java]: Hm, thanks for link!

File: impl/src/main/java/org/jboss/weld/xml/BeansXmlStreamParser.java (hunk 3 above)
Comment (mkouba) [reply impl/src/main/java/org/jboss/weld/xml/BeansXmlStreamParser.java]: I think that Woodstox is an impl of StAX. It's up to the environment /user which impl is used.

File: impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java [hunk 6]
```diff
@@ -16,144 +16,66 @@
...
-        List<Metadata<String>> interceptors = new ArrayList<Metadata<String>>();
-        List<Metadata<Filter>> includes = new ArrayList<Metadata<Filter>>();
-        List<Metadata<Filter>> excludes = new ArrayList<Metadata<Filter>>();
+    protected BeansXmlHandler getHandler(final URL beansXml) {
+        return null;
+    }
+
+    // Merging utils
+
+    public static <T> BeansXml merge(Iterable<? extends T> items, Function<T, BeansXml> function, boolean removeDuplicates) {
```
manovotn: Why do we want to preserve these util methods in a "legacy" class (only kept because of integrators)?
Since you changed the signature, I suppose integrators aren't using it anyway, so we could move it to some util class perhaps?

mkouba: Good point. These should be placed in a separate util class but since they are also used in `BeansXmlParser.parse(Iterable<URL>)` I decided to keep them here.

manovotn: Well, since there is no easy way we can let go of old parser, I suppose there isn't much harm in it.

ljnelson:
Hi, @mkouba; no, but SAX to StAX is good so I'd say it's all good.

manovotn:
No objections from me either.

PR #1806, opened 2018-03-22
WELD-2474 Use Weld's internal literals.

manovotn:
I did grep our codebase for these and I think this is the sole problematic case.

There were other occurrences of course, but:
* those in tests can be ignored (SE or integration)
* those is SE impl too as they are not relevant for WFLY
* those in core but related to `Configurators` API can only be used with CDI 2.0, so we are safe there as well


mkouba:
> those in tests can be ignored (SE or integration)

 :+1: Assuming we don't run our test suite against WildFly 12 EE 7 (default) mode.

> those is SE impl too as they are not relevant for WFLY

:+1:

> those in core but related to Configurators API can only be used with CDI 2.0, so we are safe there as well

:+1:

PR #1807, opened 2018-03-23
WELD-2477: Optimize string representation of AnnotatedTypeIdentifier

mkouba:
- use SHA-1 hash for generated ids for synthetic types
- shorten bdaId - bean deployment archive id very often starts with context id
- shorten suffix - AnnodatedType added by an extension often has suffix starting with class FQCN
- shorten modified - use 0 and 1

### DIFF
--- a/impl/src/main/java/org/jboss/weld/annotated/slim/AnnotatedTypeIdentifier.java
+++ b/impl/src/main/java/org/jboss/weld/annotated/slim/AnnotatedTypeIdentifier.java
@@ -141,13 +141,15 @@ public class AnnotatedTypeIdentifier implements Identifier {
         StringBuilder builder = new StringBuilder();
         builder.append(contextId);
         builder.append(ID_SEPARATOR);
-        builder.append(bdaId);
+        // Bean deployment archive id very often starts with context id
+        builder.append(bdaId != null ? (bdaId.startsWith(contextId) ? bdaId.substring(contextId.length()) : bdaId) : bdaId);
         builder.append(ID_SEPARATOR);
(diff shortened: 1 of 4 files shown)


mkouba:
retest this please

impl/src/main/java/org/jboss/weld/bootstrap/BeanDeployer.java
```diff
@@ -122,7 +122,8 @@ private void processPriority(AnnotatedType<?> type) {
 
     public <T> BeanDeployer addSyntheticClass(AnnotatedType<T> source, Extension extension, String suffix) {
         if (suffix == null) {
-            suffix = AnnotatedTypes.createTypeId(source);
+            // If there is a synthetic AnnotatedType added for the given class then calculate the suffix, otherwise use extension FQCN
+            suffix = getEnvironment().hasSyntheticAnnotatedType(source.getJavaClass()) ? AnnotatedTypes.createTypeId(source) : extension.getClass().getName();
```
mkouba: This is actually wrong as it makes the id non-deterministic (if multiple ATs for a given class exist).

manovotn:
This PR is bumping into CDI TCK problem, that failing test is faulty because it can result in an unsatisfied dependency if certain sequence of method invocations in the extension occurs. E.g. it can happen that the IP only has `@Default` and the actual bean had its `@Default` removed and has `@Any` qualifier.

Still, we should look into why this PR triggered this...

manovotn:
@mkouba see also https://issues.jboss.org/browse/CDITCK-614 and the attached PR

PR #1808, opened 2018-03-26
WELD-2479 Allow to modify lists of alternatives/interceptors/decorato…

manovotn:
…rs via List.remove(Object). Add test case for this.


manovotn:
I added one container test with `remove(Object)` but most tests are added to single unit-test (`EnablementListViewTest`) as it seemed like a good fit.

impl/src/main/java/org/jboss/weld/bootstrap/enablement/EnablementListView.java
```diff
@@ -105,23 +106,26 @@ public boolean removeAll(Collection<?> c) {
...
     }
 
     @Override
     public boolean remove(Object o) {
         if (getExtension() != null) {
             BootstrapLogger.LOG.typeModifiedInAfterTypeDiscovery(getExtension(), o, REMOVE_OPERATION, getViewType());
         }
-        return getDelegate().remove(o);
+        // implementation of Item does equals/hashCode only based on Java class
+        return getDelegate().remove(createSource((Class<?>) o, 0));
```
mkouba: Why don't you use `objectToItemIfNeeded()` here?

manovotn: Eeehm, because I implemented this method before introducing the helper one.
Will fix this, thanks!

PR #1809, opened 2018-03-27
WELD-2478: Always use constructed guard for some proxy factories

mkouba:


mkouba:
Well, even a separate test would have to contain the config logic. So it would be more or less the same.

manovotn:
I know but once it comes to regression it's convenient to see a test failing which does one thing at a time. But as I said, it's tiny detail in this case, that's why it's approved anyway.

PR #1815, opened 2018-04-10
WELD-2460, WELD-2490, WELD-2487 - JDK 9+ fixes and tweaks

manovotn:
EDITED: This PR resolves WELD-2460 and most related subtasks.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
@@ -74,7 +74,6 @@ import org.jboss.weld.util.bytecode.ConstructorUtils;
 import org.jboss.weld.util.bytecode.DeferredBytecode;
 import org.jboss.weld.util.bytecode.MethodInformation;
 import org.jboss.weld.util.bytecode.RuntimeMethodInformation;
-import org.jboss.weld.util.collections.Arrays2;
 import org.jboss.weld.util.collections.ImmutableSet;
 import org.jboss.weld.util.collections.Sets;
 import org.jboss.weld.util.reflection.Reflections;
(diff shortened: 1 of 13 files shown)


manovotn:
Note - PR executors currently run against patched WFLY 12. That means WFLY itself has illegal access and that will crash the build on JDK 10. I'll look around for some nightly WFLY 13 we could use until release.

manovotn:
retest this please

manovotn:
retest this please

manovotn:
Turns out we cannot yet `deny` illegal access on WFLY 13.
There are quite a few errors coming from XNio :-/

manovotn:
Note that illegal access is currently not denied.
Though looking at Jenkins log I see no illegal access coming from Weld itself.

manovotn:
Trying to run with denied illegal access (in container) yields an error coming from [infinispan](http://pastebin.test.redhat.com/576691).
Hence the PR won't include this option unless we want to force it only for tests outside of container (where it seems to be working just fine). I've also added a commit which makes sure that JDK 9+ opts will propagate to Arq. config for WFLY.

pom.xml
```diff
@@ -53,7 +59,8 @@
         <bundle.plugin.version>2.5.3</bundle.plugin.version>
         <cdi.tck-1-2.version>1.2.10.Final</cdi.tck-1-2.version>
         <cdi.tck-2-0.version>2.0.4.Final</cdi.tck-2-0.version>
-        <classfilewriter.version>1.1.2.Final</classfilewriter.version>
+        <!--1.2.2.Final is absolute minimum to eliminate IllegalAccess warning in JDK9+-->
+        <classfilewriter.version>1.2.2.Final</classfilewriter.version>
```
mkouba: If 1.2.2.Final is minimum, we can probably remove `org.jboss.weld.bean.proxy.ProxyFactory.CONFIGURABLE_ACCESS_FLAGS` and `org.jboss.weld.bean.proxy.ProxyFactory.CONFIGURABLE_CLASSLOADER_CONSTRUCTOR` too?

Anyway, this would mean we cannot create a patch for older WildFly releases, right?

manovotn: Hmm, I wanted to keep a note on why we needed that version specifically, but it's probably pointless.
As for the flags - probably yes, I can try and remove those, hope I won't break anything :)

As for patches - TBH I have no idea how this correlates with older WFLYs. I mean even WFLY must have had newer version for ages. We actually only compiled against older one.

mkouba: > We actually only compiled against older one.

Not really. We compiled against an older version but tried to leverage functionality from the newer ones.

manovotn: That's what I mean. WFLY had newer version for a long time. So I cannot see why we couldn't create a patch for, say, WFLY 12.

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -501,13 +474,8 @@ protected void addAdditionalInterfaces(Set<Class<?>> interfaces) {
...
     private ClassFile newClassFile(String name, int accessFlags, String superclass, String... interfaces) {
         try {
-            if (CONFIGURABLE_CLASSLOADER_CONSTRUCTOR != null) {
-                return CONFIGURABLE_CLASSLOADER_CONSTRUCTOR.newInstance(name, accessFlags, superclass, classLoader, interfaces);
-            } else if (CONFIGURABLE_ACCESS_FLAGS) {
-                return new ClassFile(name, accessFlags, superclass, interfaces);
-            } else {
-                return new ClassFile(name, superclass, interfaces);
-            }
+            return SecurityActions.getConstructor(ClassFile.class, String.class, int.class, String.class, ClassLoader.class, Arrays2.EMPTY_STRING_ARRAY.getClass())
```
mkouba: Why not use the constructor right away, i.e. without reflection?

manovotn: Hmm yea, I was just brainlessly removing dead code. Will do tomorrow.

mkouba:
It seems there are also checkstyle errors?

manovotn:
retest this please

manovotn:
retest this please

PR #1817, opened 2018-04-19
WELD-2484 Configurators should override toString()

manovotn:
Not sure this is 'sensible enough'. It can blow up (and call toString()) from many scenarios so amount of useful information may vary.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/InjectionPointConfiguratorImpl.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/InjectionPointConfiguratorImpl.java
@@ -210,6 +210,10 @@ public class InjectionPointConfiguratorImpl implements InjectionPointConfigurato
            return isTransient;
        }
 
+       @Override
+       public String toString() {
+           return "InjectionPoint with type=" + requiredType + ", qualifiers=" + qualifiers +
+               ", delegate=" + isDelegate + ", transient=" + isTransient + ".";
+       }
    }
-
 }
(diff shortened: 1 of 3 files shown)


impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/InjectionPointConfiguratorImpl.java
```diff
@@ -210,6 +210,10 @@ public boolean isTransient() {
            return isTransient;
        }
 
+       @Override
+       public String toString() {
+           return "InjectionPoint of type " + requiredType + " with qualifiers " + qualifiers +
+               " which is " + (isDelegate ? "" : "not ") + "delegate and is" + (isTransient ? "" : "not ") + "transient.";
```
mkouba: `delegate=true` style would be just fine.

mkouba:
I wonder if we should add sensible `toString()` to all products of all configurators. Some of them already have it...

manovotn:
Hmm, I think only `BeanConfigurator` had this altered so far.

impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/ObserverMethodConfiguratorImpl.java
```diff
@@ -330,6 +330,14 @@ public boolean isEventMetadataRequired() {
            return true;
        }
 
-   }
+       @Override
+       public String toString() {
+           return "Configurator observer method [Bean class = " + getBeanClass().toString()
```
mkouba: I'm not sure why but we allow to set `null` bean class. If you remove `toString()` here, then this method should not throw NPE and the resulting string should be the same.

manovotn: Nice catch!

PR #1818, opened 2018-04-19
Add a property that causes retreving a session or conversation scoped…

benjamin-confino:
… bean from the session store to mark that attribute as modified

### DIFF
--- a/impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
+++ b/impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
@@ -322,6 +322,13 @@ public enum ConfigurationKey {
      */
     UNUSED_BEANS_EXCLUDE_ANNOTATION("org.jboss.weld.bootstrap.unusedBeans.excludeAnnotation", "javax\\.ws\\.rs.*"),
 
+    /**
+     * If set to true then when a contextual reference for a @SessionScoped or @ConversationScoped bean is obtained from a context backed by an HTTP session
+     * the instance is set again using HttpSession.setAttribute(). This allows to trigger session replication in some application servers.
+     */
+    @Description("If set to true then when a contextual reference for a @SessionScoped or @ConversationScoped bean is obtained from a context backed by an HTTP session the instance is set again using HttpSession.setAttribute(). This allows to trigger session replication in some application servers.")
+    RESET_HTTP_SESSION_ATTR_ON_BEAN_ACCESS("org.jboss.weld.context.resetHttpSessionAttributeOnBeanAccess", false),
+
     ;
 
     /**
--- a/modules/web/src/main/java/org/jboss/weld/module/web/context/beanstore/http/AbstractSessionBeanStore.java
+++ b/modules/web/src/main/java/org/jboss/weld/module/web/context/beanstore/http/AbstractSessionBeanStore.java
@@ -7,6 +7,9 @@ import java.util.Iterator;
 
 import javax.servlet.http.HttpSession;
 
+import org.jboss.weld.bootstrap.api.ServiceRegistry;
+import org.jboss.weld.config.ConfigurationKey;
+import org.jboss.weld.config.WeldConfiguration;
 import org.jboss.weld.context.api.ContextualInstance;
 import org.jboss.weld.contexts.beanstore.AttributeBeanStore;
 import org.jboss.weld.contexts.beanstore.LockStore;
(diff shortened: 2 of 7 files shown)


benjamin-confino:
Note that I saw this tck failure. 

  AnnotatedTypeConfiguratorTest>Arquillian.run:138->configuratorInitializedWithOriginalAT:168 expected [true] but found [false]

However I do not think it is related, I saw the same failure on master.

mkouba:
test this please

modules/web/src/main/java/org/jboss/weld/module/web/context/beanstore/http/AbstractSessionBeanStore.java
```diff
@@ -34,13 +37,16 @@
 
     protected abstract HttpSession getSession(boolean create);
 
+    private final ServiceRegistry serviceRegistry;
```
mkouba: `WeldConfiguration` is immutable so I'd rather store the boolean value of the config property.

impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
```diff
@@ -322,6 +322,12 @@
      */
     UNUSED_BEANS_EXCLUDE_ANNOTATION("org.jboss.weld.bootstrap.unusedBeans.excludeAnnotation", "javax\\.ws\\.rs.*"),
 
+    /**
+     * If true accessing a SessionScope bean will result it in being refreshed in the HTTP Session attributes.
+     */
+    @Description("Upon access refresh SessionScoped and ConversationScoped beans into the HTTP Session.")
+    ALWAYS_REFRESH_SESSION_CONVERSATION_SCOPE("org.jboss.weld.context.alwaysRefreshSessionAndConversationScope", false),
```
mkouba: Hm, the name is always the tricky part. What about the key `ENFORCE_HTTP_SESSION_REPLICATION` and property `org.jboss.weld.context.enforceHttpSessionReplication`? I don't think we should mention SCOPE, CONVERSATION (conversation context is backed by an HTTP session anyway) and REFRESH sounds like the whole context is refreshed which is not necessarily true and also feels costly ;-).
 
Also I think the description should be more clear. Maybe something like:
> If set to `true` then when a contextual reference for a `@SessionScoped` or `@ConversationScoped` bean is obtained from a context backed by an HTTP session the instance is set again using `HttpSession.setAttribute()`. This allows to trigger session replication in some application servers.

mkouba:
The commit message should start with `WELD-1130: ` prefix

impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
manovotn: It's not `ENFORCE` really, as it's still best effort and not entirely guaranteed albeit common enough to look as if it is enforced most of the time.
Also, the replication is something we do not handle directly with this, we just re-store it into session attributes often enough.
I was thinking about `EAGER_SESSION_ATTR_REFRESH`? Or perhaps `REPEATED_SESSION_ATTR_STORE`? WDYT?

manovotn: Otherwise +1 for the suggested change of description

modules/web/src/main/java/org/jboss/weld/module/web/context/beanstore/http/AbstractSessionBeanStore.java
manovotn: +1

impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
mkouba: > I was thinking about EAGER_SESSION_ATTR_REFRESH? Or perhaps REPEATED_SESSION_ATTR_STORE? WDYT?

-1 but I understand your concerns about `ENFORCE`. I think we shoud keep `HTTP_SESSION` and `REPLICATION` parts because user would understand these. And the third part should be something like _"Weld tries to urge the container to trigger..."_. 

On the other hand, this property will be probably used by advanced users so we could probably reference the replication semantics. AFAIK in JBoss AS/EAP 6 this was called `SET replication trigger` (it's not supported in WildFly/EAP 7).

mkouba: `REPLICATE_HTTP_SESSION_ON_ACCESS ` sounds good but is not very accurate. The problem is that replication usually happens on bean access when some kind of "GET trigger" is used (e.g. the default in WildFly). What about `RESET_HTTP_SESSION_ATTR_ON_BEAN_ACCESS?`

benjamin-confino: `RESET_HTTP_SESSION_ATTR_ON_BEAN_ACCESS?` Sounds good to me.

mkouba: Ok, then pls incorporate all required changes. Thanks ;-)

benjamin-confino: But of course ;)

benjamin-confino:
All three changes made on both branches.

impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
```diff
@@ -322,6 +322,12 @@
      */
     UNUSED_BEANS_EXCLUDE_ANNOTATION("org.jboss.weld.bootstrap.unusedBeans.excludeAnnotation", "javax\\.ws\\.rs.*"),
 
+    /**
+     * If true accessing a SessionScope bean will result it in being refreshed in the HTTP Session attributes.
```
mkouba: Pls modify also the javadoc (@Description is ok).

benjamin-confino:
Thanks for the feedback and the merge :)

PR #1819, opened 2018-04-19
Add a property that causes retreving a session or conversation scoped…

benjamin-confino:
… bean from the session store to mark that attribute as modified


mkouba:
test this please

mkouba:
add to whitelist

PR #1821, opened 2018-04-20
WELD-2480: Include JS dependencies licenses in weld-probe-core.jar

mkouba:


manovotn:
Shouldn't we maybe review and update all those dependencies while at it?

mkouba:
AFAIR we cannot easily upgrade Ember.js and its deps because versions 2.0+ require some additional migration work as it is not backward compatible. So I think we could create a new issue but since probe is a dev phase tool it should be probably low priority.

manovotn:
If you want to, create that separate JIRA for dep update.

manovotn:
retest this please

PR #1822, opened 2018-04-23
WELD-2482 Globally selected alternatives with same priority should re…

manovotn:
…sult in ambiguous dependency.


impl/src/main/java/org/jboss/weld/bootstrap/enablement/GlobalEnablementBuilder.java
```diff
@@ -173,7 +173,7 @@ public boolean isDirty() {
             Map<Class<?>, Integer> map = new HashMap<Class<?>, Integer>();
             for (ListIterator<Item> iterator = alternatives.listIterator(); iterator.hasNext(); ) {
                 Item item = iterator.next();
-                map.put(item.getJavaClass(), iterator.previousIndex());
+                map.put(item.getJavaClass(), item.getPriority());
```
mkouba: I think there is no need to use `ListIterator` here anymore...

PR #1825, opened 2018-04-25
Few fixes and exclusions for Weld examples on JDK 9.

manovotn:
**Not to be merged**, this is a tracking PR for changes needed in Weld examples for JDK 9+

Some of the dep updates should go to parent.
Groovy + webstart are disabled altogether as I am not sure how to fix and if it's worth fixing in the first place.

### DIFF
--- a/environments/se/core/pom.xml
+++ b/environments/se/core/pom.xml
@@ -14,12 +14,6 @@
 
     <description>Weld support for Java SE</description>
 
-    <properties>
-        <groovy.version>2.4.7</groovy.version>
-        <groovy.eclipse.compiler.version>2.9.2-01</groovy.eclipse.compiler.version>
-        <groovy.eclipse.batch.version>2.4.3-01</groovy.eclipse.batch.version>
-    </properties>
-
     <url>http://weld.cdi-spec.org</url>
     <licenses>
         <license>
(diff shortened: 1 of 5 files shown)


manovotn:
Alright, so what do we do with these examples?
As far as I am concerned, `groovy-numberguess` and `webstart` are by now obsolete and can be removed altogether.

@mkouba what's your take on this?

mkouba:
I would not remove them completely but instead create a profile for JDK8.

manovotn:
Hmm I am against excluding them for certain versions of Java. That way it will simply be forgotten for any future version. I'd rather fix or delete them.

`webstart` example uses Java Web Start. According to Oracle, this technology is considered deprecated since JDK 9 - see also [this article](http://www.oracle.com/technetwork/java/javase/9-deprecated-features-3745636.html). There is also a related JDK mail list discussion [here](http://mail.openjdk.java.net/pipermail/discuss/2017-November/004586.html). I really don't think it's worth keeping a deprecated example be it on JDK 10 or 8. 

As for Groovy parts in Weld, I am unsure. We haven't seen any question or incentive around groovy and weld for quite some time. The sample probably could be reworked to run on new Java (I've glanced some new [plugin](http://groovy.github.io/GMavenPlus/) used for that). While at it we should also look at `env/se/core/pom.xml` where I recall we excluded some groovy compilation on JDK 9+ because back there groovy had zero support for JDK 9.

manovotn:
Seems like Groovy compilation and (test)source management can now be done via `gmavenplus-plugin` (added a commit with that). The Groovy tests we had in SE core now execute on JDK 10.
However, the SE Groovy example keeps resisting on JDK 10 (runs on 8) with loads of `WELD-000119`. I have no idea why ATM.

manovotn:
Closing this PR, will open new one shortly.

PR #1827, opened 2018-05-03
WELD-2250 Fix a scenario with qualifier class present in multiple arc…

manovotn:
…hives within EAR. Add automated test.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/resolution/QualifierInstance.java
+++ b/impl/src/main/java/org/jboss/weld/resolution/QualifierInstance.java
@@ -18,6 +18,7 @@ package org.jboss.weld.resolution;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.security.AccessController;
 import java.util.Collections;
 import java.util.Map;
@@ -28,12 +29,14 @@ import javax.enterprise.inject.Any;
 import javax.enterprise.inject.Default;
 import javax.enterprise.inject.spi.AnnotatedMethod;
 import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.util.Nonbinding;
 import javax.inject.Named;
 
 import org.jboss.weld.bean.RIBean;
 import org.jboss.weld.logging.ResolutionLogger;
 import org.jboss.weld.metadata.cache.MetaAnnotationStore;
 import org.jboss.weld.metadata.cache.QualifierModel;
+import org.jboss.weld.security.GetDeclaredMethodsAction;
 import org.jboss.weld.security.SetAccessibleAction;
 import org.jboss.weld.util.collections.ImmutableMap;
 import org.jboss.weld.util.collections.ImmutableSet;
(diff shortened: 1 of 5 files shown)


impl/src/main/java/org/jboss/weld/resolution/QualifierInstance.java
```diff
@@ -134,7 +136,22 @@ public QualifierInstance(Class<? extends Annotation> annotationClass, Map<String
                     }
                     builder.put(method.getJavaMember().getName(), method.getJavaMember().invoke(instance));
                 } catch (IllegalAccessException | InvocationTargetException | IllegalArgumentException e) {
-                    throw ResolutionLogger.LOG.cannotCreateQualifierInstanceValues(instance, Formats.formatAsStackTraceElement(method.getJavaMember()), e);
+                    // it may happen that we are in EAR and have stored the annotation's method from different WAR class loader
+                    // an invocation will then lead to IAE, we can re-try with reflection
+                    Method[] methods;
+                    builder = ImmutableMap.builder();
+                    if (System.getSecurityManager() != null) {
+                        methods = AccessController.doPrivileged((PrivilegedAction<Method[]>)() -> annotationClass.getDeclaredMethods());
```
mkouba: You could use `GetDeclaredMethodsAction`.

manovotn: Hmm, I was looking for one such. But failing to find it I just did this.

impl/src/main/java/org/jboss/weld/resolution/QualifierInstance.java
```diff
@@ -134,7 +136,22 @@ public QualifierInstance(Class<? extends Annotation> annotationClass, Map<String
...
+                    } else {
+                        methods = annotationClass.getDeclaredMethods();
+                    }
+                    for (Method m : methods) {
```
mkouba: We have to filter out members annotated with `javax.enterprise.util.Nonbinding`.

impl/src/main/java/org/jboss/weld/resolution/QualifierInstance.java
```diff
@@ -134,7 +137,24 @@ public QualifierInstance(Class<? extends Annotation> annotationClass, Map<String
                     }
                     builder.put(method.getJavaMember().getName(), method.getJavaMember().invoke(instance));
                 } catch (IllegalAccessException | InvocationTargetException | IllegalArgumentException e) {
-                    throw ResolutionLogger.LOG.cannotCreateQualifierInstanceValues(instance, Formats.formatAsStackTraceElement(method.getJavaMember()), e);
+                    // it may happen that we are in EAR and have stored the annotation's method from different WAR class loader
```
mkouba: I think we should only re-try if `IllegalArgumentException` is thrown following the `java.lang.reflect.Method.invoke(Object, Object...)` javadoc.

tests-arquillian/src/test/java/org/jboss/weld/tests/qualifier/ear/LibraryWithQualifierInTwoWarsTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+        // setup EAR
+        EnterpriseArchive ear = ShrinkWrap.create(EnterpriseArchive.class, Utils.getDeploymentNameAsHash(LibraryWithQualifierInTwoWarsTest.class, Utils.ARCHIVE_TYPE.EAR))
+            .addAsModules(war1, war2);
+        return ear;
+    }
+
+    @Test
+    public void test() {
+        // verification is merely the ability to deploy this scenario
```
mkouba: Maybe we should doublecheck the relevant qualifier instance is created... but you're hopefully writing a failing test first so it should be ok ;-)

manovotn: Yes, the test was written prior to the fix, so it was failing.

PR #1831, opened 2018-05-16
WELD-2494 Set TCCL to null when runnning with ForkJoinPool ExecutorSe…

manovotn:
…rvices.

Test was failing without fix, works with it.

### DIFF
--- a/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/service/executorService/ContextClassLoaderNullifiedTest.java
+++ b/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/service/executorService/ContextClassLoaderNullifiedTest.java
@@ -1,0 +1,82 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2018, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.weld.environment.se.test.service.executorService;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.jboss.arquillian.container.se.api.ClassPath;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.BeanArchive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.weld.bean.builtin.BeanManagerProxy;
+import org.jboss.weld.bootstrap.api.ServiceRegistry;
+import org.jboss.weld.config.ConfigurationKey;
+import org.jboss.weld.environment.se.Weld;
+import org.jboss.weld.environment.se.WeldContainer;
+import org.jboss.weld.executor.ExecutorServicesFactory;
+import org.jboss.weld.manager.api.ExecutorServices;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * Tests that TCCL is set to null when running in SE with enabled discovery (in which case we use ForkJoinPool and hence CommonForkJoinPoolExecutorServices)
+ *
+ * @see WELD-2494
+ * @author <a href="mailto:manovotn@redhat.com">Matej Novotny</a>
+ */
+@RunWith(Arquillian.class)
+public class ContextClassLoaderNullifiedTest {
+
+    @Deployment
+    public static Archive<?> createTestArchive() {
+        return ClassPath.builder().add(ShrinkWrap.create(BeanArchive.class)
+            .addClasses(ContextClassLoaderNullifiedTest.class, DummyBean.class))
+            .build();
+    }
+
+    @Test
+    public void testTcclNullifiedInSeEnvironment() {
+        Weld weld = new Weld()
+            // in case default changes, make sure this tests uses ForkJoinPool
+            .addProperty(ConfigurationKey.EXECUTOR_THREAD_POOL_TYPE.toString(), ExecutorServicesFactory.ThreadPoolType.COMMON.toString());
+        try (WeldContainer container = weld.initialize()) {
+            AtomicBoolean isTCCLNull = new AtomicBoolean(false);
+            
+            // retrieve service and use it to invoke Callable
+            ServiceRegistry registry = BeanManagerProxy.unwrap(container.getBeanManager()).getServices();
+            ExecutorServices es = registry.get(ExecutorServices.class);
+            assertNotNull(es);
+            Collection<Callable<Void>> callables = new ArrayList<>();
+            callables.add(() -> {
+                isTCCLNull.set(Thread.currentThread().getContextClassLoader() == null);
+                return null;
+            });
+            // callable should check if the TCCL was truly null
+            es.invokeAllAndCheckForExceptions(callables);
+            // now we assert it
+            assertTrue(isTCCLNull.get());
+        }
+    }
+}
--- a/impl/src/main/java/org/jboss/weld/Container.java
+++ b/impl/src/main/java/org/jboss/weld/Container.java
@@ -19,9 +19,11 @@ package org.jboss.weld;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.jboss.weld.annotated.slim.AnnotatedTypeIdentifier;
 import org.jboss.weld.bootstrap.BeanDeploymentArchiveMapping;
+import org.jboss.weld.bootstrap.api.Environment;
 import org.jboss.weld.bootstrap.api.ServiceRegistry;
 import org.jboss.weld.bootstrap.api.Singleton;
 import org.jboss.weld.bootstrap.api.SingletonProvider;
@@ -39,6 +41,8 @@ public class Container {
     public static final String CONTEXT_ID_KEY = "WELD_CONTEXT_ID_KEY";
 
     private static Singleton<Container> instance;
+
+    private static final AtomicReference<Environment> ENV_REFERENCE = new AtomicReference<>(null);
 
     static {
         instance = SingletonProvider.instance().create(Container.class);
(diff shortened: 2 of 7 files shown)


impl/src/main/java/org/jboss/weld/executor/AbstractExecutorServices.java
```diff
@@ -55,7 +55,8 @@ public ScheduledExecutorService getTimerExecutor() {
     @Override
     public <T> List<Future<T>> invokeAllAndCheckForExceptions(Collection<? extends Callable<T>> tasks) {
         try {
-            return checkForExceptions(getTaskExecutor().invokeAll(tasks));
+            // the cast is needed to compile this expression in JDK 8, works without it on JDK 10+
+            return checkForExceptions(getTaskExecutor().invokeAll((Collection<? extends Callable<T>>)wrap(tasks)));
```
manovotn: This cast should be harmless as it is the exact same type as is returned by `wrap()` function.
However, I found no one-liner solution to avoid this and still be able to compile on JDK 8.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/service/executorService/ContextClassLoaderNullifiedTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void testSimpleAdditionalService() {
+        try (WeldContainer container = new Weld().initialize()) {
+            ServiceRegistry registry = BeanManagerProxy.unwrap(container.getBeanManager()).getServices();
+            ExecutorServices es = registry.get(ExecutorServices.class);
+            assertNotNull(es);
+            Collection<Callable<Void>> callables = new ArrayList<>();
+
+            callables.add(() -> {
+                isTCCLNull.set(Thread.currentThread().getContextClassLoader() == null);
```
mkouba: Hm, there is no need to use a static field, or? Method scope should work just fine.

impl/src/main/java/org/jboss/weld/executor/AbstractExecutorServices.java
mkouba: Np, we already have similar casts in the codebase here and there...

impl/src/main/java/org/jboss/weld/executor/CommonForkJoinPoolExecutorServices.java
```diff
@@ -42,4 +46,21 @@ public void cleanup() {
     protected int getThreadPoolSize() {
         return ForkJoinPool.getCommonPoolParallelism();
     }
+
+    @Override
+    public <T> Collection<? extends Callable<T>> wrap(Collection<? extends Callable<T>> tasks) {
+        List<Callable<T>> wrapped = new ArrayList<>(tasks.size());
```
mkouba: It wouldn't hurst to add a brief comment about why we do this...

impl/src/main/java/org/jboss/weld/executor/AbstractExecutorServices.java
```diff
@@ -132,4 +133,15 @@ protected void shutdown() {
             Thread.currentThread().interrupt();
         }
     }
+
+    /**
+     * This method is invoked with the body of invokeAllAndCheckForExceptions(Collection<? extends Callable<T>> tasks).
+     *
+     * It allows to wrap the tasks with some additional logic. For instance, Weld's CommonForkJoinPoolExecutorServices
```
mkouba: It's always better to link classes by means of `{@link}`... and make refactoring possible ;-)

impl/src/main/java/org/jboss/weld/executor/AbstractExecutorServices.java
```diff
@@ -132,4 +133,15 @@ protected void shutdown() {
             Thread.currentThread().interrupt();
         }
     }
+
+    /**
+     * This method is invoked with the body of invokeAllAndCheckForExceptions(Collection<? extends Callable<T>> tasks).
```
mkouba: The same as https://github.com/weld/core/pull/1831/files#r188580194

impl/src/main/java/org/jboss/weld/executor/AbstractExecutorServices.java
```diff
@@ -132,4 +133,15 @@ protected void shutdown() {
...
+     *
+     * It allows to wrap the tasks with some additional logic. For instance, Weld's CommonForkJoinPoolExecutorServices
+     * overrides this in order to set TCCL to null prior to execution.
+     */
+    public <T> Collection<? extends Callable<T>> wrap(Collection<? extends Callable<T>> tasks) {
+        //no-op by default, CommonForkJoinPoolExecutorServices override this method in order to set TCCL to null
```
mkouba: I find this comment redundant... `no-op by default` is enough.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/service/executorService/ContextClassLoaderNullifiedTest.java
manovotn: Probably yes, didn't give it much thought since it's inside a test.

manovotn:
Hmm, I am not sure this fix is correct.
We actually permit to use `CommonForkJoinPoolExecutorServices` in servlet env (see [this test](https://github.com/weld/core/blob/3.0.4.Final/environments/servlet/tests/base/src/main/java/org/jboss/weld/environment/servlet/test/bootstrap/configuration/CommonExecutorServicesConfigurationTest.java#L54)). In that case, if you nullify TCCL you enforce system CL instead of `WebappClassLoader` and the test fails.

manovotn:
The underlying problem is [setting the value into the field](https://github.com/weld/core/blob/3.0.4.Final/impl/src/main/java/org/jboss/weld/injection/FieldInjectionPoint.java#L94) in the test. We get IAE when trying to do that. Probably some class loading issue - to be investigated.

manovotn:
retest this please

manovotn:
retest this please

manovotn:
The PR should now fix the problem.
FJP executor now tries to detect the environment (not nice, but best I could come up with) and if it's SE, we `null` the TCCL (thread context class loader), otherwise we leave it as it is.

impl/src/main/java/org/jboss/weld/executor/CommonForkJoinPoolExecutorServices.java
```diff
@@ -42,4 +49,42 @@ public void cleanup() {
...
+        }
+    }
+
+    /**
+     * Tries to detect SE environment by loadability of SE-specific class. If successful, returns true; false otherwise.
+     *
+     * @return Environments.SE if {@link org.jboss.weld.environment.se.Weld} is loadable, null otherwise
+     */
+    private boolean isSeEnvironment() {
+        return Reflections.isClassLoadable("org.jboss.weld.environment.se.Weld", WeldClassLoaderResourceLoader.INSTANCE);
```
mkouba: There is already `org.jboss.weld.interceptor.util.InterceptionUtils.detectEnvironment()`. We should probably extract this logic into a util class. Or find a better way, e.g. enhance the `org.jboss.weld.Container`.

manovotn: I was thinking of extracting it but `InterceptionUtils.detectEnvironment()` cares about SE and servlet, while here we only care about SE (e.g. this method is way simpler).
Not to mention this "hacky" approach probably shouldn't be publicly exposed as common util method.

mkouba: I think it's better to "expose" one hacky util method than to have two hidden almost identical hacky methods. Util classes are not supposed to be public API of Weld anyway.

Also we could try to improve the dection logic so that it's not so hacky...

manovotn:
@mkouba I tried to draft an approach which would store the `Environment` information in `Container` (which in turn gets this from `WeldStartup`). WDYT?

I also updated `InterceptionUtils` to leverage this as well. It's used from static initializer but it "seems" that it can still access this information in time (at least for SE env).

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/service/executorService/ContextClassLoaderNullifiedTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Deployment
+    public static Archive<?> createTestArchive() {
+        return ClassPath.builder().add(ShrinkWrap.create(BeanArchive.class)
+            .addClasses(ContextClassLoaderNullifiedTest.class, DummyBean.class))
+            .build();
+    }
+
+    @Test
+    public void testTcclNullifiedInSeEnvironment() {
+        try (WeldContainer container = new Weld().initialize()) {
```
mkouba: I know Weld SE is using `CommonForkJoinPoolExecutorServices` by default but it wouldn't hurt to add `.add(EXECUTOR_THREAD_POOL_TYPE.get(), COMMON.toString())` to make sure the test is correct even if the default value changes.

impl/src/main/java/org/jboss/weld/Container.java
```diff
@@ -40,6 +41,8 @@
 
     private static Singleton<Container> instance;
 
+    private static Environment environment;
```
mkouba: This field can be accessed concurrently so I would use `AtomicReference` and `AtomicReference#compareAndSet()` in the `setEnvironment()` method below.

impl/src/main/java/org/jboss/weld/Container.java
```diff
@@ -208,4 +215,20 @@ public void setState(ContainerState state) {
         this.state = state;
     }
 
+    /**
+     * Returns {@link Environment} that this container was booted in, or null if not set or invoked too
```
mkouba: We should be clear that this method returns the env the **first** container was started in.

manovotn: Should be, not needn't be true if you start it and not set environment (using one of the older methods, not the new one I added).
I will add this to the doc.

impl/src/main/java/org/jboss/weld/Container.java
manovotn: good point

impl/src/main/java/org/jboss/weld/Container.java
```diff
@@ -208,4 +216,21 @@ public void setState(ContainerState state) {
...
+     * Note that should there be multiple containers active, this will return the {@link Environment} that the first container
+     * was bootstrapped in, unless it was null.
+     *
+     * @return {@link Environment} or null
+     */
+    public static Environment getEnvironment() {
+        return environmentReference.get();
+    }
+
+    private static void setEnvironment(Environment env) {
```
mkouba: Do we actually need this method? I think we could use `Container.environmentReference.compareAndSet(null, env)` directly in constructor...

manovotn: That's what I originally had, but Spotbugs complained that I'm setting value directly to static field.
That was before using `AtomicReference` though, should be ok now, let's see...

manovotn:
Removed that method (spotbugs seemed happy enough now) and squashed commits.

mkouba:
@manovotn Thanks! I wonder what's wrong with setting a static field... spotbugs.

impl/src/main/java/org/jboss/weld/Container.java
```diff
@@ -40,6 +42,8 @@
 
     private static Singleton<Container> instance;
 
+    private static AtomicReference<Environment> environmentReference = new AtomicReference<>(null);
```
mkouba: Actually, this should be a constant, i.e. `private static final AtomicReference<Environment> ENV_REFERENCE = new AtomicReference<>();`

manovotn:
@mkouba from what i recall, the problem with spotbugs only manifests if you are setting the value to a variable which is not thread safe. I recall it was similar to [this problem in SO](https://stackoverflow.com/questions/3630485/whats-the-best-way-to-fix-this-write-to-static-field-from-instance-method-fin).

manovotn:
retest this please

PR #1839, opened 2018-06-06
[WELD-2435] Moved WeldProvider to Servlet Core module

panga:
WeldProvider is only used and referenced by Servlet Core it doesn't make sense to keep it into commons anymore.

Note: SE already has its own provider implementation.

### DIFF
--- a/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/WeldProvider.java
+++ b/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/WeldProvider.java
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.weld.environment;
+package org.jboss.weld.environment.servlet;
 
 import java.util.ArrayList;
 import java.util.Collections;
--- a/environments/servlet/core/src/main/resources/META-INF/services/javax.enterprise.inject.spi.CDIProvider
+++ b/environments/servlet/core/src/main/resources/META-INF/services/javax.enterprise.inject.spi.CDIProvider
@@ -14,4 +14,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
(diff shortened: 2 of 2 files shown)


manovotn:
ok to test

manovotn:
Alright, I did some googling and it seems that when using module path and JPMS, you no longer need `META-INF/services` to declare services.
However, if you have that and use automatic modules, than any such module declaring `META-INF/services/...` (like `servlet-core-impl`) is equal to having a named module and using the `provides x.y.z.MyService with org.bar.foo.MyServiceImpl;` (hence the need to have `WeldProvider` inside `servlet-core-impl` automatic module).
I also found no way to make an automatic module simply use service impl from another module (e.g. equivalent to having `uses x.y.z.MyService` in a named module) apart from not having any META-INF/services which would blow up on JDK 8.

@panga does what I say above make sense? I would like to understand the problem somewhat better before proceeding any further.

manovotn:
Also, for the record, I found this when reading [`ServiceLoader` javadoc](https://docs.oracle.com/javase/9/docs/api/java/util/ServiceLoader.html) which answers part of my previous questions:

> There is no support for a module specifying, in a provides directive, a service provider in another module.

panga:
Hi @manovotn,

> I also found no way to make an automatic module simply use service impl from another module (e.g. equivalent to having uses x.y.z.MyService in a named module) apart from not having any META-INF/services which would blow up on JDK 8.

Yes, this is a design decision of JPMS and I tend to agree with them.
IMHO sharing common classes is good, but I don't like to reuse a provider in two different JARs.

Anyway, that's not the case in Weld, since SE has it's own WeldSEProvider and WeldProvider is only used by Weld Servlet, correct?

Making this move, things won't break in JDK 8 and also will work with JDK 9+ modular applications with automatic modules.

manovotn:
>Anyway, that's not the case in Weld, since SE has it's own WeldSEProvider and WeldProvider is only used by Weld Servlet, correct?

Yes, that is correct.

PR #1840, opened 2018-06-06
[WELD-2504] Remove java.beans dependency of Reflections library

panga:


manovotn:
ok to test

manovotn:
Hi @panga,
thanks for the issue and the PR.
To put things straight, this is only an issue with JDK 10, yes?
There, you introduce an additional dependency on JPMS module, hence increasing the runtime footprint.
Whereas on JDK 8, this doesn't really matter. Am I getting it right?

panga:
Hi @manovotn ,

Conceptually, this issue happens with any application using JPMS (meaning JDK 9 and up + module-info.java).

However, I'm also using jlink to to create minimal JRE for existing applications that are non modularized, in that case, these apps also should benefit from the change.

You are correct, JDK 8 and below doesn't matter because you can only run the whole JRE (not exactly true since it also has compact profile and 'java.beans' are not part of them https://blogs.oracle.com/jtc/an-introduction-to-java-8-compact-profiles )

manovotn:
> I'm also using jlink

Yea, I figured jlink is probably involved as well.

> it also has compact profile

Interesting, that I didn't know.

Thanks for clarification. 
One note though, since we'll now have this in our codebase, it'd be great to have at least some unit tests for it (can be placed in `impl/src/test/...`).

panga:
Done, unit tests added.

impl/src/test/java/org/jboss/weld/tests/unit/util/reflection/ReflectionsTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class ReflectionsTest {
+
+    @Test
+    public void decaptalizeTest() {
+        assertNull(Reflections.decapitalize(null));
+        assertEquals("", Reflections.decapitalize(""));
+        assertEquals("foo", Reflections.decapitalize("foo"));
+        assertEquals("foo", Reflections.decapitalize("Foo"));
+        assertEquals("FOO", Reflections.decapitalize("FOO"));
```
manovotn: I know it's nitpicking, sorry, but to have it test what your javadoc states, you should also have this here:
`AssertEquals("fooBar", Reflections.decapitalize("FooBar"));`

panga: Fair enough, done.

PR #1841, opened 2018-06-07
Ignores .factorypath files in all project modules

smoyer64:
These files seem to be commonly generated in projects that are edited in VS Code (which uses headless Eclipse under the covers).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .project
 .classpath
 .settings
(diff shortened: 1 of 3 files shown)


smoyer64:
@manovotn I've seen it occasionally with Eclipse (and @mkouba previously ignored it in just the test module) but I'm seeing it all the time with Visual Studio Code.

PR #1844, opened 2018-06-18
WELD-1130: Prevent setting a null instance as an attribute

benjamin-confino:
I had a report of an NPE being thrown when resetHttpSessionAttributeOnBeanAccess is enabled. I was able to recreate locally and it looks like a simple null check to prevent AbstractSessionBeanStore from trying to set the attribute before it is created is sufficient.


benjamin-confino:
See https://github.com/weld/core/pull/1845 for the master branch.

PR #1847, opened 2018-06-26
WELD-2512 FacesUrlTransformer should be more strict when looking for …

manovotn:
…cid parameter.


manovotn:
Theoretically this is already tested by some integration tests we have(such as `ClientConversationContextTest`), but I still added some basic unit test.

modules/jsf/src/main/java/org/jboss/weld/module/jsf/FacesUrlTransformer.java
```diff
@@ -31,6 +33,7 @@
     private static final String QUERY_STRING_DELIMITER = "?";
     private static final String PARAMETER_PAIR_DELIMITER = "&";
     private static final String PARAMETER_ASSIGNMENT_OPERATOR = "=";
+    private static final Pattern cidParamPattern = Pattern.compile("(\\?|&)cid=");
```
mkouba: It's good to name constants like `CID_PARAM_PATTERN`...

manovotn: Ah, good point, I extracted it from method and forgot to change it.

PR #1853, opened 2018-06-29
WELD-2507 Abstract package private class not intercepted correctly.

manovotn:


manovotn:
retest this please

manovotn:
Hmm, so just like I expected, my previous '30 min fix' didn't quite cover complex hierarchies.
The tests looks robust enough now, should cover even weirder cases than the originally reported one :)
Thanks for improving the PR!

PR #1856, opened 2018-07-02
WELD-2501 Add reproducer, first attempt to fix this

manovotn:


impl/src/main/java/org/jboss/weld/util/reflection/Reflections.java
```diff
@@ -465,14 +465,20 @@ public static Exception unwrapInvocationTargetException(InvocationTargetExceptio
         }
     }
 
-    public static Set<Class<?>> getInterfaceClosure(Class<?> clazz) {
+    public static Set<Class<?>> getInterfaceClosure(Class<?> clazz, Set<Class<?>> setOfInterfaces) {
```
manovotn: We need to add a unit test to `org.jboss.weld.tests.unit.util.reflection.ReflectionsTest` to have a verification of this new behaviour.

mkouba: Something like this should work:
```java
public static Set<Class<?>> getInterfaceClosure(Class<?> clazz) {
   Set<Class<?>> result = new HashSet<>();  
   for (Class<?> classToDiscover = clazz; classToDiscover != null; classToDiscover = classToDiscover.getSuperclass()) {
     addInterfaces(classToDiscover, result);
   }
   return result;
}
```

impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedSubclassFactory.java
```diff
@@ -238,6 +254,16 @@ void doReturn(CodeAttribute b, ClassMethod method) {
         }
     }
 
+    @Override
+    public Set<Class<?>> getAdditionalInterfaces() {
```
mkouba: I think we should not override the `getAdditionalInterfaces()` method but use the `interfacesToInspect` directly where needed. We don't want to extend the `additionalInterfaces` set as defined in `ProxyFactory` after all.

manovotn:
Agreed via chat that the state of this PR is good to merge.

PR #1857, opened 2018-07-02
WELD-2514 Add testcase and propose a fix

manovotn:


manovotn:
The idea of this fix is to fine-grain the check of known bridge methods.
So far we based our check (for non-null return type) on `bridgeMethod.returnType.equals(returnType)`.
This is insufficient for generics as the `returnType` is in fact a type variable (e.g. `S`) while bridge return type can be a class (`SuperFoo`). That's exactly the case we want to pin down and return `true`.

PR #1858, opened 2018-07-03
WELD-2498 Add test (runnable in both, embedded and container mode) an…

manovotn:
…d make sure Weld does not create AT for local and anonymous classes.

CI job will crash here as it requires API release (with https://github.com/weld/api/pull/76) and also changes on WFLY side which should be sent as part of component upgrade after Weld release.


manovotn:
retest this please

manovotn:
Test failures are expected - WFLY used in CI does not yet implement new SPI.
We tested this locally on patched WFLY where it worked.

PR #1859, opened 2018-07-09
WELD-2515 Synthetic archive should use empty Scanning object instead …

manovotn:
…of null.


manovotn:
Missing automated test - to trigger this behaviour one would need two bean archives (on of them being synthetic) in the test plus flat deployment configuration to trigger merging. Not sure how to write that and avoid using auto discovery.

mkouba:
I think we should also improve the `org.jboss.weld.bootstrap.spi.BeansXml#getScanning()` javadoc so that it's clear wheter it can return `null` or not.

PR #1860, opened 2018-07-09
WELD-2502 Do not force quit SE applications with success code, defaul…

manovotn:
…t behaviour should take care of that.

This change breaks any app which hooks into container lifecycle and then runs in another thread (e.g. Swing app) - for instance our Groovy SE sample.
Removing the line should not affect the exit code as Java always returns 0 unless killed as a process.

### DIFF
--- a/environments/se/core/src/main/java/org/jboss/weld/environment/se/StartMain.java
+++ b/environments/se/core/src/main/java/org/jboss/weld/environment/se/StartMain.java
@@ -55,7 +55,6 @@ public class StartMain {
     public static void main(String[] args) {
         try {
             new StartMain(args).go();
-            System.exit(0);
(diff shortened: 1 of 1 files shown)


smoyer64:
LGTM ... sorry I missed that

manovotn:
No problem, I just wanted to check with you before I push this change.
Thanks!

mkouba:
So the only backward incompatibility we have now is that if the bootstrap fails the JVM process will be terminated with exit code 1 (was 0 previously)...

manovotn:
Theoretically yes but I haven't found any notion in JLS of exit codes and what numbers are to be expected in what situation. E.g. it is undefined and should not be relied upon anyway.

PR #1861, opened 2018-07-10
WELD-2495 Inspect Weld examples, make them JDK 10 compatible, remove obsolete parts

manovotn:
Note that PR executor does **not** build or test examples hence the result of the jobs is irrelevant to the changes contained in this PR.


mkouba:
I wonder if it makes sense to maintain **4** JSF examples..

examples/jsf/login/src/main/webapp/users.xhtml
```diff
@@ -44,11 +44,16 @@
...
 				</tr>
 				<tr>
-					<td></td>
 					<td><h:commandButton action="#{userManager.addUser}" value="Add User" /></td>
 				</tr>
+				<tr>
+					<td><h:link outcome="/home.xhtml">View login page</h:link></td>
+				</tr>
 			</table>
+            <!--TODO FIX THE LOOKS HERE-->
```
mkouba: ??

manovotn: Oops, was fiddling with the looks amidst commits and forgot to erase this. Thanks

manovotn:
@mkouba I was wondering the same but every sample covers different "setup" - WAR, EAR, with other EE technologies and some are deployable to servlets as well. We would need to maybe squash them together a bit.

But I agree that having some more up to date tech wouldn't hurt though I am surely not the one to write that :smile:

mkouba:
Well, I don't see a big differenece between `login` and `numberguess` for example.. but we can revisit this after 3.0.5.

examples/README.md
```diff
@@ -7,14 +7,12 @@ Weld currently comes with a number of examples:
...
-* `se/helloworld` (a simple example for Java SE)
-* `osgi/paint` (a simple OSGi example)
+* `se/groovy-numberguess` (the numberguess in Groovy leveraging Swing)
 
-Before running the examples, you'll need to ensure your server supports CDI (WildFly 8 and
-GlassFish v4 both have built in support, and Weld provides support for Tomcat, Jetty and
-Google App Engine). Weld also supports Java SE and OSGi. 
+Before running the examples, you'll need to ensure your server supports CDI (WildFly and
+GlassFish both have built in support, and Weld provides support for Tomcat and Jetty servlets).
+Weld also supports Java SE and OSGi.
```
nziakova: Not sure but do we want to have the OSGi support mentioned in the examples README after the OSGi examples were removed? Would it be suitable to add "through Pax CDI" as the docs say and point to their examples?

examples/jsf/numberguess/README.md
```diff
@@ -45,7 +45,7 @@ Make sure you have assigned the absolute path of your installation to the
 
 3. Now you can view the application at <http://localhost:8080/weld-numberguess>.
 
-Deploying to standalone Tomcat 8
+Deploying to standalone Tomcat 9
```
nziakova: Can the examples be Tomcat version-agnostic the same way as WildFly version-agnostic?

manovotn: Probably could but depending on Java version you want certain version of Tomcat.
The example should be definite in with which version it will run correctly. And 9 *seems* to do the trick (working on both, Java 8 and 9+).

examples/README.md
manovotn: Good point. We could perhaps remove this notion from examples altogether.
It is enough to have this mentioned in our docs, WDYT?

nziakova: +1

examples/jsf/numberguess/README.md
nziakova: OK, thanks!

manovotn:
Removed the `TODO`, updated SE numberguess README and one pom dependency, removed OSGi notions altogether from examples.

manovotn:
@mkouba with the SM issue moved to a separate JIRA, do you request any additional changes in this PR?

mkouba:
Maybe also create an issue to consider removing the `login` JSF example? Otherwise looks good.

arjantijms:
> But I agree that having some more up to date tech wouldn't hurt though I am surely not the one to write that 

You mean using the upcoming Faces 4.0? ;)

manovotn:
> > But I agree that having some more up to date tech wouldn't hurt though I am surely not the one to write that
> 
> You mean using the upcoming Faces 4.0? ;)

Yeah...no. That's definitely not what I meant :-D

PR #1866, opened 2018-08-13
WELD-2524 Give ProbeExtension highest priority amongst AfterDeploymen…

manovotn:
…tValidation observers.

I haven't managed to reproduce the original problem (it was not reported on WFLY - it works there by chance) so I attached no test. We already have general tests for OM ordering with priority.


mkouba:
I wonder if we should add the priority to all observer methods?

manovotn:
@mkouba I was thinking the same, but it seems that no other parts suffer from this "race condition" so I did just this for the time being.
Also, having to real way to test the impact, I am hesitant to add anything above the minimum required change, but that's just my paranoia :)

mkouba:
Ok, np.

PR #1867, opened 2018-08-13
WELD-2523 Allow to register additional bean defining annotations in W…

manovotn:
…eld SE.

### DIFF
--- a/docs/reference/src/main/asciidoc/environments.asciidoc
+++ b/docs/reference/src/main/asciidoc/environments.asciidoc
@@ -723,6 +723,14 @@ Another approach is to use `Weld.property()` and `SeContainerInitializer.addProp
 
 NOTE: The bean discovery mode of `annotated` is supported from version 2.2.0.Final. Previous versions processed implicit bean archives in the same way as explicit bean archives.
 
+==== Extending Bean Defining Annotations
+
+If you are running with discovery mode `annotated`, then only classes with bean defining annotations will be picked up as beans.
+The set of these annotations is given by CDI but Weld SE allows you to expand it via `Weld.addBeanDefiningAnnotations(Class<? extends Annotation>... annotations)`.
+Any annotation added this way will be considered a bean defining annotation when performing discovery.
+
+Just note that added annotations are ignored if you are also using `<trim/>` option or Weld configuration key `org.jboss.weld.bootstrap.vetoTypesWithoutBeanDefiningAnnotation`.
+
 === Weld SE and Weld Servlet cooperation
 
 Sometimes it could be convenient to start Servlet container programmatically.
--- a/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
+++ b/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
@@ -237,6 +237,8 @@ public class Weld extends SeContainerInitializer implements ContainerInstanceFac
 
     protected final Set<Class<?>> beanClasses;
 
+    protected final Set<Class<? extends Annotation>> extendedBeanDefiningAnnotations;
+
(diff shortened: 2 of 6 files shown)


environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -923,6 +939,8 @@ protected Deployment createDeployment(ResourceLoader resourceLoader, CDI11Bootst
             .addAll(typeDiscoveryConfiguration.getKnownBeanDefiningAnnotations())
             // Add ThreadScoped manually as Weld SE doesn't support implicit bean archives without beans.xml
             .add(ThreadScoped.class)
+            // Add all custom BDA user registered via Weld.addBeanDefiningAnnotations()
```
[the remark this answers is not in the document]
manovotn: Hehe, acronym overload.
Will change that.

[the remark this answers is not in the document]
nziakova: "BDA" is still used in the test - maybe it should be updated too (at least in the comment where it does not hurt to have the full term)?

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -887,6 +890,22 @@ public Weld enableDevMode() {
...
+    /**
+     * Registers annotations which will be considered as bean defining annotations.
+     *
+     * NOTE - If used along with {@code <trim/>} bean archives and/or with Weld configuration key
+     * {@code org.jboss.weld.bootstrap.vetoTypesWithoutBeanDefiningAnnotation}, these annotations will be ignored.
+     *
+     * @param annotations annotations which will be considered as Bean Defining Annotations.
+     * @return self
+     */
+    public Weld addBeanDefiningAnnotations(Class<? extends Annotation>... annotations) {
```
nziakova: Does it make sense to document this, e.g. in 15.6.2. Implicit bean archive (there's a NOTE in that section listing what are bean defining annotations)?

manovotn: Chapter 15 is about EE while this feature only concerns SE.
But you are right that we might want to mention this, but maybe in `18.4.2. Bootstrapping CDI SE` instead?

manovotn:
@nziakova @mkouba added docs as well, take a glance please

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
nziakova: Of course you are right, apologies for confusion.

docs/reference/src/main/asciidoc/environments.asciidoc
```diff
@@ -723,6 +723,14 @@ Another approach is to use `Weld.property()` and `SeContainerInitializer.addProp
 
 NOTE: The bean discovery mode of `annotated` is supported from version 2.2.0.Final. Previous versions processed implicit bean archives in the same way as explicit bean archives.
 
+==== Extending Bean Defining Annotations
+
+If you are running with discovery mode `annotated`, then only classes with bean defining annotations will be picked up as beans.
+The set of these annotations is given by CDI but Weld SE allows you to expand it via `Weld.addBeanDefiningAnnotations(Class<? extends Annotation>... annotations)`.
+Any annotation added this via will be considered a bean defining annotation when performing discovery.
```
nziakova: Look like a wrong word order: this via -> via this? But I don't like "via this" too much, maybe e.g. "this way" would be better?

manovotn: Sure, that's a typo, meant to be `this way`, thanks.

PR #1869, opened 2018-09-10
WELD-2519 Update tests and docs to Tomcat 9

nziakova:


nziakova:
I have added new profile for JDK9+ excluding the failing test.

environments/servlet/tests/tomcat/pom.xml
```diff
@@ -176,48 +171,44 @@
...
-
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <!-- https://issues.apache.org/bugzilla/show_bug.cgi?id=57314 -->
+                                <exclude>**/ListenerSymmetryTest.java</exclude>
```
manovotn: This should be inspected, why do we still have it excluded and what can we do about it?

manovotn: Also, the `incontainer` profile seems to be excluding it already, is this duplication needed?

docs/reference/src/main/asciidoc/environments.asciidoc
```diff
@@ -92,7 +92,7 @@ Setting the identifier is as simple as adding one context parameter into `web.xm
 
 ==== Tomcat
 
-Tomcat 7 and 8 are supported. Context activation/deactivation and dependency
+Tomcat 9 is supported. Context activation/deactivation and dependency
 injection into Servlets and Filters works out of the box. Injection into Servlet listeners works on
 Tomcat 7.0.50 and newer.
```
manovotn: Looking at this I am thinking we might as well drop the last two lines (as we are long past 7.0.50 and it all works out of the box now). Instead we could maybe declare why is 9+ supported (e.g. alignment with EE 8 and Servlet 4 specs?)

environments/servlet/tests/tomcat/pom.xml
```diff
@@ -158,6 +152,7 @@
                 <property>
                     <name>incontainer</name>
                 </property>
+                <jdk>(,1.8]</jdk>
```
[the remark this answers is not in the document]
nziakova: Not to activate this profile for JDK9+, i.e. activate only when incontainer property is set and JDK<=1.8. Without having this, both incontainer and jdk9 profiles are activated (mvn help:active-profiles). TBH I'm not sure what the result actually is when having both profiles activated because when running the tests, the result looks the same regardless of this line. I think it might be because the profiles are picked in some order - I just don't know if the order is well defined or if it is just a coincidence.

environments/servlet/tests/tomcat/pom.xml
nziakova: Looks like it is needed as Maven does not merge the exclusions in 2 profiles; it uses the last one instead.

environments/servlet/tests/tomcat/pom.xml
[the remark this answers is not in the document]
nziakova: You are right, the jdk condition is not needed here, thanks! The test exclusion is still needed though as related [Tomcat BZ](https://bz.apache.org/bugzilla/show_bug.cgi?id=57314) was closed as WONTFIX. The same test works on other servlet containers (e.g. Jetty) so we probably should not remove it entirely.

environments/servlet/tests/tomcat/pom.xml
nziakova: The test exclusion is still needed as related [Tomcat BZ](https://bz.apache.org/bugzilla/show_bug.cgi?id=57314) was closed as WONTFIX. The same test works on other servlet containers (e.g. Jetty) so we probably should not remove it entirely.

nziakova:
I have addressed all your comments except for the test exclusion as explained above.

PR #1870, opened 2018-09-13
WELD-2532 Add AtomicBoolean check on bean index creation.

manovotn:
Backport of WELD-2492


manovotn:
Note - this PR has no tests as it isn't easily reproducible, only the reporter was able to reproduce it in his environment (and we had the fix verified the same way).

PR #1873, opened 2018-09-19
WELD-2533 InterceptionFactory with interface.

manovotn:
A PoC how `InterceptionFactory` could work on interfaces.
The main idea revolves around tempering with the AT of the actual implementation class provided by user and adding extra binding which were configured on another AT (interface).

Thoughts and suggestions are welcome.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedProxyFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedProxyFactory.java
@@ -42,6 +42,7 @@ import org.jboss.weld.security.GetDeclaredMethodsAction;
 import org.jboss.weld.util.bytecode.BytecodeUtils;
 import org.jboss.weld.util.bytecode.MethodInformation;
 import org.jboss.weld.util.bytecode.RuntimeMethodInformation;
+import org.jboss.weld.util.reflection.Reflections;
 
 /**
  * Generates proxies used to apply interceptors to custom bean instances and return values of producer methods.
@@ -61,10 +62,12 @@ public class InterceptedProxyFactory<T> extends ProxyFactory<T> {
 
     private final Set<MethodSignature> interceptedMethodSignatures;
 
+    private Set<Class<?>> interfacesToInspect;
+
     private final String suffix;
 
     public InterceptedProxyFactory(String contextId, Class<?> proxiedBeanType, Set<? extends Type> typeClosure, Set<MethodSignature> enhancedMethodSignatures,
-            Set<MethodSignature> interceptedMethodSignatures, String suffix) {
+        Set<MethodSignature> interceptedMethodSignatures, String suffix) {
         super(contextId, proxiedBeanType, typeClosure, null);
         this.enhancedMethodSignatures = enhancedMethodSignatures;
         this.interceptedMethodSignatures = interceptedMethodSignatures;
@@ -73,6 +76,28 @@ public class InterceptedProxyFactory<T> extends ProxyFactory<T> {
 
     protected String getProxyNameSuffix() {
         return PROXY_SUFFIX + suffix;
+    }
+
+    @Override
+    public void addInterfacesFromTypeClosure(Set<? extends Type> typeClosure, Class<?> proxiedBeanType) {
+        // these interfaces we want to scan for method and our proxies will implement them
+        for (Class<?> c : proxiedBeanType.getInterfaces()) {
+            addInterface(c);
+        }
+        // now we need to go deeper in hierarchy and scan those interfaces for additional interfaces with default impls
+        for (Type type : typeClosure) {
+            Class<?> c = Reflections.getRawType(type);
+            if (c.isInterface()) {
+                addInterfaceToInspect(c);
+            }
+        }
+    }
+
+    private void addInterfaceToInspect(Class<?> iface) {
+        if (interfacesToInspect == null) {
+            interfacesToInspect = new HashSet<>();
+        }
+        this.interfacesToInspect.add(iface);
     }
 
     @Override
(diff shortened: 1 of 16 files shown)


impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
```diff
@@ -137,4 +169,25 @@ public T createInterceptedInstance(T instance) {
         return proxy;
     }
 
+    private boolean isInterceptorBinding(Annotation annotation) {
+        return annotation.annotationType().getAnnotation(InterceptorBinding.class) != null;
```
mkouba: You should probably use `BeanManager#isInterceptorBinding(Class<? extends Annotation>)` instead..

impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
```diff
@@ -137,4 +169,25 @@ public T createInterceptedInstance(T instance) {
...
+        if (first == null || second == null) {
+            return false;
+        }
+        if (!first.getName().equals(second.getName())
+            || !first.getReturnType().equals(second.getReturnType())
+            || !(first.getParameterCount() == second.getParameterCount())) {
+            return false;
+        }
+        List<Class<?>> secondMethodParams = Arrays.asList(second.getParameterTypes());
+        for (Class<?> paramType : first.getParameterTypes()) {
```
mkouba: This would not work correctly for methods with generic params, or? Maybe `Method.getGenericParameterTypes()`?

impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
```diff
@@ -112,15 +118,41 @@ public T createInterceptedInstance(T instance) {
...
         used = true;
 
-        if (annotatedType.getJavaClass().isInterface()) {
-            throw InterceptorLogger.LOG.interceptionFactoryNotOnInstance(annotatedType.getJavaClass().getCanonicalName());
+        AnnotatedType<T> originalAT = configurator != null ? configurator.complete() : annotatedType;
+        // this is the resulting AT we want to work with
+        AnnotatedType<T> resultingAnnotatedType = null;
+        // if it is an interface, we will have to 'merge' the AT with the AT of whatever user supplied as an instance
+        if (originalAT.getJavaClass().isInterface()) {
+            AnnotatedType<?> instanceAnnotatedType = beanManager.createAnnotatedType(instance.getClass());
```
mkouba: In the future it might be useful to have this extracted into a reusable unit of code ;-)

impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
```diff
@@ -112,15 +118,41 @@ public T createInterceptedInstance(T instance) {
...
+            AnnotatedTypeConfiguratorImpl<?> newAtConfigurator = new AnnotatedTypeConfiguratorImpl<>(instanceAnnotatedType);
+
+            // first check class-level annotation
+            for (Annotation annotation : originalAT.getAnnotations()){
+                if (isInterceptorBinding(annotation)){
+                    newAtConfigurator.add(annotation);
```
mkouba: I believe you should first check whether an annotation of the same annotated type is already present or not - interceptor bindings may have different values. In fact, different values should result in definition error.

impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
manovotn: Good point, will look at that.

impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
manovotn: Not sure, will try to put together a test to prove that (and add it here).

impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
manovotn: Not sure I want to see this kind of "AT merging" anywhere else ;-)
But yea, I can create some util class and put it there.

impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
manovotn: Yes, thought about it but for my first version, this was "good enough".
Do you happen to know if we have similar behaviour in, say, configurators?

mkouba: I don't think so - it's purely the responsibility of extensions ;-)

manovotn: Reminder to self - still have to add a test for this and then a fix.
Not entirely sure what is the 'correct' approach - we can either override anything user did (thinking that user actually wanted to do that) or go fool-proof and blow up if it goes against other annotations. Either way, we first need to detect it.

impl/src/main/java/org/jboss/weld/injection/InterceptionFactoryImpl.java
manovotn: I tried this but in the end decided not to do that.
The code relies on a hierarchy of those two ATs, works only on interceptor bindings and ultimately needs to throw factory specific exception.
I don't see much re-usability there to justify it, at least for the time being.

manovotn:
This should be working approach to allow `InterceptionFactory` to operate on an interfaces, the question of whether we should support it still stand though (we just know we could).

See the conversation on [this CDI issue](https://issues.jboss.org/browse/CDI-686).

@nziakova @mkouba reviews, additional untested corner cases or just your plain thoughts on either my impl or the nature of this "feature" are very much welcome ;-)

manovotn:
A friendly poke for reviews @mkouba @nziakova ;-)

impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedProxyFactory.java
```diff
@@ -75,6 +78,28 @@ protected String getProxyNameSuffix() {
         return PROXY_SUFFIX + suffix;
     }
 
+    @Override
+    public void addInterfacesFromTypeClosure(Set<? extends Type> typeClosure, Class<?> proxiedBeanType) {
```
manovotn: Theoretically, this covers the case when user supplies `FooInterface` for `FooImpl` class and adds a class-level annotation and at the same time there is a `SuperFooInterace` with some default method.
The above (crazy) case needs to have a test added before we merge this.

manovotn:
We also need to temper Weld documentation to mention this possibility (which is beyond the scope of what CDI defines ATM).

manovotn:
Rebased onto master.
All comments were accounted for and the feature is now documented. Will merge this once tests pass.

PR #1875, opened 2018-10-18
WELD-2538 Make WeldManager injectable.

manovotn:


mkouba:
I wonder if we could just extend the set of bean types of `BeanManagerImplBean` instead?

manovotn:
Had the same thought, but wasn't sure it's good approach to re-use the same bean for that.
I'll do that.

manovotn:
@mkouba updated

PR #1876, opened 2018-10-18
WELD-2537, WELD-2539 Add WeldManager#isContextActive method. Add WeldManager#getScopes.

manovotn:
Depends on #1875 (for test) and on https://github.com/weld/api/pull/79 (for API).

### DIFF
--- a/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/weldManager/scopes/GetScopesTests.java
+++ b/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/weldManager/scopes/GetScopesTests.java
@@ -1,0 +1,62 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2018, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.weld.environment.se.test.weldManager.scopes;
+
+import java.lang.annotation.Annotation;
+import java.util.Collection;
+
+import org.jboss.arquillian.container.se.api.ClassPath;
+import org.jboss.arquillian.container.test.api.Deployment;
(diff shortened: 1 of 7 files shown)


mkouba:
Shouldn't we test `getActiveContexts()` and `getActiveWeldAlterableContexts()` as well?

manovotn:
I was planning on using those two as part of the context propagation tests.
But if you think it better, we can have some basic coverage here as well.

manovotn:
Added the tests for active [weld alterable] contexts.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/weldManager/scopes/GetScopesTests.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Test
+    public void testGettingAllScopesInApplication() {
+        try (WeldContainer container = new Weld().initialize()) {
+            WeldManager wm = container.select(WeldManager.class).get();
+            Collection<Class<? extends Annotation>> scopes = wm.getScopes();
+            // test brings no additional scopes we are in SE, hence 7 scopes
+            // Session, request, conversation, application, singleton, thread, dependent
+            Assert.assertEquals(7, scopes.size());
```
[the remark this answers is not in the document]
manovotn: Well, there isn't. The point of the method is to give you access to just that :)
The default is unlikely to change (although if the test were in WFLY, the set would be different).

impl/src/main/java/org/jboss/weld/bean/builtin/BeanManagerProxy.java
```diff
@@ -272,4 +273,13 @@ public BeanManagerImpl unwrap() {
         delegate().disposeAnnotatedType(type, id);
     }
 
+    @Override
+    public boolean isContextActive(Class<? extends Annotation> scopeType) {
```
nziakova: Should this be documented?

impl/src/main/java/org/jboss/weld/bean/builtin/BeanManagerProxy.java
```diff
@@ -272,4 +273,13 @@ public BeanManagerImpl unwrap() {
...
+        return delegate().isContextActive(scopeType);
+    }
+
+    @Override
+    public Collection<Class<? extends Annotation>> getScopes() {
```
nziakova: Should this be documented?

manovotn: Yes, but not quite yet - what needs documenting is Weld API and we are not sure about it's final form yet. But granted, we need a backing JIRA to keep in mind to document all API changes once we settle them - I'll create that.

manovotn:
retest this please

impl/src/main/java/org/jboss/weld/bean/builtin/BeanManagerProxy.java
nziakova: Makes sense, thanks.

PR #1877, opened 2018-11-01
WELD-2497 Context propagation for built-in scopes.

manovotn:
A basis PR for discussions around context propagation in Weld (WELD-2497)

Shows propagation for buillt-in contexts (req., session, conversation + application).
There are two very similar tests, in SE and in EE.
Both test the same thing however note that for instance `RequestContext` implementation in SE differs versus that in EE (unbound versus bound versus HTTP). Also only subset of scopes can be used in SE.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/contexts/AbstractBoundContext.java
+++ b/impl/src/main/java/org/jboss/weld/contexts/AbstractBoundContext.java
@@ -16,8 +16,16 @@
  */
 package org.jboss.weld.contexts;
 
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+
 import org.jboss.weld.context.BoundContext;
+import org.jboss.weld.context.api.ContextualInstance;
+import org.jboss.weld.contexts.beanstore.BeanStore;
 import org.jboss.weld.contexts.beanstore.BoundBeanStore;
+import org.jboss.weld.contexts.beanstore.LockedBean;
+import org.jboss.weld.serialization.spi.BeanIdentifier;
 
 /**
  * Base class for contexts using a thread local to store a bound bean context
@@ -83,6 +91,40 @@ public abstract class AbstractBoundContext<S> extends AbstractManagedContext imp
             }
         } else {
             return false;
+        }
+    }
+
+    @Override
+    public Collection<ContextualInstance<?>> getAllContextualInstances() {
+        Set<ContextualInstance<?>> result = new HashSet<>();
+        // for instance lazily initialized conversation scope may be active but have null here
+        BeanStore beanStore = getBeanStore();
+        if (beanStore != null) {
+            getBeanStore().iterator().forEachRemaining((BeanIdentifier beanId) -> {
+                result.add(getBeanStore().get(beanId));
+            });
+        }
+        return result;
+    }
+
+    @Override
+    public void clearAndSet(Collection<ContextualInstance<?>> setOfInstances) {
+        // bound context can be multithreaded, in such case we perform locking
+        BoundBeanStore boundBeanStore = getBeanStore();
+        boundBeanStore.clear();
+        for (ContextualInstance<?> contextualInstance : setOfInstances) {
+            BeanIdentifier id = getId(contextualInstance.getContextual());
+            LockedBean lock = null;
+            try {
+                if (isMultithreaded()) {
+                    lock = boundBeanStore.lock(id);
+                }
+                getBeanStore().put(getId(contextualInstance.getContextual()), contextualInstance);
+            } finally {
+                if (lock != null) {
+                lock.unlock();
+                }
+            }
         }
     }
 }
--- a/tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
+++ b/tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
@@ -1,0 +1,106 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2018, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.weld.tests.contexts.propagation;
+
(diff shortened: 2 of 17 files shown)


manovotn:
Topics to discuss (feel free to expand):
* Custom scope propagation
  * Any context implementing `WeldAlterableContext` can be propagated
  * But we lack means to easily activate custom context? Needed API change?
* Test templates
  * Is the thread offloading used in test realistic? Should we use something else?
* Corner case coverage
  * Tests to add? What is the basic scenario we will be pursuing for this release?
* Documentation
  * Once we settle on how this should work, we should create a documentation which not only describes what it is but also shows it with code
  * If tests are well designed, we could use them as example?

tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
```diff
@@ -0,0 +1,111 @@
...
+        for (WeldAlterableContext context : CDI.current().select(WeldManager.class).get().getActiveWeldAlterableContexts()) {
+            // TODO shouldnt context.getAllContextualInstances() return Collection<ContextualInstance<Object>>> instead? what is T for there
+            scopeToContextualInstances.put(context.getScope(), context.getAllContextualInstances());
+        }
+        // We create a task wrapper which will make sure we have contexts propagated
+        Callable<T> wrappedTask = new Callable<T>() {
+
+            @Override
+            public T call() throws Exception {
+                WeldManager weldManager = CDI.current().select(WeldManager.class).get();
```
manovotn: I used `CDI.current()` to get hold of the container in another thread. Ideas on hold to do this in a cleaner way?

tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
```diff
@@ -0,0 +1,111 @@
...
+public class ContextPropagationService {
+
+    private static final ExecutorService executor = Executors.newFixedThreadPool(1);
+
+    public static <T> Future<T> propagateContextsAndSubmitTask(Callable<T> task) {
+        // gather all the contexts we want to propagate and the instances in them
+        Map<Class<? extends Annotation>, Collection<ContextualInstance<Object>>> scopeToContextualInstances = new HashMap<>();
+        WeldManager get = CDI.current().select(WeldManager.class).get();
+        for (WeldAlterableContext context : CDI.current().select(WeldManager.class).get().getActiveWeldAlterableContexts()) {
+            // TODO shouldnt context.getAllContextualInstances() return Collection<ContextualInstance<Object>>> instead? what is T for there
```
manovotn: Not sure why we originally wanted `WeldAlterableContext.getAllContextualInstances()` to return `Collection<ContextualInstance<T>>>`, the `T` doesn't really bring any value, might as well be `?` or plain old `Object`?

tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
```diff
@@ -0,0 +1,111 @@
...
+                }
+                if (scopeToContextualInstances.get(conversationContext.getScope()) != null) {
+                    conversationContext.clearAndSet(scopeToContextualInstances.get(conversationContext.getScope()));
+                }
+
+                // now execute the actual original task
+                T result = task.call();
+
+                // cleanup, context deactivation
+                // TODO - this can mess up state in original thread?
```
manovotn: I think this will trigger `@PreDestroy` calls (if there are any) on the beans. However, those beans may still be active in the original thread which would corrupt their state. Untested ATM.

impl/src/main/java/org/jboss/weld/contexts/AbstractBoundContext.java
```diff
@@ -85,4 +92,25 @@ public boolean dissociate(S storage) {
...
+        if (beanStore != null) {
+            getBeanStore().iterator().forEachRemaining((BeanIdentifier beanId) -> {
+                result.add(getBeanStore().get(beanId));
+            });
+        }
+        return result;
+    }
+
+    @Override
+    public <T> void clearAndSet(Collection<ContextualInstance<T>> setOfInstances) {
```
mkouba: This needs to be synchronized in some way for multithreaded contexts.

manovotn: By that you mean session and conversation I suppose? And only the HTTP backed I presume, so we could override it there perhaps. Not sure how to sync it yet, but good point.

tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
manovotn: It might be better to just `deactivate()` the context, skipping the pre destroy calls here.
Should be cleared at MP conc spec level.

manovotn:
Documentation is part of https://github.com/weld/core/pull/1880

tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
```diff
@@ -0,0 +1,111 @@
...
+import org.jboss.weld.context.bound.BoundConversationContext;
+import org.jboss.weld.context.bound.BoundLiteral;
+import org.jboss.weld.context.bound.BoundRequest;
+import org.jboss.weld.context.bound.BoundRequestContext;
+import org.jboss.weld.context.bound.BoundSessionContext;
+import org.jboss.weld.context.bound.MutableBoundRequest;
+import org.jboss.weld.manager.api.WeldManager;
+
+/**
+ * Util class allowing to offload tasks to another thread. Takes case of context activation/propagation.
```
nziakova: Takes care...?

tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
manovotn: For the sake of test, this should be fine.

tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
manovotn: Will be changed via https://github.com/weld/api/pull/82

tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationService.java
manovotn: I have decided to leave out `invalidate()` in this test with a comment explaining it. We can always chose otherwise but since this test is for now the only "how-to reference" for context propagation, I would like to have it minimal and clear.

manovotn:
In the last commit I have added locking for when we populate bean store - this should prevent some conc. access problem with multi threaded contexts supporting propagation.

@mkouba @nziakova I will leave this open till Mon for any additional comments and edits - after that I would like to squash&merge it and proceed with release.

PR #1881, opened 2018-11-20
WELD-2508 Give Weld-specific methods on proxies prefixed names to avo…

manovotn:
…id collision with user code.

@mkouba @nziakova I took a result of a test with decorator, interceptor and client proxy in place, decompiled proxies and came up with methods from - `TargetInstanceProxy` interface, `ProxyObject` interface and `LifecycleMixin`.
`LifecycleMixin` already uses obscure names, so I left that one untouched and only altered the other two.

I also left methods on `WeldClientProxy` untouched as those are part of API and have specific return type to avoid these kinds of problems. Not to mention backward compatibility.


mkouba:
Looks good but let's wait for the CI results ;-)

PR #1886, opened 2018-12-03
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these LGTM [code quality badges](https://lgtm.com/projects/g/weld/core/ci/#badges) to your project. They indicate how you care about code quality and encourage your future contributors to do the same.
To get an idea of the analyses reflected by these grades, check the [alerts discovered by LGTM](https://lgtm.com/projects/g/weld/core).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.


manovotn:
Hello @xcorail 

While we most certainly care about code quality, I don't think we need to show it off with a fancy badge.
Besides on the Java side, which is main here, we already use SpotBugs which does something similar as your LGTM alerts and is baked into every PR done even now.

Furthermore I checked the alerts LGMT is giving and loads of them are false positives meaning a non trivial time would be needed to eliminate them (which is again already done for SpotBugs).

So thanks for the suggestion, we might consider this some time later, but for now, I am rejecting the PR.

xcorail:
Hello @manovotn 
Thanks very much for this feedback.

PR #1887, opened 2018-12-12
WELD-2529 Update WildFly versions in the docs

nziakova:
Additional changes will be needed for GlassFish, see https://issues.jboss.org/browse/WELD-2552

### DIFF
--- a/docs/reference/src/main/asciidoc/environments.asciidoc
+++ b/docs/reference/src/main/asciidoc/environments.asciidoc
@@ -3,12 +3,12 @@
 
 === Using Weld with WildFly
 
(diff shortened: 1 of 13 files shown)


manovotn:
@nziakova please look at the CI failures, servlet tests seem to be failing with the JSF version bump.

nziakova:
@manovotn I have reverted the JSF version changes to only bump the versions in docs and examples as the JIRA says. I have to investigate why Tomcat (and Jetty) tests fail with JSF 2.3. The numberguess example runs fine on Tomcat according to README instructions, even with JSF 2.3.

PR #1888, opened 2019-01-15
WELD-2556 Weld core changes reflecting API changes.

manovotn:
First draft on Weld core changes reflecting https://github.com/weld/api/pull/83.

With this I was able to proceed with changes on custom WFLY and verify that our tests pass.

CI builds will fail due to API upgrade.


impl/src/main/java/org/jboss/weld/util/bytecode/ClassFileUtils.java
```diff
@@ -45,6 +45,8 @@ private ClassFileUtils() {
 
     static {
         try {
+            // TODO this needs to go away anyway, else we would blow up on JDK 12 even if we have the new approach
+            // probably revert to setAccessible() for some later version?
```
manovotn: I still need to think how to handle this static block (which is what causes JDK 12 exceptions) and if we want to address it now.

We could at least move this to a separate method (which we would need to thread-safely invoke only once) that wouldn't get invoked if we know we can use the other approach. That way, we would at least avoid crashing if the integrator did their part.

mkouba: > I still need to think how to handle this static block (which is what causes JDK 12 exceptions)

What's exactly wrong with static initializer in JDK12?

manovotn:
The WFLY bits I am playing around with can be seen on my branch - https://github.com/manovotn/wildfly/tree/weld2556 though the server isn't easily buildable (varying based on what JDK you want to use, you will need custom classfilewriter, jboss-modules, widfly-core).

impl/src/main/java/org/jboss/weld/util/bytecode/ClassFileUtils.java
manovotn: Not with initializer, with the following line that we invoke:
`theUnsafe.objectFieldOffset(AccessibleObject.class.getDeclaredField("override"));`

Having it in initializer means it will get called every time, even should we use the new API and don't need this at all.

mkouba: Ok, then we could probably use static field of type [LazyValueHolder](https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/util/LazyValueHolder.java) and call `getDeclaredMethod()` in its `computeValue()`?

manovotn:
Added two more commits, one of which fixes tests. Namely `ReentrantComutingCacheTest` since in JDK 12, the `Integer` class implements two additional interfaces (`ConstantDesc`, `Constable`). In order to stay version-agnostic, I reverted the test to use our own test classes instead. The other test was, again, using Unsafe so I reverted that to setAccessible as a temporary measure for JDK 12.

Second commit transforms the static initializer on `ClassFileUtils` (which causes JDK 12 errors) into static method and moves the invocation into Weld bootrap (to make sure this initialization happens only once) and only executes it if integrator doesn't implement new API. It also attempts to use Unsafe first (which covers JDK 8 - 11) but reverts to `setAccessible()` if Unsafe yields error (JDK 12). Note that `setAccessible()` will trigger illegal access warning in console, but ultimately it will at least work.

What are your thoughts on this approach @mkouba @nziakova ?

impl/src/main/java/org/jboss/weld/bootstrap/WeldStartup.java
```diff
@@ -213,6 +214,11 @@ public WeldRuntime startContainer(String contextId, Environment environment, Dep
         if (!registry.contains(ProxyServices.class)) {
             registry.add(ProxyServices.class, new SimpleProxyServices());
         }
+        // check if ProxyServices implementation supports class defining on integrator side
+        if (!registry.get(ProxyServices.class).supportsClassDefiningAndLoading()) {
```
manovotn: We should also add  a static `AtomicBoolean` based on which we will invoke the method (in addition to integrators implementing new SPI). That will prevent the code from being executed multiple time as `WeldStartup` is executed once _per deployment_.

manovotn:
Updated per changes in API.
Added `AtomicBoolean` check to CFUtils per @mkouba suggestion.

manovotn:
Again, JDK 11 failure is unrelated.

PR #1889, opened 2019-01-21
WELD-2535 correct Weld.handleJar behaviour.

manovotn:
Prevents a problem of picking multiple packages from a JAR that has two or more package starting with the same String (e.g. `org.mine.foo`, `org.mine.foobar`).


mkouba:
I think you should be able to test this with `arquillian-container-se` (used in `weld-se-tests`), or?

manovotn:
retest this please

manovotn:
@mkouba added the automated test. Thanks for the hint! :)

PR #1892, opened 2019-01-22
WELD-2551: Consider interface default methods when using WithAnnotations

mkouba:


environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/jandex/JandexClassFileInfo.java
```diff
@@ -291,9 +292,34 @@ private boolean containsAnnotation(ClassInfo classInfo, DotName requiredAnnotati
...
+            for (MethodInfo method : interfaceInfo.methods()) {
+                if (isDefault(method) && method.hasAnnotation(requiredAnnotationName)) {
+                    return true;
+                }
+            }
+        }
         return false;
     }
 
+    private boolean isDefault(MethodInfo method) {
```
manovotn: This method should either check whether the method is declared on an interface or it should be renamed to something more general such as `isPublic()`. In future it could be leveraged elsewhere in which case someone might forget to only invoke it on interface methods.

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/annotatedType/withAnnotations/WithAnnotationsTest.java
```diff
@@ -57,4 +59,12 @@ public void test() {
         assertNotNull(extension.getGroupType());
         assertEquals(Group.class, extension.getGroupType().getJavaClass());
     }
+    
+    // We need to update WildFly ClassFileInfo first
```
manovotn: Maybe we can have this test as a separate PR (and not just for embedded container) and put it on hold until we fix this on WFLY side. CI is using nightly builds, so once we fix it there, we can easily re-run the PR and see the results.

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/jandex/JandexClassFileInfo.java
```diff
@@ -300,15 +300,23 @@ private boolean containsAnnotation(ClassInfo classInfo, DotName requiredAnnotati
...
             }
             for (MethodInfo method : interfaceInfo.methods()) {
-                if (isDefault(method) && method.hasAnnotation(requiredAnnotationName)) {
-                    return true;
+                if (isDefault(method)) {
+                    if (method.hasAnnotation(requiredAnnotationName)) {
+                        return true;
+                    }
+                    // Meta-annotations
+                    for (AnnotationInstance annotation : method.annotations()) {
```
manovotn: We are going to need a test case for this as well ;-)

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/jandex/JandexClassFileInfo.java
mkouba: Good point. On the other hand this method is private... `isNonAbstractPublicInstanceMethod()`?

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/annotatedType/withAnnotations/WithAnnotationsTest.java
manovotn: Well, that was just really an idea so that we don't forget to enable this test on WFLY later on.

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/jandex/JandexClassFileInfo.java
mkouba: Eh, well you're so strict ;-)

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/jandex/JandexClassFileInfo.java
manovotn: +1

manovotn:
retest this please

PR #1895, opened 2019-01-28
WELD-2517: Weld SE - fix SecurityManager-related issues.

mkouba:


manovotn:
We don't really have full SM testsuite but I think it'd be fitting to verify this the way we discovered this in the first place. @nziakova that should be the SE SM example, right?

mkouba:
Yep, that's what I did - run the numberguess SE example with SM manually.

nziakova:
Yes, the se/numberguess one. Works fine now, thanks!

PR #1896, opened 2019-01-30
WELD-2550 Revert previous approach and implement new one through interfaces

manovotn:
Keeps previous tests and adds bunch of new ones.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedProxyFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedProxyFactory.java
@@ -16,16 +16,6 @@
  */
 package org.jboss.weld.bean.proxy;
 
-import static org.jboss.classfilewriter.util.DescriptorUtils.isPrimitive;
-import static org.jboss.classfilewriter.util.DescriptorUtils.isWide;
-
-import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
-import java.lang.reflect.Type;
-import java.security.AccessController;
-import java.util.HashSet;
-import java.util.Set;
-
 import org.jboss.classfilewriter.ClassFile;
 import org.jboss.classfilewriter.ClassMethod;
 import org.jboss.classfilewriter.DuplicateMemberException;
@@ -44,10 +34,22 @@ import org.jboss.weld.util.bytecode.MethodInformation;
 import org.jboss.weld.util.bytecode.RuntimeMethodInformation;
 import org.jboss.weld.util.reflection.Reflections;
 
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.Type;
+import java.security.AccessController;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import static org.jboss.classfilewriter.util.DescriptorUtils.isPrimitive;
+import static org.jboss.classfilewriter.util.DescriptorUtils.isWide;
+
 /**
  * Generates proxies used to apply interceptors to custom bean instances and return values of producer methods.
  *
  * @author Martin Kouba
+ * @author Matej Novotny
  *
  * @param <T>
  * @see InterceptionFactoryImpl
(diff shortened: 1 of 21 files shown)


impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedProxyFactory.java
```diff
@@ -142,35 +133,33 @@ protected void addMethodsFromClass(ClassFile proxyClassType, ClassMethod staticC
...
-                            } else {
-                                createNotInterceptedMethod(classMethod, methodInfo, method, staticConstructor);
+            if (builtFromInterface) {
+                // since we are just on top of an interface, we will need to add all the methods from interface
+                // hierarchy so that there are no AbstractMethodError exception popping up
+                for (Class<?> c : interfacesToInspect) {
+                    for (Method method : c.getMethods()) {
+                        MethodSignature signature = new MethodSignatureImpl(method);
+                        try {
+                            if (isMethodAccepted(method, getProxySuperclass())
```
manovotn: I am not really sure what is the correct check here. I have simplified it a lot because we are operating on method which belong to interface classes only. Compared to previous solution:

* Using `enhancedMethodSignatures.contains(signature)` won't work as those methods are not part of the signatures we know. The reason is that signatures seem to be built only from impl methods and default methods in interfaces.
* Removed `!finalMethods.contains(signature)` is pointless, there cannot be final methods on interfaces
* Retained `!processedBridgeMethods.contains(signature)` - not sure this was a smart move, TBF I am not 100% we can see bridge method on an interface?

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
```diff
@@ -138,27 +138,55 @@ public void testFactoryFromGenericInterface(@Produced SomeGenericInterface<List<
...
-    public void testFactoryFromInterfaceWithTypeAnnotation() {
-        try {
-            beanInstance.select(UninterestingInterface2.class, Produced.Literal.INSTANCE).get().ping();
-            Assert.fail();
-        } catch (IllegalStateException e) {
-            // expected
-        }
+    public void testFactoryFromGenericInterfaceStructure(@Produced InterfaceWithGenericsB<String, Integer> bean) {
+        Producer.reset();
+        bean.ping(); // this doesn't trigger anything as we don't inherit annotations ATM
```
manovotn: Class-level annotation is not applied to a predecessor in interface hierarchy. Spec doesn't define this, it's up to us.
Ideas on how it should behave?

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
```diff
@@ -138,27 +138,55 @@ public void testFactoryFromGenericInterface(@Produced SomeGenericInterface<List<
...
+        assertEquals("Hello " + UnproxyableInterfaceWithGenericsChainImpl.class.getSimpleName(), bean.pong());
+        assertEquals(1, Producer.INVOCATIONS.size());
+    }
+
+    @Test
+    public void testFactoryWhenInterfaceHasEqualAnnotationsToThoseWeAreAdding(@Produced InterfaceWithAnnotation bean) {
+        // Currently, we do not verify this shouldn't happen, checking it would require comparison between
+        // added annotation and the AT we create from interface
+        // hence this test should just "silently succeed"
```
manovotn: Another grey area, do we want to resolve clashes when there is a binding on interface and we add one? I am not sure we should as normally you need to place interceptor bindings on bean class anyway.

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
```diff
@@ -138,27 +138,54 @@ public void testFactoryFromGenericInterface(@Produced SomeGenericInterface<List<
         assertEquals(2, Producer.INVOCATIONS.size());
     }
 
-    @Inject
-    @Any
```
nziakova: This makes the imports unused.

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -17,32 +17,6 @@
 
 package org.jboss.weld.bean.proxy;
 
-import static org.jboss.classfilewriter.util.DescriptorUtils.isPrimitive;
```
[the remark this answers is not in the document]
manovotn: They are of utmost importance! At least to IntelliJ it seems... ;-)
You are right, I probably accidentally formated the whole class.

impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedProxyFactory.java
```diff
@@ -62,61 +64,47 @@
 
     private final Set<MethodSignature> interceptedMethodSignatures;
 
-    private Set<Class<?>> interfacesToInspect;
-
     private final String suffix;
 
+    private boolean builtFromInterface;
```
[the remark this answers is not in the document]
manovotn: Good point

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
manovotn: Thanks

impl/src/main/java/org/jboss/weld/bean/proxy/InterceptedProxyFactory.java
mkouba: Hm, even interfaces can have bridge methods. So it's ok to retain the check.

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
mkouba: I'd say we should follow the JLS and `@Inherited`, i.e. annotations are not inherited by default.

tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptionFactoryTest.java
mkouba: I don't think we should resolve the clashes. We don't do this for `ProcessAnnotatedType` and friends either.

PR #1897, opened 2019-01-31
Uncommented test reproducing WFLY-11125

nziakova:
Fix for WFLY-11125 has been merged to WF master, should be available in WF16. We might want to enable this test now.


manovotn:
retest this please

PR #1906, opened 2019-02-15
Fix for weld-2567

doychin:
As described in the bug report getCDI method should not return null. Instead it should throw IllegalStateException


manovotn:
ok to test

manovotn:
I am going to close this for now as we cannot merge before this gets accepted in CDI specification.
I'll like this PR to the relevant WELD issue so that it doesn't get lost.

PR #1908, opened 2019-03-05
WELD-2566 Enforce cache clearing for Request/Session/Conversation con…

manovotn:
…texts on WeldAlterableContext.clearAndSet().

### DIFF
--- a/impl/src/main/java/org/jboss/weld/contexts/unbound/RequestContextImpl.java
+++ b/impl/src/main/java/org/jboss/weld/contexts/unbound/RequestContextImpl.java
@@ -12,6 +12,7 @@ import java.util.HashSet;
 import java.util.Set;
 
 import org.jboss.weld.context.api.ContextualInstance;
+import org.jboss.weld.contexts.cache.RequestScopedCache;
 import org.jboss.weld.serialization.spi.BeanIdentifier;
 
 public class RequestContextImpl extends AbstractUnboundContext implements RequestContext {
@@ -50,6 +51,9 @@ public class RequestContextImpl extends AbstractUnboundContext implements Reques
     @Override
     public void clearAndSet(Collection<ContextualInstance<?>> setOfInstances) {
         getBeanStore().clear();
+        // invalidate caches for req., session, conv. scopes
+        // this might be needed for propagation on the thread where there are existing contexts
+        RequestScopedCache.invalidate();
         for (ContextualInstance<?> contextualInstance : setOfInstances) {
             getBeanStore().put(getId(contextualInstance.getContextual()), contextualInstance);
         }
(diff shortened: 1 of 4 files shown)


impl/src/main/java/org/jboss/weld/contexts/unbound/RequestContextImpl.java
```diff
@@ -50,6 +51,9 @@ public void deactivate() {
     @Override
     public void clearAndSet(Collection<ContextualInstance<?>> setOfInstances) {
         getBeanStore().clear();
+        // invalidate caches for req., session, conv. scopes
+        // this might be needed for propagation on the thread where there are existing contexts
+        RequestScopedCache.invalidate();
```
manovotn: To be fair, I am not sure this is really needed and it sure isn't tested within this PR because those Arq. tests will always use an `AbstractBoundContext` implementation (either `Bound` or pure `HTTP` backed based on protocol used).
Even in SE tests the `Bound` variants will be in place.

@mkouba any thoughts?

mkouba: Better safe than sorry ;-)

manovotn: Yea, that was my thinking

manovotn:
The JDK 8 job seems to have gotten stuck in the process, it didn't actually fail.
Retest this please

tests-arquillian/src/test/java/org/jboss/weld/tests/contexts/propagation/ContextPropagationTest.java
```diff
@@ -104,4 +111,22 @@ private void pingBeanAndOffloadTask(WeldManager manager, Class<? extends Abstrac
...
+        // use the bean in this thread first - set counter to two
+        AbstractBeanWithState bean = manager.instance().select(beanClazz).get();
+        Assert.assertEquals(0, bean.getValue());
+        bean.incrementCounter();
+        bean.incrementCounter();
+        Assert.assertEquals(2, bean.getValue());
+
+        // prepare a callable which will further increase the counter and return the value we found there
+        Callable<Integer> callableTask = () -> {
+            AbstractBeanWithState beanInCallable = CDI.current().select(beanClazz).get();
```
mkouba: I think you can safely replace `CDI.current().select(beanClazz).get()` with `bean`... but your version is perfectly ok too ;-)

manovotn: Yeah I could, that's really just a copy of the `pingBeanAndOffloadTask()` method...

PR #1909, opened 2019-03-05
 WELD-2564 Update tools for doc building, correct one ascii doc that …

manovotn:
…doesn't pass XSLT.


manovotn:
This is the minimum to make it build for me on JDK 8 and 11.
Note that CI results here mean nothing as we don't build docs there.

I still can't figure out why I cannot upgrade the `<pressgang.version>`, the exceptions are rather useless there.

manovotn:
Made some minor changes and removals.

PR #1911, opened 2019-03-28
WELD-2573 Make sure TCCL is preserved within async observer notificat…

manovotn:
@nziakova could you please run this down with SM enabled (in SE and EE).
Both, the actual fix and the test are juggling with TCCL which is likely SM-sensitive and might cause some troubles.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
+++ b/impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
@@ -332,6 +332,8 @@ public class ObserverNotifier {
         final NotificationMode mode = initModeOption(options.get(WeldNotificationOptions.MODE));
         final Long timeout = initTimeoutOption(options.get(WeldNotificationOptions.TIMEOUT));
         final Consumer<Runnable> securityContextActionConsumer = securityServices.getSecurityContextAssociator();
+        // grab current TCCL
+        ClassLoader tccl = SecurityActions.getContextClassLoader();
         final ObserverExceptionHandler exceptionHandler;
         CompletableFuture<U> completableFuture;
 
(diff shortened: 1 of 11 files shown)


impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -332,6 +334,8 @@ public RuntimeException apply(Type eventType) {
         final NotificationMode mode = initModeOption(options.get(WeldNotificationOptions.MODE));
         final Long timeout = initTimeoutOption(options.get(WeldNotificationOptions.TIMEOUT));
         final Consumer<Runnable> securityContextActionConsumer = securityServices.getSecurityContextAssociator();
+        // grab current TCCL
+        ClassLoader tccl = GetContextClassLoaderAction.INSTANCE.run();
```
mkouba: You can't use a `PrivilegedAction` like this. You should pass the instance as a param of an `AccessController.doPrivileged()` invocation. You could also create a package-private `SecurityActions` util class like https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/bean/proxy/SecurityActions.java and optimize the code if SM is not enabled.

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -427,19 +431,22 @@ private NotificationMode initModeOption(Object value) {
...
-    private <T, U extends T> Supplier<T> createSupplier(Consumer<Runnable> securityContextActionConsumer, U event, EventMetadata metadata, ObserverExceptionHandler exceptionHandler,
+    private <T, U extends T> Supplier<T> createSupplier(ClassLoader threadContextClassLoader, Consumer<Runnable> securityContextActionConsumer, U event, EventMetadata metadata, ObserverExceptionHandler exceptionHandler,
             boolean handleExceptions, Runnable notifyAction) {
         return () -> {
+            ClassLoader originalCl = GetContextClassLoaderAction.INSTANCE.run();
             final ThreadLocalStackReference<EventMetadata> stack = currentEventMetadata.pushIfNotNull(metadata);
             final RequestContext requestContext = requestContextHolder.get();
             securityContextActionConsumer.accept(() -> {
                 try {
+                    new SetContextClassLoaderAction(threadContextClassLoader).run();
```
mkouba: Again, this will not work with SM enabled.

manovotn:
I was worried this will make a fuss out of Weld SE without SM (where FJP is used by default), so i cobbled together additional SE test. Please review that one as well.

manovotn:
retest this please

PR #1912, opened 2019-03-28
WELD-2574 Make PrivateMethodHandler serializable.

manovotn:
Fix for WELD-2574 that I verified with the reproducer in linked issue.
I'll try to think of an automated test for this (requires to enforce serialization) before proceeding to merge.


manovotn:
Turns out `PassivationTest` can be easily altered to test this scenario so I did just that.

PR #1913, opened 2019-04-09
[WELD-2578] Permits DiscoveryStrategy instances to be loaded via ServiceLoader.

ljnelson:

This experimental PR allows `DiscoveryStrategy` instances to be loaded using a `ServiceLoader`.  All tests in `weld/core` continue to pass.  Any and all feedback welcome.  A [discussion in the Weld forums](https://developer.jboss.org/thread/279931) sparked this idea.

### DIFF
--- a/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
+++ b/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
@@ -63,7 +63,7 @@ public abstract class AbstractDiscoveryStrategy implements DiscoveryStrategy {
     private final List<BeanArchiveHandler> handlers;
 
     protected AbstractDiscoveryStrategy() {
-        this.handlers = new LinkedList<BeanArchiveHandler>();
+        handlers = new LinkedList<BeanArchiveHandler>();
     }
 
     /**
@@ -73,10 +73,10 @@ public abstract class AbstractDiscoveryStrategy implements DiscoveryStrategy {
      * @param initialBeanDefiningAnnotations
      */
     public AbstractDiscoveryStrategy(ResourceLoader resourceLoader, Bootstrap bootstrap, Set<Class<? extends Annotation>> initialBeanDefiningAnnotations) {
-        this();
-        this.setResourceLoader(resourceLoader);
-        this.setBootstrap(bootstrap);
-        this.setInitialBeanDefiningAnnotations(initialBeanDefiningAnnotations);
+        handlers = new LinkedList<BeanArchiveHandler>();
+        setResourceLoader(resourceLoader);
+        setBootstrap(bootstrap);
+        setInitialBeanDefiningAnnotations(initialBeanDefiningAnnotations);
     }
 
     @Override
@@ -94,14 +94,8 @@ public abstract class AbstractDiscoveryStrategy implements DiscoveryStrategy {
         this.scanner = scanner;
     }
 
-    public void setInitialBeanDefiningAnnotations(Set<? extends Class<? extends Annotation>> initialBeanDefiningAnnotations) {
-        if (initialBeanDefiningAnnotations == null) {
-            this.initialBeanDefiningAnnotations = null;
-        } else if (initialBeanDefiningAnnotations.isEmpty()) {
-            this.initialBeanDefiningAnnotations = Collections.emptySet();
-        } else {
-            this.initialBeanDefiningAnnotations = Collections.unmodifiableSet(new HashSet<>(initialBeanDefiningAnnotations));
-        }
+    public void setInitialBeanDefiningAnnotations(Set<Class<? extends Annotation>> initialBeanDefiningAnnotations) {
+        this.initialBeanDefiningAnnotations = initialBeanDefiningAnnotations;
     }
 
     @Override
--- a/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
+++ b/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
@@ -47,7 +47,8 @@ import org.jboss.weld.util.ServiceLoader;
  * @author Matej Briškár
  * @author Martin Kouba
  * @author Jozef Hartinger
- * @author <a href="https://about.me/lairdnelson" target="_parent">Laird Nelson</a>
+ * @author <a href="https://about.me/lairdnelson"
+ * target="_parent">Laird Nelson</a>
  */
 public abstract class AbstractDiscoveryStrategy implements DiscoveryStrategy {
 
@@ -78,10 +79,12 @@ public abstract class AbstractDiscoveryStrategy implements DiscoveryStrategy {
         this.setInitialBeanDefiningAnnotations(initialBeanDefiningAnnotations);
     }
 
+    @Override
     public void setResourceLoader(ResourceLoader resourceLoader) {
         this.resourceLoader = resourceLoader;
     }
 
+    @Override
     public void setBootstrap(Bootstrap bootstrap) {
         this.bootstrap = bootstrap;
     }
--- a/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
+++ b/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source
- * Copyright 2014, Red Hat, Inc. and/or its affiliates, and individual
+ * Copyright 2014-2019, Red Hat, Inc. and/or its affiliates, and individual
  * contributors by the @authors tag. See the copyright.txt in the
  * distribution for a full listing of individual contributors.
  *
@@ -47,18 +47,23 @@ import org.jboss.weld.util.ServiceLoader;
  * @author Matej Briškár
  * @author Martin Kouba
  * @author Jozef Hartinger
+ * @author <a href="https://about.me/lairdnelson" target="_parent">Laird Nelson</a>
  */
 public abstract class AbstractDiscoveryStrategy implements DiscoveryStrategy {
 
-    protected final ResourceLoader resourceLoader;
+    protected ResourceLoader resourceLoader;
 
-    protected final Bootstrap bootstrap;
+    protected Bootstrap bootstrap;
 
-    protected final Set<Class<? extends Annotation>> initialBeanDefiningAnnotations;
+    protected Set<Class<? extends Annotation>> initialBeanDefiningAnnotations;
 
     protected BeanArchiveScanner scanner;
 
     private final List<BeanArchiveHandler> handlers;
+
+    protected AbstractDiscoveryStrategy() {
+        this.handlers = new LinkedList<BeanArchiveHandler>();
+    }
 
     /**
      *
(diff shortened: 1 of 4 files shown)


ljnelson:
@manovotn I'd be curious to see what you think about this.

mkouba:
@ljnelson Do you really need to implement your own `DiscoveryStrategy`? I mean you can register a custom `org.jboss.weld.environment.deployment.discovery.BeanArchiveHandler` and disable Jandex using the `org.jboss.weld.discovery.disableJandexDiscovery` property.

manovotn:
@ljnelson can you elaborate on what's the gain and goal of having your own discovery strategy here?
I can understand not wanting to use Jandex in which case you fall back to reflection one but I don't see the benefits of a wholly different/custom strategy?

ljnelson:
I had one concrete use case and one fuzzy abstract one.

The concrete one is to expose the Jandex index to end user applications.

The fuzzy abstract one is to use [ClassGraph](https://github.com/classgraph/classgraph) for discovery instead of straight reflection when Jandex is not present.

Leaving aside whether I _should_ be doing what I'm doing, is my PR sound?

manovotn:
> Leaving aside whether I should be doing what I'm doing, is my PR sound?

It's not about whether you should or shouldn't. I was merely curious in what cases does this make any sense for SE. I know Weld docs talk about allowing integrators to implements this service (and they do), but that it mainly intended for EE integration.

So it's more about whether it's needed or whether we should look for another way.

> The concrete one is to expose the Jandex index to end user applications.

Hmm, if I am not mistaken then we build jandex index on a per archive basis and then put them together in a `CompositeIndex`. Is it this `CompositeIndex` you want to expose or are you building even more complex one that you then want to feed to Weld?

> The fuzzy abstract one is to use ClassGraph for discovery instead of straight reflection when Jandex is not present.

This is an interesting case, I can see the need to have the service loader mechanism in place for it although we've never had anyone coming in with that request so far and since you mention this as abstract one, I suppose you don't really care about it either :)

manovotn:
Also, if this were to pass (and I am not saying it would), then relying on the abstract class sounds very awkward. Furthermore `DiscoveryStrategy` isn't even in official API/SPI repository which it should be if users were to have the option to implement it. So that would imply changes and release in API and then changes in core. That seems like quite a lot of hassle for a corner case...

ljnelson:
> Is it this CompositeIndex you want to expose or are you building even more complex one that you then want to feed to Weld?

It's the `CompositeIndex`.

> relying on the abstract class sounds very awkward.

Without a doubt.  I didn't want to touch a `public` interface and add methods to it, potentially breaking implementations.

I do actually care about ClassGraph.  😁  It's just not holding me up right now.

> Furthermore DiscoveryStrategy isn't even in official API/SPI repository which it should be if users were to have the option to implement it.

`BeanArchiveHandler` is not in the API or SPI either, right?  But I can register one of those as a service, yes?  Isn't this a similar case?  Both interfaces are in the `common` repo.

manovotn:
> > Furthermore DiscoveryStrategy isn't even in official API/SPI repository which it should be if users were to have the option to implement it.
> 
> `BeanArchiveHandler` is not in the API or SPI either, right? But I can register one of those as a service, yes? Isn't this a similar case? Both interfaces are in the `common` repo.

It is also not advertised as a user-endpoint in the docs.

Having this public would require us to move it to API/SPI which actually breaks implementations and is not a material for a micro release. What we could do, is look at how to optionally make the built Jandex index public for Weld SE if that's sufficient.

ljnelson:
> Having this public would require us to move it to API/SPI which actually breaks implementations and is not a material for a micro release. 

Please don't take this the wrong way, but how is having an optional service loading step a breaking change?  I don't understand that part.

ljnelson:
> What we could do, is look at how to optionally make the built Jandex index public for Weld SE if that's sufficient.

Unfortunately, it is not, as it's not the fact that it's `private` that is the problem (I can always hack around that).  It is that it is (understandably, for most cases) `null`ed out before I can get to it.

manovotn:
> > Having this public would require us to move it to API/SPI which actually breaks implementations and is not a material for a micro release.
> 
> Please don't take this the wrong way, but how is having an optional service loading step a breaking change? I don't understand that part.

Because we would move the class to API artifact instead of env-common and possibly change the contract a bit so that it's more usable when it is designed to be publicly exposed.



> > What we could do, is look at how to optionally make the built Jandex index public for Weld SE if that's sufficient.
> 
> Unfortunately, it is not, as it's not the fact that it's `private` that is the problem (I can always hack around that). It is that it is (understandably, for most cases) `null`ed out before I can get to it.

It's `null`ed in the `ClassFileServices` but still present in the `JandexDiscoveryStrategy`. So if you are hacking around it, you might as well get it from there, it even has a getter ;-)

ljnelson:
> It's nulled in the ClassFileServices but still present in the JandexDiscoveryStrategy. So if you are hacking around it, you might as well get it from there, it even has a getter ;-)

Note that it is `null`ed out before I can get to it since `cleanupAfterBootstrap()` fires long before a portable extension is activated.

ljnelson:
> > Please don't take this the wrong way, but how is having an optional service loading step a breaking change? I don't understand that part.
> 
> Because we would move the class to API artifact instead of env-common

Is it therefore a bug that `BeanArchiveHandler` is not an API artifact?

manovotn:
> > > Please don't take this the wrong way, but how is having an optional service loading step a breaking change? I don't understand that part.
> > 
> > 
> > Because we would move the class to API artifact instead of env-common
> 
> Is it therefore a bug that `BeanArchiveHandler` is not an API artifact?

`BeanArchiveHandler` isn't even mentioned in Weld docs, it's not really a public API. I suppose it was needed for EE integrators at some point.

> > It's nulled in the ClassFileServices but still present in the JandexDiscoveryStrategy. So if you are hacking around it, you might as well get it from there, it even has a getter ;-)
> 
> Note that it is `null`ed out before I can get to it since `cleanupAfterBootstrap()` fires long before a portable extension is activated.

Yea, the services are, I meant grabbing the discovery strategy directly, but now that I think about it, you can't really do that.

manovotn:
Going over the options here, we don't want this in public API at this point and I don't like the usage of `AbstractDiscoveryStrategy` in this PR. OTOH I can see that you cannot easily grab the Jandex here or use ClassGraph which might be legit use case.

So if you could alter the PR, i.e. add the setter methods directly on the `DiscoveryStrategy` interface and base the service loading logic purely around that interface, I think we could come to an agreement. Obviously a JIRA issue would be neat. Does that sound alright @ljnelson ?

ljnelson:
Certainly.  I originally didn't alter the `DiscoveryStrategy` interface because, well, it's an interface.  But yes, if changing the contract of a `public` interface is OK (just want to make really sure) then yes, it makes things cleaner.  I'll alter the PR later today.  Thanks @manovotn.

mkouba:
@ljnelson It's an interface but it's not a part of the official API so +1 from me ;-)

manovotn:
I have created backing issue for this - https://issues.jboss.org/browse/WELD-2578
Once you implement the changes, also please change the PR title to contain JIRA issue link.

manovotn:
ok to test

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/DiscoveryStrategyFactory.java
```diff
@@ -42,20 +47,31 @@ private DiscoveryStrategyFactory() {
...
+            final DiscoveryStrategy candidate = iterator.next().getValue();
+            if (candidate != null) {
+                candidate.setResourceLoader(resourceLoader);
+                candidate.setBootstrap(bootstrap);
+                candidate.setInitialBeanDefiningAnnotations(initialBeanDefiningAnnotations);
+                returnValue = candidate;
             }
+        } else if (jandexStrategyDisabled) {
+            CommonLogger.LOG.jandexDiscoveryStrategyDisabled();
+        } else if (Jandex.isJandexAvailable(resourceLoader)) {
```
manovotn: You swapped the checks for `jandexStrategyDisabled` and `Jandex.isJandexAvailable(resourceLoader)`.
Please keep them in the original order as the logging is meant to happen only in case Jandex is available but deliberately not being used.

ljnelson: Oh!  My mistake!  I'll fix this shortly.

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
```diff
@@ -47,37 +47,63 @@
...
-        this.initialBeanDefiningAnnotations = initialBeanDefiningAnnotations;
     }
 
     @Override
     public void setScanner(BeanArchiveScanner scanner) {
         this.scanner = scanner;
     }
 
+    public void setInitialBeanDefiningAnnotations(Set<? extends Class<? extends Annotation>> initialBeanDefiningAnnotations) {
+        if (initialBeanDefiningAnnotations == null) {
```
mkouba: I think a simple setter is all we need because it's Weld who provides the set and it's already immutable.

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
```diff
@@ -47,37 +47,63 @@
...
+    }
+
     /**
      *
      * @param resourceLoader
      * @param bootstrap
      * @param initialBeanDefiningAnnotations
      */
     public AbstractDiscoveryStrategy(ResourceLoader resourceLoader, Bootstrap bootstrap, Set<Class<? extends Annotation>> initialBeanDefiningAnnotations) {
+        this();
```
mkouba: There is no need to call `this()` and use `this` keyword for the setters...

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/DiscoveryStrategyFactory.java
```diff
@@ -42,20 +47,33 @@ private DiscoveryStrategyFactory() {
      */
     public static DiscoveryStrategy create(ResourceLoader resourceLoader, Bootstrap bootstrap,
         Set<Class<? extends Annotation>> initialBeanDefiningAnnotations, boolean jandexStrategyDisabled) {
-        if (Jandex.isJandexAvailable(resourceLoader)) {
+        DiscoveryStrategy returnValue = null;
+        final Iterator<Metadata<DiscoveryStrategy>> iterator = ServiceLoader.load(DiscoveryStrategy.class, resourceLoader).iterator();
+        if (iterator != null && iterator.hasNext()) {
+            final DiscoveryStrategy candidate = iterator.next().getValue();
```
mkouba: It could be useful to log a DEBUG message that a service provider is used...?

manovotn: I like this idea, we should use the `org.jboss.weld.environment.logging.CommonLogger` for it.

manovotn:
@ljnelson please incorporate Martin's suggestions as well so we can merge this.
As for the debug message

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
ljnelson: OK; I'll remove `this` wherever possible and will duplicate the `handlers` assignment code.

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/DiscoveryStrategyFactory.java
ljnelson: OK; I'll invent a message and add it to `CommonLogger` and use it here.

ljnelson:
@manovotn @mkouba I've made the changes in https://github.com/weld/core/pull/1913/commits/29232725fb758bc85a560f963b9814c9c9f5ff39.

ljnelson:
All tests pass for me in `weld/core`.  If there's some other kind of error that is not accessible to me (I can't see the CI results pages), let me know what it is and I'll try to fix it.

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
mkouba: Ah sorry, my comment about `this()` was wrong. I thought it's `super()` ;-). For setters it's useless but for handlers assignment it's ok.

mkouba:
The build failed with ` RequestScopeEventMessageDeliveryTest.testEventsFired:69 expected:<1> but was:<0>` from `weld-core-test-arquillian` which seems to be unrelated.

mkouba:
retest this please

manovotn:
> All tests pass for me in `weld/core`. If there's some other kind of error that is not accessible to me (I can't see the CI results pages), let me know what it is and I'll try to fix it.

The CI job results are sadly not public, however, those checks named `Weld Publish CI` should be visible to all. Its simply a gist copy of the results to bypass the CI server limitations.
Either way the failure wasn't your making but likely a flaky test.

ljnelson:
Hi, @manovotn; let me know if there's anything else I need to do here; I'm assuming not.

PR #1914, opened 2019-04-15
WELD-2577 Skip setting WELD_S_HASH session attribute if bean index is empty

pferraro:
https://issues.jboss.org/browse/WELD-2577


manovotn:
ok to test

manovotn:
@pferraro have you tested this somehow?
The CI runs our tests, but Weld itself doesn't have clustering tests. I'm running some extra WFLY tests right now but that's about all I can do on Weld side. So if you know any other tests that should be executed with this, give me a shout.

pferraro:
@manovotn I've effectively validated this against WildFly's clustering testsuite.
This behavior is simulated here: https://github.com/wildfly/wildfly/pull/12241/commits/b214cc7cfc923a0efc2b3da78bc149bff244114f
Which is part of a larger PR for WildFly:
https://github.com/wildfly/wildfly/pull/12241

manovotn:
Ah, thanks for the links. Then I think we're good to go.

PR #1916, opened 2019-04-25
[WELD-2580] Ensure InjectionTarget#dispose(Object) method is called in ManagedBean implementation

ljnelson:
Fixes https://issues.jboss.org/browse/WELD-2580.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/ManagedBean.java
+++ b/impl/src/main/java/org/jboss/weld/bean/ManagedBean.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source
- * Copyright 2008, Red Hat, Inc., and individual contributors
+ * Copyright 2008-2019, Red Hat, Inc., and individual contributors
  * by the @authors tag. See the copyright.txt in the distribution for a
  * full listing of individual contributors.
  *
@@ -55,6 +55,8 @@ import org.jboss.weld.util.reflection.Reflections;
  * @author Marius Bogoevici
  * @author Ales Justin
  * @author Marko Luksa
+ * @author <a href="https://about.me/lairdnelson"
+ * target="_parent">Laird Nelson</a>
  */
 public class ManagedBean<T> extends AbstractClassBean<T> {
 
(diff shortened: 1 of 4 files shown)


manovotn:
ok to test

manovotn:
Theoretically, this should be alright.
After some poking around code I would say that any Producer that is actually backed by java class (as a bean) is based off `BasicInjectionTarget` which already overrides `dispose()` to no-op.
With this in mind, I haven't been able to think of a valid corner case that would be broken by this change. If anyone overrode IT at this point and declared some disposal method, then it has never been called and thus it's a bug in their code.

We'll also need a test for this (basically the content of issue reproducer made into an Arq. test).

manovotn:
@ljnelson I am fine incorporating this change in next release. Though I would like a test that verifies this functionality. I have no problems adding one but since its your PR I was thinking you maybe want to do that?

ljnelson:
@manovotn Yep; absolutely; just have to get to it.  😄

ljnelson:
Hi, @manovotn; not sure where to start here; I see tests all over the place.  Maybe add something here (https://github.com/weld/core/blob/master/tests-arquillian/src/test/java/org/jboss/weld/tests/injectionTarget/InjectionTargetTest.java)?

manovotn:
@ljnelson You were headed in right direction.
I'd create new package under https://github.com/weld/core/tree/master/tests-arquillian/src/test/java/org/jboss/weld/tests/injectionTarget, probably named `dispose`.

As for how to use the extension in a test, here is one example https://github.com/weld/core/blob/master/tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/qualifiers/SpiAddedQualifierTest.java#L35-L38

ljnelson:
Hi, @manovotn; I've added a test.  CI failed for some reason, but building `weld/core` via `mvn clean install` works for me.

manovotn:
@ljnelson the failure happens when running in container tests - that means against actual WFLY server.
Can be reproduced by doing `export JBOSS_HOME=/path/to/wfly` and running tests with `-Dincontainer` parameter

manovotn:
@ljnelson pushed a commit that should fix the test, let's see what CI has to say.
I think your original test only worked  incidentally because of how Arquillian operates outside of WFLY. Using `Instance` to obtain and destroy the bean should be a sure shot in all environments.

manovotn:
Yeah, that worked! :)

PR #1921, opened 2019-06-06
WELD-2557 Move bean deployment processing to validation phase to avoi…

manovotn:
…d race conditions during bootstrap.

This should address WELD-2557 and WFLY-11601 by reshuffling some part of bootstrap.


fanste:
I've just tested this fix against the same setup from WFLY-11601. A lot of restarts later and the error never appeared again.

This seems to be a working fix.

PR #1923, opened 2019-06-20
WELD-2583 INtercepted subclasses should not contain private methods w…

manovotn:
…ith pack-private parameters, add automated test.


manovotn:
As a  side note, I originally thought we should also log this case (WARN), but then I decided against it as this case wouldn't work before and even if you needed to subclass such class, then you couldn't invoke the method to trigger interception because it is private. Feel free to convince me otherwise.

PR #1926, opened 2019-07-01
Issues Weld-2587 OWB-1293 Jetty #3804

gregw:
Proposed fix for https://issues.jboss.org/browse/WELD-2587,  https://issues.apache.org/jira/browse/OWB-1293 and https://github.com/eclipse/jetty.project/issues/3804

Implement a new style of integration with the jetty container that does not require access to jetty
private APIs (as enabled in the jetty cdi module).

Previous mechanism has been trimmed of deprecated support and retained as a legacy container that will
continue to work for users that do expose the jetty APIs (as enabled in the jetty cdi2 module).

See https://github.com/eclipse/jetty.project/pull/3838 for example usage

Signed-off-by: Greg Wilkins <gregw@webtide.com>


gregw:
@manovotn What do you think of this approach?  I have separated the approaches into `JettyContainer` and `JettyLegacyContainer`.   It could be done as a single container that supports both approaches, but have to be careful that the class loading of dependent classes for the legacy mode is not triggered by default.

manovotn:
I am on a PTO this week, I'll look into it next week.
If you need feedback faster, we can try and get hold of @mkouba.
In the meantime, I'll make sure CI runs against this PR.

manovotn:
ok to test

environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/LegacyWeldDecorator.java
```diff
@@ -0,0 +1,57 @@
...
+ * to the webapp:<ul>
+ *     <li>org.eclipse.jetty.server.handler.ContextHandler</li>
+ *     <li>org.eclipse.jetty.servlet.ServletContextHandler</li>
+ *     <li>org.eclipse.jetty.util.DecoratedObjectFactory</li>
+ *     <li>org.eclipse.jetty.util.Decorator</li>
+ * </ul>
+ *
+ * @author <a href="mailto:ales.justin@jboss.org">Ales Justin</a>
+ */
+public class LegacyWeldDecorator extends WeldDecorator implements Decorator {
```
manovotn: I am thinking whether it's worth to keep legacy stuff around or whether we should just say that next Weld version supports Jetty 9.4.20+. Currently our docs mention 9.4.14 but we have done similar micro/minor version bumps previously within Weld's micro releases (Weld 3.1.0 -> 3.1.1 changed Jetty support from 9.3.6 to 9.4.14). Personally, I would lean towards dropping the legacies...
WDYT @mkouba?

environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/JettyContainer.java
```diff
@@ -21,75 +21,45 @@
...
-        if (p < 0) {
-            return false;
-        }
-        String version = si.substring(p + 1);
-        String[] split = version.split("\\.");
-        int major = parseVersion(split[0]);
-        int minor = parseVersion(split[1]);
-        return (major > MAJOR_VERSION || (major == MAJOR_VERSION & minor >= MINOR_VERSION));
+        // The jetty cdi module from 9.4.20 sets this attribute to indicate that a DecoratingListener is registered.
+        return "DecoratingListener".equals(sc.getAttribute("org.eclipse.jetty.cdi"));
```
manovotn: I'd make both string constants.

environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/JettyContainer.java
```diff
@@ -21,75 +21,45 @@
...
-
-    private static final int MAJOR_VERSION = 7;
-    private static final int MINOR_VERSION = 2;
-
     protected String classToCheck() {
-        // This is not used anyway - server implementation classes are hidden from the web application in Jetty by default
-        // Instead ServletContext.getServerInfo() is parsed
-        return JETTY_REQUIRED_CLASS_NAME;
+        // Never called because touch is overridden below.
+        throw new UnsupportedOperationException();
```
manovotn: I know this isn't supposed to be invoked and I am nitpicking, but should that ever accidentally happen, I'd prefer to either see dummy string returned or an exception with explanation.

environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/LegacyWeldDecorator.java
manovotn: Obviously we could only do that once the new Jetty supporting this approach gets released.

gregw: Yeh, this will only work on jetty-9.4.20, which gives a pretty hard change for Weld releases on a point release.  A lot of users might want to upgrade Weld prior to upgrading Jetty, so having legacy support (perhaps deprecated) for a while would lessen the impact on users.   Note also it is kind of useful for checking the signature of the non legacy decorator !

gregw:
Note that we are having a similar conversation with OWB [OWB-1293](https://issues.apache.org/jira/browse/OWB-1293) and the outcome of that may affect the integration here.   We'd like the same integration for all CDI contains... ideally moving towards an SPI.

gregw:
The suggestion from OWB is that Jetty provides the decorator that uses standard CDI SPIs to decorate objects.  I think this can be done with:
```java
public <T> T decorate(T o) {
  BeanManager manager = CDI.current().getBeanManager();
  manager.createInjectionTarget(manager.createAnnotatedType((Class<T>)o.getClass()))
    .inject(o,manager.createCreationalContext(null));
  return o;
}
```

Unless told otherwise, I'm going to try for that integration.... stand by...

gregw:
@manovotn I have updated this to the approach suggested by OWB and now implemented in https://github.com/eclipse/jetty.project/pull/3838 since https://github.com/eclipse/jetty.project/pull/3838/commits/ce10f6bcf4de08f413497668427d4d2b18c461b3

Jetty now has a CdiServletContextInitializer that looks in the webapp for CDI classes. If found, it registers a CdiDecorator that uses MethodHandles to invoke the standard CDI SPI to decorate Listeners, Filters and Servlets.
This means that the JettyContainer in Weld need not do anything to arrange to be decorated.

My only concern with this approach is that there is no ordering between SCIs, so I have written the CdiDecorator with lazy instantiation, so it should work even if the Weld SCI is invoked before or after.

What do you think of this approach?

manovotn:
Hmm, I preferred the previous solution more, what exactly was wrong with that? Current solution, while making it seemingly easy for CDI implementations, takes away any control we have over how injection is done. Besides, you have do to a lot of class loading and method handle magic on Jetty-side this way while having it on CDI impl side doesn't require these. Would it even work with JDK 12+ in regards to CL restrictions?

Furthermore, I am a bit concerned about `CDI.current()`. Its fidgety construct that is caller sensitive and might cause trouble during things like concurrent bootstrap. The BeanManager you get that way can also differ based on where was the invocation (in what bean archive), but you hopefully aren't affected by that since you only need to create injection target.

manovotn:
After discussing this with @mkouba we both agreed that the previous approach (where CDI-impl owns the decorator impl) would be better.
We don't have 'hard' arguments but the previous solution:
* Allows CDI impl to customize the injection target if needed
  * For instance now we create the `InjectionTarget` through Weld specific builder and we disable resource injection (there was issue for that specifically, WELD-1919) and disable class callbacks
  * See the code [here](https://github.com/weld/core/blob/master/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/inject/AbstractInjector.java#L47-L50), note that you cannot do this with CDI API only
* Generally makes the CDI-related code reside in CDI-impl code bases
  * This gives us chance to adapt to any possible change needed and there is a fair chance there are CDI people around who understand why things are done specifically some way
* Jetty doesn't need to implement cumbersome class loading and reflection based invocation of methods which is generally error-prone and hard to manage code
  * All the more when this can be done cleanly on CDI-impl side

manovotn:
@gregw BTW next Weld version is due early Aug, do you think Jetty will have an agreed upon solution by then or should we postpone this issue to next release?

gregw:
@manovotn Still trying to convince the OWB folks of an approach.  But we have a 9.4.20 release due this week, so I think I will revert to the first approach I showed you and try to get that included in the release.  I will work on that tomorrow.

gregw:
@manovotn I think we will just have to go with two integration mechanism in jetty.    I have reverted to have the DecoratingListener, but now it is in a `decorate` module, as it is independent of CDI.  I've updated this PR to match.

manovotn:
Thanks, I added a comment to their issue as well, although I am afraid that won't change their minds in an instant.
I wonder why they so stubbornly resist the option a). It seems to be superior and even a way to introduce portability in this regard. Well, let's see what they have to say.

gregw:
I have updated this PR to support all 3 integration mechanisms: legacy, DecoratingListener of CdiDecorator.  The integration tests in https://github.com/eclipse/jetty.project/pull/3838 now test all 3 mechanism as from 323024c22a6a96995d8666fcf00928eaaeab8692

Note sure how to release this, as these tests rely on a SNAPSHOT build of weld.  Might have to revert to a released version and disable tests until this PR is available in a release.

gregw:
Oh I have just realised in https://issues.apache.org/jira/browse/OWB-1293 that the CdiDecorator approach needs to be stateful and keep a map like the one in org.jboss.weld.environment.servlet.inject.AbstractInjector.   But the CdiDecorator lives on the server classpath, so such a map is a big memory leak!  Unless we can put the decorator in the webapp classloader, so it is disposed with the webapp, I think that is the end of the CdiDecorating approach.

manovotn:
I am getting a bit lost here.
Why do we need both integration ways for new Jetty version?
Even if Jetty supports two ways (which I still dislike but that's your decision to make), then all Weld needs is legacy + one new way. And I think we both agreed that the one where Weld owns the integration code looked better, right?
Or has something changed in that regard?

I've glanced the OWB discussions, but it looks like you are stuck making OWB work with anything at all so I don't see any link between that and the changes in this PR.

gregw:
@manovotn it's not so much that Weld needs three integrations, this PR could get by with just 2 (legacy and the DecoratingListener).   However, it looks like Jetty will also support the CdiDecorator integration for OWB... and if that is applied to Weld it works fine, but if Weld doesn't know about it then it gives a warning that injection will not work  when it is in fact working.    So I thought it best to at least make Weld aware of the CdiDecorator approach so that if it is detected it can do nothing other than produce a reasonable info message.

manovotn:
@gregw  but if Jetty supports both ways and prefers to invoke CDI SPI on its own(at least that's what I understood from the commits here), then `WeldDecorator` will never be invoked anyway. That makes it kind of useless, right?

I find it weird that we are strongly tilting towards the solution OWB wanted to push forward and that is with both of us thinking its worse approach and even with OWB having none of those two solutions working at all(according to OWB-1293)...

gregw:
@manovotn I definitely would prefer a single CDI integration mechanism.  But the need to support legacy integration means that we have to support at least 2 mechanism, so by the 0,1,infinity principle then why not support 3 :)    

But ultimately I'm not experienced enough in CDI to adjudicate between the two proposed integration mechanisms as I see goodness in them both:
- The CdiDecorator approach is attractive because it does use the CDI SPI and should be portable over implementations - but it's moderately complex reflection code that will suffer from jakarta renaming complexity.  It also needs to signal to the CDI implementation it is in use to stop warnings or attempts at other integration.  It also has the other issue that the CDI implementation itself cannot influence the creation of the injector and finally there is the problem of ordering of SCIs which also can make this approach a bit less predictable.
- The DecoratingListener approach is a bit simpler and more general purpose as it allows for arbitrary decoration.  However, while it is a step in a direction that might eventually be supported by the servlet specification, it is not currently a standard approach an may never be.

Currently my PR has 3 modules:
- **cdi2**, the legacy module that uses exposed APIs.  This is deprecated and will eventually be removed.
- **cdi**, the CdiDecorator approach that calls the CDI SPI.
- **decorate**, the DecorationListener approach that can be used by more than CDI.

I can see arguments for both **cdi** and **decorate** modules to exist in Jetty.  If there is an SPI, then I guess we should provide a mechanism to invoke it... but then having a general purpose decoration mechanism can also be justified and can be more flexible.   So now, even if Weld and OWB agreed on a single integration approach, I'd be kind of sad to see the other one go.  But I don't have a preference for the **cdi** module and personally would choose **decorate** if I needed to deploy, simply because I trust the CDI implementers to do a better job of setting up CDI than the CdiDecorator. 

Flip side is that if jetty does support all 3 mechanisms, then there is no compulsion for Weld to do the same.  I'm happy to convert my PR to just support **cdi** or **decorate** modules, but then it's not too much code to support them both.

manovotn:
> But the need to support legacy integration means that we have to support at least 2 mechanism, so by the 0,1,infinity principle then why not support 3 :)

That's different mindset though; at least to me. Legacy is something you add to keep supporting the prior approach but you aim to eventually rule it out which would mean you are left with one. This way, if you ever need to make another API change, you are going to have to handle two legacy mechanisms.

>  but then it's not too much code to support them both.

My point is that if there are two, then Jetty will inevitably try one of them first and then the secondary approach becomes a dead code that never gets executed.
For instance in the current code within this PR, `WeldDecorator` is more or less useless because all it does is check for context attribute and leave it at that. Why have that code in the first place?

Two side notes for the current code in this PR (regardless of what we agree on as solution):
* You are still creating `new JettyWeldInjector` and storing that into context even though it never gets used since Jetty's CdiDecorator will kick in instead
* The two log messages in `JettyLogger` should be merged in one since they have identical text. You can add code comments to say that given part is handled by either weld code or jetty code

gregw:
I was only joking about the 0,1,infinity thing... there was a smiley.

I expect it will be the CDI implementations in their documentation that will drive users to use one approach or the other.  So jetty is not going to drive one mechanism or the other.  You are completely right that Weld doesn't need to support both mechanisms - you can definitely just use DecoratingListener as your integration and document how to set that up. I expect the the vast majority of your users will follow your instructions and all will be good.

However, it looks likely that jetty will also have the CdiDecorator mechanism to support OWB and a small number of users may find that and try to use it with Weld.  It will work, but without the code I'm proposing it will give an warning saying that it wont.  Perhaps that's OK, but it is not much code to produce a reasonable message?  Either way, I don't see `WeldDecorator` as useless.  A worst it has a static method that determines if it is applicable or not, a best it is the preferred integration recommended by your documentation and will be used by most users.

I will update the PR to avoid creating the injector unless it is needed.
The text of the two messages is not identical and I do like being told exactly what integration mechanism was discovered....  but I'll merge to one if you want.

manovotn:
> I was only joking about the 0,1,infinity thing... there was a smiley.

Sorry, I am switching between projects and went all too serious missing the joke altogether. My fault :)

> However, it looks likely that jetty will also have the CdiDecorator mechanism to support OWB and a small number of users may find that and try to use it with Weld. 

So, I am probably missing bits here. How exactly will people enable CdiDecorator in Jetty?

Because to enable current integration we have in Weld (and future one using our decorator), you need to add a `weld-servlet` artifact which triggers the integration code. So if there is someone who would want to try Jetty's CdiDecorator with Weld, they wouldn't really need our artifacts, would they? And the logging you added would then only kick in if they tried to leverage both ways at the same time?

> The text of the two messages is not identical and I do like being told exactly what integration mechanism was discovered

Oops, you are right, in that case we either keep it as is or make it even more detailed, e.g. something like `You have enabled CdiDecorator in Jetty via XY, Weld's integration will be ignored.`, or vice versa depending on which way is prioritized if both are detected.

Overall, I am not against detecting the other integration and logging it. Do you think it wold make sense for Jetty to prioritize custom decorators (such WeldDecorator) over the built-in one (CdiDecorator) if it can detect both? Because, well, it is general approach versus a custom one that can go more in depth.

gregw:
The various mechanisms for CDI integration are enabled in jetty by modules.   So if weld wants to document how to use the decorate module you can say in your document to use the command:
```
java -jar $JETTY_HOME/start.jar --add-to-start=decorate
```

In jetty itself, when you list our modules you will see descriptions of them.  If you want just cdi related modules you can do:
```
java -jar $JETTY_HOME/start.jar --list-modules=cdi
```
for which the result will be:
```
2139] java -jar start.jar --list-modules=cdi

Available Modules:
==================
tags: [cdi]

Modules for tag 'cdi':
----------------------

     Module: cdi 
           : CDI SPI integration for CDI inside the webapp.
           : This module does not provide CDI, but configures jetty to look for the CDI SPI within
           : a webapp.  If the CDI SPI is found, then a CdiDecorator will be registered to
           : decorate Listeners, Filters and Servlets using the standard CDI SPI.
           : The module indicates to the webapp that this mechanism is available by setting the
           : "org.eclipse.jetty.cdi" context attribute to "CdiDecorator".
       Tags: cdi
     Depend: deploy
        LIB: lib/jetty-cdi-${jetty.version}.jar
        LIB: lib/apache-jsp/org.mortbay.jasper.apache-el-*.jar

     Module: cdi2 
           : Deprecated support for CDI integrations inside the webapp.
           : This module does not provide CDI, but configures jetty so that a CDI implementation
           : can enable itself as a decorator for Filters, Servlets and Listeners.
           : This modules uses the deprecated technique of exposing private Jetty decorate APIs to the CDI
           : implementation in the webapp.
       Tags: cdi
     Depend: deploy
        LIB: lib/apache-jsp/org.mortbay.jasper.apache-el-*.jar
        XML: etc/cdi/jetty-cdi2.xml

     Module: decorate 
           : Jetty setup to support Decoration of Listeners, Filters and Servlets within a deployed
           : webapp (as used by some CDI integrations).
           : This module uses the DecoratingListener to register an object set as a context attribute
           : as a dynamic decorator. This module sets the "org.eclipse.jetty.webapp.DecoratingListener"
           : context attribute with the name of the context attribute that will be listened to.
           : By default the attribute is "org.eclipse.jetty.decorator".
       Tags: cdi
     Depend: deploy
        XML: etc/jetty-decorate.xml
```

By default, Jetty will have no CDI related module enabled, so users will have to select one or the other.  I expect it will be the CDI documentation that tells them which one (and then you can get rid of any documentation that describes how to directly manipulate the classpath visible to the webapp).

Note also that I'm currently asking if the `cdi` module should be renamed `cdi-spi`

My preference would be to favour custom integrations over the CDI SPI, but I'm not sure how to do that without being fragile.  If the cdi module is turned on it will see the CDI SPI in a webapp, but I'm not sure how to make it see that an integration is there and active.  Perhaps the CDI SPI module needs a way for the webapp to turn it off?

gregw:
I'm testing the CdiDecorator approach against Weld 3.1.1.FINAL  I would have expected it to work with perhaps a wrong message.  However it fails with:
```
2019-07-31 10:14:42.520:INFO::main: Logging initialized @631ms to org.eclipse.jetty.util.log.StdErrLog
2019-07-31 10:14:42.529:INFO::main: Logging initialized @641ms to org.eclipse.jetty.util.log.StdErrLog
2019-07-31 10:14:42.722:WARN:oejx.XmlConfiguration:main: Deprecated method public void org.eclipse.jetty.server.HttpConfiguration.setBlockingTimeout(long) in file:///usr/local/home/gregw/src/jetty-9.4.x/tests/test-distribution/target/jetty_home_12839819278412769992/jetty-distribution-9.4.20-SNAPSHOT/etc/jetty.xml
2019-07-31 10:14:42.817:INFO:oejs.Server:main: jetty-9.4.20-SNAPSHOT; built: 2019-07-31T00:05:24.261Z; git: f58b00bff8b528b5edb3385419d0d81f2c0fdbf0; jvm 11.0.4+10-LTS
2019-07-31 10:14:42.834:INFO:oejdp.ScanningAppProvider:main: Deployment monitor [file:///usr/local/home/gregw/src/jetty-9.4.x/tests/test-distribution/target/tests/test-bases/jetty_base_10763367958342193000/webapps/] at interval 1
2019-07-31 10:14:43.326:INFO:oeja.AnnotationConfiguration:main: Scanning elapsed time=225ms
2019-07-31 10:14:43.404:INFO:oejc.CdiServletContainerInitializer:main: CdiDecorator enabled in ServletContext@o.e.j.w.WebAppContext@6440112d{Weld CDI Integration Test WebApp,/demo,file:///usr/local/home/gregw/src/jetty-9.4.x/tests/test-distribution/target/tests/test-bases/jetty_base_10763367958342193000/webapps/demo/,STARTING}{/usr/local/home/gregw/src/jetty-9.4.x/tests/test-distribution/target/tests/test-bases/jetty_base_10763367958342193000/webapps/demo}
Jul. 31, 2019 10:14:43 AM org.jboss.weld.environment.servlet.EnhancedListener onStartup
INFO: WELD-ENV-001008: Initialize Weld using ServletContainerInitializer
Jul. 31, 2019 10:14:43 AM org.jboss.weld.bootstrap.WeldStartup <clinit>
INFO: WELD-000900: 3.1.1 (Final)
Jul. 31, 2019 10:14:43 AM org.jboss.weld.bootstrap.WeldStartup startContainer
INFO: WELD-000101: Transactional services not available. Injection of @Inject UserTransaction not available. Transactional observers will be invoked synchronously.
2019-07-31 10:14:43.994:WARN:oejw.WebAppContext:main: Failed startup of context o.e.j.w.WebAppContext@6440112d{Weld CDI Integration Test WebApp,/demo,file:///usr/local/home/gregw/src/jetty-9.4.x/tests/test-distribution/target/tests/test-bases/jetty_base_10763367958342193000/webapps/demo/,UNAVAILABLE}{/usr/local/home/gregw/src/jetty-9.4.x/tests/test-distribution/target/tests/test-bases/jetty_base_10763367958342193000/webapps/demo}
java.lang.NoClassDefFoundError: org/eclipse/jetty/util/Decorator
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1016)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:550)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:458)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:452)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:451)
	at org.eclipse.jetty.webapp.WebAppClassLoader.foundClass(WebAppClassLoader.java:648)
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadAsResource(WebAppClassLoader.java:621)
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:534)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	at org.jboss.weld.environment.jetty.JettyContainer.initialize(JettyContainer.java:74)
	at org.jboss.weld.environment.servlet.WeldServletLifecycle.initialize(WeldServletLifecycle.java:213)
	at org.jboss.weld.environment.servlet.EnhancedListener.onStartup(EnhancedListener.java:62)
	at org.eclipse.jetty.plus.annotation.ContainerInitializer.callStartup(ContainerInitializer.java:145)
	at org.eclipse.jetty.annotations.ServletContainerInitializersStarter.doStart(ServletContainerInitializersStarter.java:64)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.servlet.ServletContextHandler.startContext(ServletContextHandler.java:342)
	at org.eclipse.jetty.webapp.WebAppContext.startWebapp(WebAppContext.java:1443)
	at org.eclipse.jetty.webapp.WebAppContext.startContext(WebAppContext.java:1407)
	at org.eclipse.jetty.server.handler.ContextHandler.doStart(ContextHandler.java:818)
	at org.eclipse.jetty.servlet.ServletContextHandler.doStart(ServletContextHandler.java:276)
	at org.eclipse.jetty.webapp.WebAppContext.doStart(WebAppContext.java:524)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.deploy.bindings.StandardStarter.processBinding(StandardStarter.java:46)
	at org.eclipse.jetty.deploy.AppLifeCycle.runBindings(AppLifeCycle.java:188)
	at org.eclipse.jetty.deploy.DeploymentManager.requestAppGoal(DeploymentManager.java:513)
	at org.eclipse.jetty.deploy.DeploymentManager.addApp(DeploymentManager.java:154)
	at org.eclipse.jetty.deploy.providers.ScanningAppProvider.fileAdded(ScanningAppProvider.java:172)
	at org.eclipse.jetty.deploy.providers.WebAppProvider.fileAdded(WebAppProvider.java:422)
	at org.eclipse.jetty.deploy.providers.ScanningAppProvider$1.fileAdded(ScanningAppProvider.java:66)
	at org.eclipse.jetty.util.Scanner.reportAddition(Scanner.java:622)
	at org.eclipse.jetty.util.Scanner.reportDifferences(Scanner.java:538)
	at org.eclipse.jetty.util.Scanner.scan(Scanner.java:402)
	at org.eclipse.jetty.util.Scanner.doStart(Scanner.java:318)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.deploy.providers.ScanningAppProvider.doStart(ScanningAppProvider.java:145)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.deploy.DeploymentManager.startAppProvider(DeploymentManager.java:599)
	at org.eclipse.jetty.deploy.DeploymentManager.doStart(DeploymentManager.java:249)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:167)
	at org.eclipse.jetty.server.Server.start(Server.java:407)
	at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:119)
	at org.eclipse.jetty.server.handler.AbstractHandler.doStart(AbstractHandler.java:106)
	at org.eclipse.jetty.server.Server.doStart(Server.java:371)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.xml.XmlConfiguration.lambda$main$0(XmlConfiguration.java:1824)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at org.eclipse.jetty.xml.XmlConfiguration.main(XmlConfiguration.java:1773)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.eclipse.jetty.start.Main.invokeMain(Main.java:218)
	at org.eclipse.jetty.start.Main.start(Main.java:491)
	at org.eclipse.jetty.start.Main.main(Main.java:77)
Caused by: 
java.lang.ClassNotFoundException: org.eclipse.jetty.util.Decorator
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:565)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1016)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:550)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:458)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:452)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:451)
	at org.eclipse.jetty.webapp.WebAppClassLoader.foundClass(WebAppClassLoader.java:648)
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadAsResource(WebAppClassLoader.java:621)
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:534)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	at org.jboss.weld.environment.jetty.JettyContainer.initialize(JettyContainer.java:74)
	at org.jboss.weld.environment.servlet.WeldServletLifecycle.initialize(WeldServletLifecycle.java:213)
	at org.jboss.weld.environment.servlet.EnhancedListener.onStartup(EnhancedListener.java:62)
	at org.eclipse.jetty.plus.annotation.ContainerInitializer.callStartup(ContainerInitializer.java:145)
	at org.eclipse.jetty.annotations.ServletContainerInitializersStarter.doStart(ServletContainerInitializersStarter.java:64)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.servlet.ServletContextHandler.startContext(ServletContextHandler.java:342)
	at org.eclipse.jetty.webapp.WebAppContext.startWebapp(WebAppContext.java:1443)
	at org.eclipse.jetty.webapp.WebAppContext.startContext(WebAppContext.java:1407)
	at org.eclipse.jetty.server.handler.ContextHandler.doStart(ContextHandler.java:818)
	at org.eclipse.jetty.servlet.ServletContextHandler.doStart(ServletContextHandler.java:276)
	at org.eclipse.jetty.webapp.WebAppContext.doStart(WebAppContext.java:524)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.deploy.bindings.StandardStarter.processBinding(StandardStarter.java:46)
	at org.eclipse.jetty.deploy.AppLifeCycle.runBindings(AppLifeCycle.java:188)
	at org.eclipse.jetty.deploy.DeploymentManager.requestAppGoal(DeploymentManager.java:513)
	at org.eclipse.jetty.deploy.DeploymentManager.addApp(DeploymentManager.java:154)
	at org.eclipse.jetty.deploy.providers.ScanningAppProvider.fileAdded(ScanningAppProvider.java:172)
	at org.eclipse.jetty.deploy.providers.WebAppProvider.fileAdded(WebAppProvider.java:422)
	at org.eclipse.jetty.deploy.providers.ScanningAppProvider$1.fileAdded(ScanningAppProvider.java:66)
	at org.eclipse.jetty.util.Scanner.reportAddition(Scanner.java:622)
	at org.eclipse.jetty.util.Scanner.reportDifferences(Scanner.java:538)
	at org.eclipse.jetty.util.Scanner.scan(Scanner.java:402)
	at org.eclipse.jetty.util.Scanner.doStart(Scanner.java:318)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.deploy.providers.ScanningAppProvider.doStart(ScanningAppProvider.java:145)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.deploy.DeploymentManager.startAppProvider(DeploymentManager.java:599)
	at org.eclipse.jetty.deploy.DeploymentManager.doStart(DeploymentManager.java:249)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.util.component.ContainerLifeCycle.start(ContainerLifeCycle.java:167)
	at org.eclipse.jetty.server.Server.start(Server.java:407)
	at org.eclipse.jetty.util.component.ContainerLifeCycle.doStart(ContainerLifeCycle.java:119)
	at org.eclipse.jetty.server.handler.AbstractHandler.doStart(AbstractHandler.java:106)
	at org.eclipse.jetty.server.Server.doStart(Server.java:371)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:72)
	at org.eclipse.jetty.xml.XmlConfiguration.lambda$main$0(XmlConfiguration.java:1824)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at org.eclipse.jetty.xml.XmlConfiguration.main(XmlConfiguration.java:1773)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.eclipse.jetty.start.Main.invokeMain(Main.java:218)
	at org.eclipse.jetty.start.Main.start(Main.java:491)
	at org.eclipse.jetty.start.Main.main(Main.java:77)
2019-07-31 10:14:44.023:INFO:oejs.AbstractConnector:main: Started ServerConnector@3d037a30{HTTP/1.1,[http/1.1]}{0.0.0.0:36121}
2019-07-31 10:14:44.023:INFO:oejs.Server:main: Started @2135ms
```

I don't understand why it is trying to call `org.jboss.weld.environment.jetty.JettyContainer.initialize`  if the `org.eclipse.jetty.util.Decorator` is not visible? Surely it should have failed the touch and not tried the Jetty container?

manovotn:
> Surely it should have failed the touch and not tried the Jetty container?/

If you are running on 3.1.1Final, then the `touch()` method just parses versions and it gets all info from provided `ContainerContext`, no reason to fail there. Class loading is lazy and the invocation of static `WeldDecorator.process()` is the first time that class is accessed.

I mean, the CDI module inside jetty doing integration cannot work with older Weld as we rely on that deprecated/removed interface class, right?

> Note also that I'm currently asking if the cdi module should be renamed cdi-spi

It might make sense to rename `cdi` to `cdi-spi` and `cdi2` to `cdi-legacy` perhaps?

> My preference would be to favour custom integrations over the CDI SPI, but I'm not sure how to do that without being fragile

Hmm, from Weld's perspective, users will still need to have `weld-servlet.jar` on CP to make the integration work. Simply because that's where the CDI container gets bootstrapped.
That's where we attempt to push forward our decorator; we could just check the context for a parameter indicating Jetty's decorator is applied and consider that an erroneous setting and blow up...
 But that would completely cut off Weld from using the other integration means. Then again, from Jetty perspective it also seems "wrong" to activate two modules to achieve one goal when they are mutually exclusive (for CDI integration purpose).

Or, you could add an additional method to the Decorator interface (and hence to all implementing classes) that would indicate what integration it does. First thing coming to mind is `public String decorationType()` and CDI integrations would simple return `"CDI"` string. You could then check known decorators for that and give them preference. Obviously, you could also have an enum or contants with those values (to avoid fragility) but that would once again imply the need to depend on Jetty API.

manovotn:
Just an FYI - I will have to merge this by Mon in order to proceed with Weld release so we need to put it into some final state by then.

gregw:
@manovotn I'm pretty happy with this PR as is... other than perhaps documentation... I'll try to add some now.  So merge by Monday is good.
I am renaming `cdi` module to `cdi-spi` but will leave `cdi2` as is because it is mentioned in documentation.  I'll make sure it emits a deprecated warning and we will eventually remove.

So the only think I want to get working now is to be able to run the `cdi-spi` on `3.1.1Final`... and that should just be changes in jetty obviously.  My understanding of your code is that you have a bunch of integration you can try in `WeldServletLifeCycle`:
```java
                container = checkContainers(ctx, dump,
                        Arrays.asList(TomcatContainer.INSTANCE, JettyContainer.INSTANCE, UndertowContainer.INSTANCE, GwtDevHostedModeContainer.INSTANCE));
```
and that it will call touch on each of those until it doesn't get a failure and then will use that.   So if we do not expose our `org.eclipse.jetty.util.Decorator` class on the classpath, then the touch call should fail and the JettyContainer initialize should never be called and thus we should not get the class cast exception.  No integration should trigger, a warning should be logged, but then we can still work with the CdiDecorator calling the SPI.    Obviously I'm missing something???

gregw:
Ah I see! 3.1.1.Final JettyContainer.touch doesn't check for a class, it parses the server info from serlvetContext!
So to test this with the CdiDecorator, I'll have to create a NoopContainer!

gregw:
Ah never mind... I've tested CdiDecorator with 3.1.2-SNAPSHOT.  There is no need to make it work with 3.1.1.Final.  I think this PR is good.... at least ready for a re-review.

manovotn:
> Ah I see! 3.1.1.Final JettyContainer.touch doesn't check for a class, it parses the server info from serlvetContext!

Yeah, that's what I was trying to tell you ;)

> I think this PR is good.... at least ready for a re-review.

Agreed, the PR looks good.

Only one bit remains and that is documentation.
E.g. Weld docs have a section on Jetty that is now pretty outdated -> https://github.com/weld/core/blob/master/docs/reference/src/main/asciidoc/environments.asciidoc#jetty
We'll have to:
* change versions that we support (9.4.20 and the new 10.x) 
* remove the `Class Loading` section which should now be obsolete
* replace the above with few sentences on how to enable integration now

Should I do that or will you add it into this PR?

gregw:
@manovotn updated doco

PR #1928, opened 2019-07-22
WELD-2585 Synchronize creation of conversation ID generator to avoid …

manovotn:
…race conditions.


manovotn:
Theoretical solution to WELD-2585, requires confirmation/testing.
Adding `hold` label for that reason.

manovotn:
Despite not having the confirmation on this fix, I've decided to proceed and include it in the upcoming release. If we find out it is insufficient, we can do further tweaks for future releases.

PR #1933, opened 2019-08-08
WELD-2597 Rename the attributes for Jetty integration #2587

gregw:
For https://issues.jboss.org/browse/WELD-2597

Sorry about this, but the changes for jetty-9.4.20 have had a tough time in review because of the two mechanism and somewhat confused names.
Thus we are considering a bit of a renaming... but in a way that will not break Weld-3.1.2, which will still work with either `cdi2` or `decorate` modules.
However, going forward, we'd like to have `cdi-decorate` or `cdi-spi` modules.

This PR reflects the proposed rename in https://github.com/eclipse/jetty.project/pull/3946.    It also includes support for continuing to use the `decorate` directly, but you only really need that if you are worried about people applying 3.1.2 documentation to 3.1.3 and beyond.

Note that this is still a draft proposal until 3946 is reviewed and we've gotten your feedback as well.

Signed-off-by: Greg Wilkins <gregw@webtide.com>

### DIFF
--- a/docs/reference/src/main/asciidoc/environments.asciidoc
+++ b/docs/reference/src/main/asciidoc/environments.asciidoc
@@ -176,7 +176,7 @@ jetty modules listed below needs to be enabled.
 
 ===== Jetty `cdi-decorate` Module
 
-Since Jetty-9.4.20 and Weld 3.1.2, the Weld/Jetty integration uses the jetty `cdi-decorate` module.
+Since Jetty-9.4.20 and Weld 3.1.2.Final, the Weld/Jetty integration uses the jetty `cdi-decorate` module.
 To activate this module in jetty the argument `--module=cdi-decorate` needs to be added to the
 command line, which can be done for a standard distribution by running the commands:
 
--- a/docs/reference/src/main/asciidoc/environments.asciidoc
+++ b/docs/reference/src/main/asciidoc/environments.asciidoc
@@ -174,15 +174,15 @@ No further configuration is needed when starting Jetty as an embedded webapp ser
 However, if you’re using a Jetty standalone instance one more configuration step is required, one of the
 jetty modules listed below needs to be enabled.
 
-===== Jetty `decorate` Module
+===== Jetty `cdi-decorate` Module
 
-Since Jetty-9.4.20 and Weld 3.1.2, the Weld/Jetty integration uses the jetty `decorate` module.
-To activate this module in jetty the argument `--module=decorate` needs to be added to the
+Since Jetty-9.4.20 and Weld 3.1.2, the Weld/Jetty integration uses the jetty `cdi-decorate` module.
+To activate this module in jetty the argument `--module=cdi-decorate` needs to be added to the
 command line, which can be done for a standard distribution by running the commands:
 
 -------------------------
 cd $JETTY_BASE
-java -jar $JETTY_HOME/start.jar --add-to-start=decorate
+java -jar $JETTY_HOME/start.jar --add-to-start=cdi-decorate
 -------------------------
 
 ===== Jetty `cdi2` Module
--- a/environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/JettyContainer.java
+++ b/environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/JettyContainer.java
@@ -85,7 +85,7 @@ public class JettyContainer extends AbstractJettyContainer {
                     // Initialize a JettyWeldInjector and create WeldDecorator for it
                     super.initialize(context);
                     servletContext.setAttribute(CDI_DECORATING_LISTENER_ATTRIBUTE, new WeldDecorator(servletContext));
-                    JettyLogger.LOG.jettyDecorationIsSupported();
+                    JettyLogger.LOG.jettyCdiDecorationIsSupported();
                     break;
 
                 case DECORATING_LISTENER_MODE:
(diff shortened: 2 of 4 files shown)


docs/reference/src/main/asciidoc/environments.asciidoc
```diff
@@ -174,15 +174,15 @@ No further configuration is needed when starting Jetty as an embedded webapp ser
 However, if you’re using a Jetty standalone instance one more configuration step is required, one of the
 jetty modules listed below needs to be enabled.
 
-===== Jetty `decorate` Module
+===== Jetty `cdi-decorate` Module
 
-Since Jetty-9.4.20 and Weld 3.1.2, the Weld/Jetty integration uses the jetty `decorate` module.
-To activate this module in jetty the argument `--module=decorate` needs to be added to the
+Since Jetty-9.4.20 and Weld 3.1.2, the Weld/Jetty integration uses the jetty `cdi-decorate` module.
```
manovotn: Maybe we should rather say `Weld 3.1.2.Final and newer` to indicate the doc is valid even for future versions until stated otherwise.

environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/JettyContainer.java
```diff
@@ -55,24 +57,49 @@ protected String classToCheck() {
...
+                    // This mode is only needed to match the Weld-3.1.2 documentation.
+                    // Initialize a JettyWeldInjector and create WeldDecorator for it
+                    super.initialize(context);
+                    String attribute = (String) servletContext.getAttribute(DECORATING_LISTENER_ATTRIBUTE);
+                    servletContext.setAttribute(attribute, new WeldDecorator(servletContext));
+                    JettyLogger.LOG.jettyDecorationIsSupported();
+                    break;
+
+                default:
+                    throw new IllegalStateException("unknown mode: " + mode);
```
manovotn: Let's be a bit more verbose here. 
And while at it, place this inside the `JettyLogger` as well please (here is [another logger](https://github.com/weld/core/blob/master/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/logging/WeldServletLogger.java#L105) as an example)

environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/JettyContainer.java
```diff
@@ -55,24 +57,49 @@ protected String classToCheck() {
...
+                    // Initialize a JettyWeldInjector and create WeldDecorator for it
+                    super.initialize(context);
+                    servletContext.setAttribute(CDI_DECORATING_LISTENER_ATTRIBUTE, new WeldDecorator(servletContext));
+                    JettyLogger.LOG.jettyDecorationIsSupported();
+                    break;
+
+                case DECORATING_LISTENER_MODE:
+                    // For use with the decorate module
+                    // This mode is only needed to match the Weld-3.1.2 documentation.
+                    // Initialize a JettyWeldInjector and create WeldDecorator for it
```
manovotn: I am kinda lost in what's the actual difference in `DECORATING_LISTENER_MODE` and `CDI_DECORATING_LISTENER_MODE` and why we need both?
Cause you elaborate a bit please?

manovotn:
whitelist

manovotn:
ok to test

docs/reference/src/main/asciidoc/environments.asciidoc
gregw: I can add the `Final` but I can't word smith anything that works well with `and newer` as there are too many `and`s
```
For Jetty-9.4.20 and newer and Weld 3.1.2.Final and newer...
```
I think the `Since` covers the `and newer`:

```
Since Jetty-9.4.20 and Weld 3.1.2.Final...
```

environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/JettyContainer.java
gregw: The `DECORATING_LISTENER_MODE` is not needed other than that is the style of usage that we documented in the Weld-3.1.2 release, so by supporting it we have backwards compatibility with the integration introduced in 3.1.2.  Essentially it is an identical mechanism to `CDI_DECORATING_LISTENER_MODE` but uses the unrenamed attributes.

Going forward `CDI_DECORATING_LISTENER_MODE` is the preferred integration mode.

manovotn: Alright, but I still don't see the backward compat problem - all users do is use Weld servlet dependency and then on Jetty side they activate a module. They don't care what attributes are used underneath.
So you either end up with Weld 3.1.2 and new Jetty - in which case it depends whether Jetty will recognize that.
Or you will have this new attribute in Weld 3.1.3 in which case it works as expected on Jetty side.

I am probably missing some obvious scenario in which it would blow up...

gregw: In Jetty we now have the raw `DecoratingListener` mechanism that can be instantiated in a number of ways and has a particular default setup.   We also have the `CdiDecoratingListener` mechanism which specializes `DecoratingListener` and changes the defaults to be more suitable for CDI.

In Weld-3.1.2 there were instructions on how to use the raw `DecoratingListener`, which if followed will configure jetty in a specific way that will be detected by the `DECORATING_LISTENER_MODE`.      

With this PR, we change those instructions to configure jetty to use the `CdiDecoratingListener` and that matching the `CDI_DECORATING_LISTENER_MODE`.

So the backward compatibility problem is only that Weld-3.1.2 documentation might be used with later releases.  It's not a big deal, but might save a little bit of confusion.   Perhaps we should deprecate the `DECORATING_LISTENER_MODE` and produce a warning if it is used?

manovotn: Alright, thanks for explanation.

manovotn:
retest this please

PR #1934, opened 2019-09-03
WELD-2595 Allow for private final methods on intercepted beans.

manovotn:
Note that `InterceptionModelInitializer` throws this exception on two places.
Second one is inside `initMethodDeclaredEjbInterceptors()` method but that one searches for method-level annotations in which case having the method private final is an error.


manovotn:
In order for this pass, we first need https://github.com/weld/core/pull/1936 to be merged, then this can be rebased.

PR #1938, opened 2019-09-11
Permits extending the set of bean-defining annotations in SE by using…

ljnelson:
… SeContainerInitializer#property(String, Object)

### DIFF
--- a/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
+++ b/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
@@ -177,6 +177,14 @@ import org.jboss.weld.util.collections.WeldCollections;
 public class Weld extends SeContainerInitializer implements ContainerInstanceFactory {
 
     /**
+     * By default, the set of bean-defining annotations is fixed. If set to a {@link Set} of annotation classes, the set of bean-defining annotations is
+     * augmented with the contents of the {@link Set}.
+     * <p>
+     * This key can be used through {@link #property(String, Object}.
+     */
+    public static final String ADDITIONAL_BEAN_DEFINING_ANNOTATIONS_PROPERTY = "org.jboss.weld.se.additionalBeanDefiningAnnotations";
+
+    /**
      * By default, bean archive isolation is enabled. If set to false, Weld will use a "flat" deployment structure - all bean classes share the same bean
      * archive and all beans.xml descriptors are automatically merged into one.
      * <p>
@@ -621,10 +629,20 @@ public class Weld extends SeContainerInitializer implements ContainerInstanceFac
      * @see #ARCHIVE_ISOLATION_SYSTEM_PROPERTY
      * @see #SHUTDOWN_HOOK_SYSTEM_PROPERTY
      * @see #DEV_MODE_SYSTEM_PROPERTY
+     * @see #ADDITIONAL_BEAN_DEFINING_ANNOTATIONS_PROPERTY
      * @see ConfigurationKey
      */
     public Weld property(String key, Object value) {
         properties.put(key, value);
+        if (ADDITIONAL_BEAN_DEFINING_ANNOTATIONS_PROPERTY.equals(key) && value instanceof Collection) {
+            for (Object element : ((Collection<?>)value)) {
+                if (element instanceof Class && Annotation.class.isAssignableFrom((Class<?>) element)) {
+                    @SuppressWarnings("unchecked")
+                    final Class<? extends Annotation> annotation = (Class<? extends Annotation>) element;
+                    extendedBeanDefiningAnnotations.add(annotation);
+                }
+            }
+        }
         return this;
     }
 


ljnelson:
This little PR makes it so you can (effectively) call `Weld#addBeanDefiningAnnotations()` via the `SeContainerInitializer#addProperty(String, Object)` method.

I'm sure I'm not including everything that would need to go into this PR but I wanted to see if this was something that might be useful.

manovotn:
Hi @ljnelson 

Weld SE already allows this - see [`Weld.addBeanDefiningAnnotations(..)`](https://github.com/weld/core/blob/master/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java#L893-L907).
And it should even be documented - https://docs.jboss.org/weld/reference/latest/en-US/html_single/#_extending_bean_defining_annotations

manovotn:
I'll close this PR as I think the above is what you were after, if that's not the case, feel free to reopen or just continue the conversation :)

ljnelson:
I'm very much aware of that.  It would be convenient to allow the property via `SeContainerInitializer` so no casting is necessary.

manovotn:
I see.
Well, in that case I suppose it makes sense, reopening.

@ljnelson can you please create JIRA issue for this?

manovotn:
ok to test

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -621,10 +629,20 @@ public Weld addAlternativeStereotype(Class<? extends Annotation> alternativeSter
      * @see #ARCHIVE_ISOLATION_SYSTEM_PROPERTY
      * @see #SHUTDOWN_HOOK_SYSTEM_PROPERTY
      * @see #DEV_MODE_SYSTEM_PROPERTY
+     * @see #ADDITIONAL_BEAN_DEFINING_ANNOTATIONS_PROPERTY
      * @see ConfigurationKey
      */
     public Weld property(String key, Object value) {
         properties.put(key, value);
+        if (ADDITIONAL_BEAN_DEFINING_ANNOTATIONS_PROPERTY.equals(key) && value instanceof Collection) {
```
manovotn: No other properties are parsed here, so I'd stick with that and move this logic over to the `initialize()` method.
You're gonna have to reshuffle it there a bit (or I can if you don't mind me adding to this PR) - parse properties sooner then `createDeployment()` is invoked (which already needs complete set of bean defining annotations) and also don't add this property to `configurationBuilder`

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -621,10 +629,20 @@ public Weld addAlternativeStereotype(Class<? extends Annotation> alternativeSter
...
+            for (Object element : ((Collection<?>)value)) {
+                if (element instanceof Class && Annotation.class.isAssignableFrom((Class<?>) element)) {
```
manovotn: I think it would make sense to blow up in an `else` block if these conditions aren't met - fail fast.

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -176,6 +176,14 @@
 @Vetoed
 public class Weld extends SeContainerInitializer implements ContainerInstanceFactory {
 
+    /**
+     * By default, the set of bean-defining annotations is fixed. If set to a {@link Set} of annotation classes, the set of bean-defining annotations is
+     * augmented with the contents of the {@link Set}.
+     * <p>
+     * This key can be used through {@link #property(String, Object}.
+     */
+    public static final String ADDITIONAL_BEAN_DEFINING_ANNOTATIONS_PROPERTY = "org.jboss.weld.se.additionalBeanDefiningAnnotations";
```
manovotn: Might be worth adding this in documentation as well.

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -176,6 +176,14 @@
 @Vetoed
 public class Weld extends SeContainerInitializer implements ContainerInstanceFactory {
 
+    /**
+     * By default, the set of bean-defining annotations is fixed. If set to a {@link Set} of annotation classes, the set of bean-defining annotations is
+     * augmented with the contents of the {@link Set}.
```
manovotn: Perhaps we should copy over the limitations mentioned on the other method, as those will still apply.
See this - https://github.com/weld/core/blob/master/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java#L896-L897

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
ljnelson: I went back and forth on this.  If I supply "`cheese`" for one of the System properties in this class that is expecting `true` or `false`, nothing blows up.  I followed suit here.  WDYT?

manovotn: Hmm, good point.
OTOH checking true/false is much simpler and you always have a default value. In this case you can have three scenarios - all correct, all wrong, a corrupted list where some are good and some are bad. And you have no fallback values, do you think it makes sense to silently proceed?

Or we could add some INFO logging for all wrong values that user tried to register perhaps?

manovotn:
Hi @ljnelson, are you planning to pick up the this issue and address the comments or had the interest dropped and should we just close it?

ljnelson:
Sorry, I have a lot of things going on.  😄  I'm still interested, if you could kindly leave this open.

manovotn:
Fine by me.

manovotn:
This PR has been hanging stale for quite some time. We should either conclude it somehow, or drop it.
@ljnelson are you planning to pick it up, or should I?

ljnelson:
That's fair.  Honestly, I'm not going to get around to it any time soon.  I still think it's a nice-to-have.

manovotn:
Fair enough, I agree it's nice to have, I'll add it to my TODO list. Thanks for fast response.

manovotn:
Superseeded by https://github.com/weld/core/pull/2167

PR #1939, opened 2019-10-15
WELD-2600 Fixing issue about spec-descriptor-property-replacement and beans.xml

rhuan080:
Fixing the issue described in https://issues.jboss.org/browse/WELD-2600


manovotn:
ok to test

manovotn:
@rhuan080 hello and thanks for the PR.
Would it be possible to also add an automated test for this? We're housing WFLY tests under [tests-arquillian](https://github.com/weld/core/tree/master/tests-arquillian) so you can take a look there for how its done.

rhuan080:
Hi @manovotn I will see how can I configure spec-descriptor-property-replacement using arquilian. I have not seen any test doing it.

manovotn:
Hmm, the WFLY Arq. adapter would probably have to allow you to do that.
I'd be curious to see if it can do that, but if not, then I guess we'll have to go without a test.
Let me know if you figured something out.

PR #1943, opened 2019-11-13
WELD-2598 Move to Jakarta EE artifacts.

manovotn:
Depends on https://github.com/weld/api/pull/89

### DIFF
--- a/environments/servlet/pom.xml
+++ b/environments/servlet/pom.xml
@@ -26,13 +26,14 @@
     </description>
 
     <properties>
-        <jsf.version>2.3</jsf.version>
+        <jsf.version>2.3.2</jsf.version>
         <tomcat.version>9.0.29</tomcat.version>
         <jetty.version>9.4.14.v20181114</jetty.version>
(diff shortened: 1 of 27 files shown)


manovotn:
Doesn't yet contain changes needed for osgi and for examples (which aren't tested on PRs).
Will add those soon as I am (more) sure the core works as it should.

manovotn:
Added bits for examples - they are now buildable but I haven't tried them all out yet.

environments/servlet/pom.xml
```diff
@@ -152,6 +153,13 @@
                 <version>${project.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>jakarta.websocket</groupId>
+                <artifactId>jakarta.websocket-api</artifactId>
```
manovotn: I don't know why we need this dependency now, it must have been part of other API and then separated?

manovotn:
Weld API was released, updated API artifacts to 3.1.SP2.

manovotn:
Found one last glitch, with that in place, I hope we didn't miss anything :)

PR #1948, opened 2019-11-27
[WELD-2603] Permits alternate ClientProxyProvider implementations

ljnelson:
Addresses [WELD-2603](https://issues.jboss.org/browse/WELD-2603) by permitting alternate [`ClientProxyProvider`](https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyProvider.java) implementations.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyProvider.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyProvider.java
@@ -17,6 +17,7 @@
 package org.jboss.weld.bean.proxy;
 
 import java.lang.reflect.Type;
+import java.util.Iterator;
 import java.util.Set;
 import java.util.function.Function;
 
@@ -25,11 +26,14 @@ import javax.enterprise.inject.spi.Bean;
 import org.jboss.weld.Container;
 import org.jboss.weld.bean.RIBean;
 import org.jboss.weld.bootstrap.api.ServiceRegistry;
+import org.jboss.weld.bootstrap.spi.Metadata;
 import org.jboss.weld.logging.BeanLogger;
+import org.jboss.weld.resources.spi.ResourceLoader;
 import org.jboss.weld.serialization.spi.BeanIdentifier;
 import org.jboss.weld.serialization.spi.ContextualStore;
 import org.jboss.weld.util.Proxies;
 import org.jboss.weld.util.Proxies.TypeInfo;
+import org.jboss.weld.util.ServiceLoader;
 import org.jboss.weld.util.cache.ComputingCache;
 import org.jboss.weld.util.cache.ComputingCacheBuilder;
 import org.jboss.weld.util.collections.ImmutableSet;
@@ -39,6 +43,8 @@ import org.jboss.weld.util.reflection.Reflections;
  * A proxy pool for holding scope adaptors (client proxies)
  *
  * @author Nicklas Karlsson
+ * @author <a href="https://about.me/lairdnelson"
+ * target="_parent">Laird Nelson</a>
  * @see org.jboss.weld.bean.proxy.ProxyMethodHandler
  */
 public class ClientProxyProvider {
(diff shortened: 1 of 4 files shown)


manovotn:
ok to test

manovotn:
I'll give this a look within next days as I first need to finish something else (read: "poke me if I don't add a comment soon").

impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyProvider.java
```diff
@@ -258,4 +301,49 @@ public void clear() {
...
+     * ClientProxyProvider} will return a {@link String} {@linkplain
+     * String#equals(Object) equal to} the supplied {@code
+     * containerId}.</p>
+     *
+     * @param resourceLoader a {@link ResourceLoader} that will be
+     * used by a {@link ServiceLoader}; may be {@code null}
+     *
+     * @param containerId the identifier of the {@link Container} on
+     * behalf of which the new {@link ClientProxyProvider} is being
+     * built; may be {@code null} but behavior of the returned {@link
```
manovotn: I don't think we should allow `null` here.

impl/src/test/resources/ClientProxyProviderTest/META-INF/services/org.jboss.weld.bean.proxy.ClientProxyProvider
```diff
@@ -0,0 +1 @@
+org.jboss.weld.tests.unit.proxy.ClientProxyProviderTest$FancyClientProxyProvider
```
manovotn: Isn't this going to interfere with other tests? As in, won't all other unit tests here have this proxy provider now?
Might be better to put this in some of the tests using arquillian as you can control whats on CP and also declare this asset programmatically for just that one test.

ljnelson: I don’t think so. My intent was for this to apply only to the current unit test.  What did I miss?

impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyProvider.java
```diff
@@ -153,31 +159,67 @@ public boolean equals(Object obj) {
     private final ComputingCache<Bean<Object>, Object> beanTypeClosureProxyPool;
     private final ComputingCache<RequestedTypeHolder, Object> requestedTypeClosureProxyPool;
 
-    private final String contextId;
+    private String containerId;
```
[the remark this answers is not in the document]
ljnelson: I’ll revert this if you like. It is called `containerId` If I remember right in whatever calls `newRootManager` so I thought I’d normalize the naming.

impl/src/main/java/org/jboss/weld/bean/proxy/ClientProxyProvider.java
ljnelson: OK; I can add a `null` check. (I was just documenting usage that was already there.)

ljnelson:
As I mentioned in the associated JIRA issue, I have in mind decoration concerns: keeping track of how many contextual references (not instances) are produced, and when, etc. etc. etc.  Being able to replace, customize or decorate the containing `BeanManager` would be even more useful.  At any rate, this was one of the few objects in the bootstrap sequence that was _not_ loaded from services or properties (even `ProxyInstantiator` is loaded from a system property).

manovotn:
To be fair, I'd rather see the current impl improved than having it as a part of SPI.

> As I mentioned in the associated JIRA issue, I have in mind decoration concerns: keeping track of how many contextual references (not instances) are produced, and when, etc. etc. etc

This sounds like something we could bake right into current implementation and just activate based on a property?

> Being able to replace, customize or decorate the containing BeanManager would be even more useful

Not sure I follow here. `ClientProxyProvider` doesn't hold reference to BM. Besides, proxy provider is supposed to create shared proxies app-wide so I don't really see how changing it would benefit you even if you could do that? 

>  (even ProxyInstantiator is loaded from a system property).
Yes, that's because of a config option to instantiate them via unsafe, hence bypassing no-args requirements if I recall correctly. Something that end user can choose to use without providing any custom impls of SPI parts.

ljnelson:
> > Being able to replace, customize or decorate the containing BeanManager would be even more useful
> 
> Not sure I follow here. ClientProxyProvider doesn't hold reference to BM.

Right.  I meant: rather than supplying a `ClientProxyProvider` I'd be just as happy if I could decorate `BeanManager` instead.

**I am, at root, interested in having a "hook" every time `BeanManager#getReference()` is called.**  `ClientProxyProvider` just seemed like a good place to jump in, since [`getReference()` always calls it](https://github.com/weld/core/blob/d8a8cb64f3a9751dcac07c8d9eda0dd3a0e682b8/impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java#L679-L696).

I am rarely a believer in setting properties to do one narrow thing.  I'm rather a believer in allowing for an implementation of something to be decorated or replaced.

> Besides, proxy provider is supposed to create shared proxies app-wide so I don't really see how changing it would benefit you even if you could do that?

As noted, `BeanManagerImpl#getReference()` always calls `ClientProxyProvider`.

manovotn:
You could have your own `BeanManager` bean (with a qualifier) that will basically delegate to our BM and do your things on top of that.
However, this BM won't replace the default one, so Weld calls won't go through it - it would only help you if you are interested in tracking your own calls through the BM.

ljnelson:
That's right.  I want to track the default `BeanManager`'s `getReference` call.  Among others.  If I could decorate the default `BeanManager` I'd be OK.  I thought that I might be able to do this from `CDIProvider.getCDI().getBeanManager()`—i.e. I thought that perhaps the default `BeanManager` used by Weld came from the `CDIProvider` (makes sense, right?)—but then realized that it is hard-coded to be `BeanManagerImpl`.

ljnelson:
@manovotn Any further thoughts here?

manovotn:
In all honesty, the more I think about it, the more I am against this PR.
You are actually after something different and only (ab)using this as a hook into the `BM.getReference()` call which doesn't justify making `ClientProxyProvider` a public SPI.

ljnelson:
OK.  I'll try to put a different PR together that somehow permits decoration of the bean manager I guess.

PR #1951, opened 2019-12-03
change to trigger CI build

manovotn:
Not to be merged, just a trigger for CI.
We are running on new images and this tests if they work correctly.


manovotn:
Looks like new images work!

PR #1954, opened 2019-12-07
Close weld.properties stream in finally block

benjamin-confino:
This PR closes the InputStream for the weld.properties file in a finally block after the file is read. This will help avoid integration problems when weld is integrated into servers that expect files to be closed quickly after use.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/config/WeldConfiguration.java
+++ b/impl/src/main/java/org/jboss/weld/config/WeldConfiguration.java
@@ -17,6 +17,7 @@
 package org.jboss.weld.config;
 
 import java.io.File;
+import java.io.InputStream;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


manovotn:
We do not plan to release more Weld 3.0 releases, using Weld 3.1 is the way to go.
For that reason I don't see much value in merging this on top of 3.0 branch; IMO it would only bring confusion. Feel free to disagree and argue :)

benjamin-confino:
I guess if something unexpected comes up that requires another 3.0 release it might be nice to have this sitting in the branch to ride alongside it. But I'm happy to follow your lead on this decision. I'm sure we'll be shipping 3.1 once it's available. I checked for a 3.1 branch and did not find one, so I presume the PR into master covers it? 

Thanks for the merge :)

PR #1955, opened 2019-12-07
Close weld.properties stream in finally block

benjamin-confino:
This PR closes the InputStream for the weld.properties file in a finally block after the file is read. This will help avoid integration problems when weld is integrated into servers that expect files to be closed quickly after use.


manovotn:
Thanks for contributing @benjamin-confino!

benjamin-confino:
You're very welcome

PR #1956, opened 2019-12-13
WIP: Initial work to port javax apis to jakarta package. 

starksm64:
Initial commit only goes far enough to allow atinject tck to run and pass with jakarta package.


manovotn:
Before we go into this, I'd like to see how the consuming EE server are planning to adapt to it (e.g. what WFLY is going to do), I've already sent an email to Brian asking about just that.
I strongly suspect nobody is in favor of fully ditching the old APIs and only supporting new package names, so we might be forced to look into some hybrid approach instead.

Based on that, we will also see about changing the major version, if we were to support both approaches, we might not need a new major (new majors usually create major problems for sustaining streams in products such as EAP and we then need to keep several versions alive).

Also, first set of changes will be needed in Weld API. Based on the Travis failure I can see you did those but didn't file a PR.

starksm64:
There is a PR for Weld API here:
https://github.com/weld/api/pull/91

starksm64:
Closing this as I have a new version that has the tests passing.

PR #1962, opened 2020-01-31
WELD-2611 Fire @Initialized event for conversation context after fail…

manovotn:
…ed attempt to recover non-existing conversation.


manovotn:
The failure is an SE test that uses Shrinkwrap to manually resolve Maven artifacts (all other tests pass).

This probably uses an old dep version as it tried to access Maven central via `http` but these days it only accepts `https`.
I will take a look at that on Mon, we will prob need to update the chain of Arq dependencies.

manovotn:
retest this please

manovotn:
retest this please

manovotn:
retest this please

PR #1968, opened 2020-02-24

manovotn:
…e same time.


manovotn:
We'll probably need to manually run the tests here, the CI certainly wasn't setup to run against this old version, so just disregard those results ;-)

manovotn:
I locally run Weld integration tests and CDI TCK against latest EAP 6 with patched weld bits containing this fix and that passed.

PR #1969, opened 2020-02-24

manovotn:
…e same time.

@mkouba I think we will probably want to add this even to master regardless of whether the deadlock can exist (even though I would like to test it). It's still one less lock we are holding at the given time. Agree?

### DIFF
--- a/impl/src/main/java/org/jboss/weld/contexts/AbstractConversationContext.java
+++ b/impl/src/main/java/org/jboss/weld/contexts/AbstractConversationContext.java
@@ -297,15 +297,26 @@ public abstract class AbstractConversationContext<R, S> extends AbstractBoundCon
     private void cleanUpConversationMap() {
         ManagedConversation currentConversation = getCurrentConversation();
         Map<String, ManagedConversation> conversations = getConversationMap();
+        Map<String, ManagedConversation> toClear = new HashMap<>();
+        S session = getSessionFromRequest(getRequest(), false);
+        // while synced, extract a map of conversations that we'll need to clean up, already removing them from map
         synchronized (conversations) {
             Iterator<Entry<String, ManagedConversation>> entryIterator = conversations.entrySet().iterator();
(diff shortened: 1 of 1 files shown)


impl/src/main/java/org/jboss/weld/contexts/AbstractConversationContext.java
```diff
@@ -297,17 +297,28 @@ public void deactivate() {
...
+                    toClear.put(entry.getKey(), entry.getValue());
                     entryIterator.remove();
                 }
             }
         }
+        // let the lock go, now destroy the conversation contexts, this will trigger locking on session
+        Iterator<Entry<String, ManagedConversation>> toClearIterator = toClear.entrySet().iterator();
+        while (toClearIterator.hasNext()) {
+            Entry<String, ManagedConversation> entry = toClearIterator.next();
+            if (entry.getValue().isTransient()) {
```
mkouba: Speaking of corner cases - so in theory you could mark a conversation as long-running (`transient=false`) after it has been removed from the session map but before it's destroyed... and boom :D. At least the spec does not seem to forbid this scenario. Anyway, the whole conversation scope seems to be underspecified and tricky to implement. So +1 for best effort.

manovotn: Hmm, the destruction is done after the request, so I am not quite sure how you would promote the conversation to long-running at that point.

mkouba: Well, it does not have to be the current conversation, ie. the conversation associated with the current request and you can even invoke `begin()/end()` upon an injected  `javax.enterprise.context.Conversation` too. Also note that the cleanup happens after we unlock the current conversation so there is still a possibility of a timing issue although it's very unlikely. But again it's very hypothetical so let's ignore my comments for now ;-).

PR #1970, opened 2020-02-24

manovotn:
…e same time.

Similarly to master branch, I think we could add this to 2.x branch as well.


manovotn:
Failing because this branch needs a similar fix as was done for master here - https://github.com/weld/core/commit/061257517d020d375b583a79238b5b1c808ee4c9

PR #1975, opened 2020-03-05
WELD-2616 Do not require specific OSGi versions of Jakarta APIs

Pandrex247:
The update to Weld API 3.1.SP2 seemed to mess up the OSGi versioning for the javax packages - for example it was requiring an import of EJB 2.x rather than 3.2. 

This change simply makes it so the version is undefined by the bundle artefact, letting it take anything.


manovotn:
ok to test

manovotn:
@Pandrex247 thanks for the PR.
Not being much of an OSGi user myself, what exactly does `version=!;` mean?

Pandrex247:
It's essentially a wildcard - it'll accept anything. 
It appears that without this it grabs the versioning for what the javax packages should be from the _jboss-x-api_ artefacts which don't necessarily match the versioning of the Java / Jakarta API they're wrapping (for example `jboss-ejb-api_**3.2**_spec artefact` is _2.0.0.Final_). 

Why this changed between _weld-api_ 3.1.SP1 & 3.1.SP2 I'm not sure - looking at the commits the only major thing in there is the jakarta switch, but they're still exporting javax packages so... ¯\\_(ツ)_/¯

The alternative to this would be to manually specify the versions it should accept, e.g. `javax.ejb; version="[3.2,4)"; resolution:=optional`

manovotn:
> https://github.com/jboss/jboss-jakarta-ejb-api_spec/blob/2.0.0.Final/api/pom.xml...

They dont match the versioning, but internally, they should use the new version of original artifact.
E.g. the EJB 2.0.0.Final is based on spec 3.2 - https://github.com/jboss/jboss-jakarta-ejb-api_spec/blob/2.0.0.Final/api/pom.xml

> major thing in there is the jakarta switch, but they're still exporting javax packages so.

Yea, I think this is in some way causing it, but I don't know the gory details.
However, a wildcard should do the trick here.

manovotn:
Thanks for the PR @Pandrex247!

PR #1977, opened 2020-03-28
WELD-2609 Fix OM configurator when reading from an AnnotatedMethod.

manovotn:
Issue only impacts reading from `AnnotatedMethod`, if you use `ObserverMethod`, then it works just fine.


manovotn:
@mkouba FYI

PR #1978, opened 2020-03-28
WELD-2613 Remove Unsafe usage from class defining logic.

manovotn:
I am not really happy with this as it means we will be seeing `IllegalAccess` warning again on JDK 11. 

However, `Unsafe` is a mess in JDK 12+ and not really working, so we cannot rely on that for future solutions anyway. Linked issue has also reported a failure with it on 11 even though we don't know the exact setup (but it has something to do with modular setup which is likely).

We now have EE integrators re-routed to the new SPI for several releases, so this problem concerns mainly SE (and possibly servlet?). I don't think there is a solution we could take that would work in both, JDK 8 and JDK 11+ and for declaring classes in JDK 11+ there is likely still the issue of not being able to declare class with different `ProtectionDomain` via new API. I am not sure we need that in SE though.


manovotn:
@mkouba FYI

PR #1981, opened 2020-04-23
WELD-2622 Update BOM, fix dependencies in examples, remove Jakarta 9 …

manovotn:
…adoc.

More updates to make master compile.
The PR excludes tons of tests because we cannot execute them without arq-weld-container.
I have at least tested against snapshot to verify that it is working.

One (hopefuly) last thing I am dealing with is that I cannot compile `examples` module.
Namely `pastecode` where `hibernate-jpamodelgen` is used. The classes it is supposed to generate and not created at all nor can I see any error. It just blows with `Cannot find symbol` when trying to compile code relying on the generated bits.


manovotn:
The `jpamodelgen` probably cannot even work as they are bound to operate on `javax` packages yet.
See [this code](https://github.com/hibernate/hibernate-orm/blob/master/tooling/metamodel-generator/src/main/java/org/hibernate/jpamodelgen/JPAMetaModelEntityProcessor.java#L46-L48).

manovotn:
Note - CI builds are bound to be failing, I have yet to alter our Jenkins; it now tries to run incontainer tests which obviously fail.

manovotn:
I temporarily excluded pastecode example due to aforementioned reason.

Next problem is JDK 11 build - in that case `core/impl` build fails with CNFE for `javax.annotation.Generated` that is coming from weld-logigng-tools.

I am not quite sure why this isn't an issue in JDK 8 though...

manovotn:
Newer version of `jboss-logging` fix that issue.

starksm64:
> I am not quite sure why this isn't an issue in JDK 8 though...

The problem is an out of date jboss.logging.processor.version. If I update to: <jboss.logging.processor.version>2.2.1.Final</jboss.logging.processor.version>

Then I am able to compile the core under JDK 11.

PR #1982, opened 2020-04-28
WELD-2626 honour resetHttpSessionAttributeOnBeanAccess in AbstractConversationC…

benjamin-confino:
…ontext

With this PR weld will - if `resetHttpSessionAttributeOnBeanAccess` is set to true - call `setAttribute` on the `conversationMap` in `AbstractConversationContext` even when weld is able to restore the `conversationMap` from a previous session.

This will resolve WELD-2626


manovotn:
The failure is an intermittent one that I am unable to locally reproduce. I've issues a tentative fix into 3.1 and master branches which could help with it.
Retest this please

PR #2000, opened 2020-06-08
WELD-2629 Correct ConversationContextDestroyedOnSessionTimeoutTest to…

manovotn:
… properly trigger session timeout.


manovotn:
Once approved and verified here, we should apply similar changes into master branch.

PR #2003, opened 2020-06-23
WELD-2570 Fix WebAppBeanArchiveScanner not excluding duplicate bean.xml correctly

karge-itestra:
See latest discussion in https://issues.redhat.com/browse/WELD-2570.

Changed WebBeanArchiveScanner such that first it resolves, the real path of the file found through the servelet context, and afterwards compares that to the beans.xml file paths found in the classpath.
This avoids duplicate entries in settings where the deployment specifies a logical path for the server that is different from the actual file path.

### DIFF
--- a/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/deployment/WebAppBeanArchiveScanner.java
+++ b/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/deployment/WebAppBeanArchiveScanner.java
@@ -72,16 +72,8 @@ public class WebAppBeanArchiveScanner extends DefaultBeanArchiveScanner {
         String contextPath = servletContext.getContextPath();
         List<ScanResult> results = new ArrayList<>(super.scan());
 
-        // All previous results for WEB-INF/classes must be ignored
-        for (Iterator<ScanResult> iterator = results.iterator(); iterator.hasNext();) {
-            ScanResult result = iterator.next();
-            String path = result.getBeanArchiveRef().toString();
-            if (path.contains(WEB_INF_CLASSES_FILE_PATH) || path.contains(WEB_INF_CLASSES)) {
-                iterator.remove();
-            } else {
-                result.extractBeanArchiveId(contextPath, WEB_INF);
-            }
-        }
+        File webInfClasses = null;
+        ScanResult webInfBeansXML = null;
 
         try {
             // WEB-INF/classes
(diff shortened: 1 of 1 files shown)


manovotn:
ok to test

environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/deployment/WebAppBeanArchiveScanner.java
```diff
@@ -100,18 +92,34 @@ public WebAppBeanArchiveScanner(ResourceLoader resourceLoader, Bootstrap bootstr
...
             }
         } catch (MalformedURLException e) {
             throw WeldServletLogger.LOG.errorLoadingResources(e);
         }
+
+        // All previous results for WEB-INF/classes must be ignored
+        for (Iterator<ScanResult> iterator = results.iterator(); iterator.hasNext();) {
+            ScanResult result = iterator.next();
+            String path = result.getBeanArchiveRef().toString();
+            if (new File(path).equals(webInfClasses)) {
```
mkouba, line 112: `webInfClasses` is `null` for WARs that are not extracted to the file system. I believe that we should also keep the original conditions, i.e. :
```java
if (path.contains(WEB_INF_CLASSES_FILE_PATH) || path.contains(WEB_INF_CLASSES) || new File(path).equals(webInfClasses)) {
 ...
}
```

karge-itestra: I wonder whether the DefaultBeanArchiveScanner would find a beans.xml in a non-extracted WAR.
Keeping the original conditions could lead to unwanted side-effects, in the unlikely case that someone calls his build directory WEB-INF. (In fact, I did that as an attempt of a quick and dirty workaround before solving it properly). We should find a condition that only excludes the _previously found_ beans.xml, as stated by the comment above the loop.

karge-itestra: However, I guess for now, any path containing WEB-INF/classes not actually being the server´s WEB-INF/classes directory can be considered a rare edge case.

karge-itestra:
Here you go, I have set up a minimal example to reproduce the issue:
https://github.com/karge-itestra/WeldWebappBeanArchiveScannerExample

manovotn:
@karge-itestra nice job with the reproducer!
I verified this locally with older and newer version and it works like a charm. Therefore, let's get this merged :)

PR #2004, opened 2020-06-25
WELD-2570 Fix WebAppBeanArchiveScanner not excluding duplicate bean.xml correctly

karge-itestra:
see https://github.com/weld/core/pull/2003


manovotn:
ok to test

PR #2005, opened 2020-07-22
WELD-2631 Destroy session context that can be left over from previous…

manovotn:
… timeouted session that ended exceptionally.

### DIFF
--- a/modules/web/src/main/java/org/jboss/weld/module/web/logging/ServletLogger.java
+++ b/modules/web/src/main/java/org/jboss/weld/module/web/logging/ServletLogger.java
@@ -97,4 +97,8 @@ public interface ServletLogger extends WeldLogger {
     @Message(id = 718, value = "No EEModuleDescriptor defined for bean archive with ID: {0}. @Initialized and @Destroyed events for ApplicationScoped may be fired twice.", format = Format.MESSAGE_FORMAT)
     void noEeModuleDescriptor(Object beanArchiveId);
 
+    @LogMessage(level = Level.DEBUG)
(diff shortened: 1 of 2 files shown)


modules/web/src/main/java/org/jboss/weld/module/web/logging/ServletLogger.java
```diff
@@ -97,4 +97,8 @@
     @Message(id = 718, value = "No EEModuleDescriptor defined for bean archive with ID: {0}. @Initialized and @Destroyed events for ApplicationScoped may be fired twice.", format = Format.MESSAGE_FORMAT)
     void noEeModuleDescriptor(Object beanArchiveId);
 
+    @LogMessage(level = Level.DEBUG)
+    @Message(id = 719, value = "When attempting to activate session context, an already active HttpSessionDestructionContext was detected. This is likely a leftover from previous sessions that ended exceptionally; this context will be terminated. ", format = Format.MESSAGE_FORMAT)
```
manovotn, line 101: @mkouba if you have an idea for better log msg, I am all ears

manovotn:
FTR I have ran this through WFLY TS as well and saw no regressions.

modules/web/src/main/java/org/jboss/weld/module/web/logging/ServletLogger.java
mkouba: Looks good but it's a bit too long. Maybe just "An active HttpSessionDestructionContext detected. This is likely a leftover from previous sessions that ended exceptionally; this context will be terminated."?

manovotn: +1, will adjust that

PR #2007, opened 2020-07-23
Add null checks into WeldInitialListener.

manovotn:
Related to weld-dev conversation - https://lists.jboss.org/pipermail/weld-dev/2020-July/003828.html

We can either blow up or just log something and skip execution. I chose to blow up as I don't think you should call other methods on a listener if you skipped init. But I don't mind going the other way either.


manovotn:
Hmm, reaction (on the email thread) was pretty much the same. Let's drop this PR then...

PR #2008, opened 2020-08-03
WELD-2630 Weld work for EE 9 servlets

manovotn:
Contains a bunch of work done for EE 9 servlets, still WIP. Contains TODOs around Undertow which is temporarily commented out.

NOTE - will require squahing before merge.


manovotn:
@starksm64 FYI this is the WIP I mentioned on weld-dev.

manovotn:
BTW at this point the CI is failing because of the missing mojarra artifact (which I had locally installed on my machine). Moving that to older version should expose the failing test.

manovotn:
Also contains test exclusion linked to [WELD-2633](https://issues.redhat.com/browse/WELD-2633).

manovotn:
retest this please
(our CI was updated with snapshotted version of WFLY we can test against; this means CI should now do full incontainer tests)

manovotn:
retest this please

manovotn:
Ok, some incontainer tests show some new failures.
For JDK 8 it's:
> ProxyCreationForSignedClassesTest.testProxyCanBeCreatedForAllBeans » UnsupportedClassVersion

Which is probably caused by the attached signed JAR which was mistakenly built using JDK 11.

For JDK 11 the error happens during execution of weld se/servlet coop test using Jetty. However, the Jetty in question here is old so it needs updating. Newer version might mean it won't run with JDK 8 anymore though.

manovotn:
Since Jetty 11 requires JDK 11, having that dependency means JDK 8 builds will blow up (at compile time, I suppose the next CI run for JDK 8 will show that).
We will have to either compile against older Jetty version and then only execute tests against 11 (we can't even run them now BTW because the Arq. adapter is busted), or switch to JDK 11-only for servlets which I am not in favor of because EE 9 in general wants to still support java 8 plus other servlets such as Tomcat do not seem to set this requirement.

manovotn:
Ok, the CI now executes even incontainer tests for EE 9 on snapshotted WFLY and passes. The only issue this PR has is the commented out parts of Undertow since there is no EE 9 version.
I might need to merge this still so that I can resolve multitude of dep. upgrades and have them properly tested...

PR #2010, opened 2020-08-11
Update javadoc generation for Weld SE shaded - this had lots package …

manovotn:
This avoids the warnings on JDK 8 which were errors with JDK 11.
However, there is still tons of other warnings (bad documentations, missing docs for return statements, ...). But those at least don't break build/release with JDK 11.

@mkouba could you please sanity-check the SE shaded javadoc jar this generates for you? With either java you have, just to have a double check on this :-)

### DIFF
--- a/environments/se/build/pom.xml
+++ b/environments/se/build/pom.xml
@@ -35,6 +35,31 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
+            <groupId>org.jboss.weld</groupId>
+            <artifactId>weld-api</artifactId>
+            <version>${weld.api.version}</version>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.spec.javax.servlet</groupId>
+            <artifactId>jboss-servlet-api_4.0_spec</artifactId>
+            <optional>true</optional>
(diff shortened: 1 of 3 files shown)


manovotn:
Argh, I forgot I need similar magic for servlet shaded, however there is a lot more dependencies over there.
I am still not certain this is correct way to do it, I would expect the plugin to pick the transitive deps on its own, but that doesn't seem to work. There is even `<includeTransitiveDependencySources>` option which I tried but that wasn't helping either.

For servlet shaded I am getting errors on non existing packages from `javax.faces`, `javax.transaction`and so on which is really weird as those are not even optional dependencies and should be present.
```
[ERROR] /home/manovotn/GitRepo/core/environments/servlet/build/target/distro-javadoc-sources/weld-jsf-3.1.5-SNAPSHOT-sources/org/jboss/weld/module/jsf/ConversationAwareViewHandler.java:23: error: package javax.faces.application does not exist
[ERROR] import javax.faces.application.ViewHandlerWrapper;
```

manovotn:
I've added a commit doing the same thing for servlet shaded JAR but that one is a lot more....massive.
Like I said, I am not convinced this is the way to do it, but I found none other so far.

PR #2042, opened 2020-08-31
Update inject and cdi TCK versions; Pending Ballots

starksm64:

### DIFF
--- a/jboss-tck-runner/pom.xml
+++ b/jboss-tck-runner/pom.xml
@@ -23,6 +23,8 @@
     <properties>
         <htmlunit.version>2.9</htmlunit.version>
         <shrinkwrap.version>1.0.1</shrinkwrap.version>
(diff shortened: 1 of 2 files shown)


manovotn:
whitelist

manovotn:
@starksm64 the build won't run by default with staging profiles (which is probably why travis failed?). Neither I think it should - is this for some temporary testing/verification you are doing?
If possible I think we should avoid having versions in master that are just staged as that can prevent reproducibility of builds.

starksm64:
@manovotn This updates the TCK versions to those that are going to be released as soon as the specification ballots pass. The staging profile is to allow testing of the Jakarta staged versions that are undergoing ballot. I updated the title to say ...Pending Ballots. I'll remove that once the ballots conclude and the TCKs are released to maven central.

manovotn:
The PR wasn't compileable.
Also, it now needs rebase because at least one of those deps is already released and was updated.

manovotn:
Closing PR for inactivity. Besides, most of the deps mentioned here should be updated at this point.

PR #2054, opened 2020-09-01
Attempt to unify htmlunit dependency used throughout the project.

manovotn:


manovotn:
Should fail for same reason as https://github.com/weld/core/pull/2053
But here we can adapt TCKs to use newer versions and then update here, but for now, we will still need to set tck runner to older version.

manovotn:
`weld-se-coop` tests are failing...
Ok, so this is where the fun (and one wasted afternoon) starts - upgrading `htmlunit` means that it out of blue starts fetching a dependency on `org.eclipse.jetty.websocket:websocket-client:jar:9.4.26.v20200117:test`.

However, for Weld 4 we are dealing with EE 9, Jetty 11. And there doesn't seem to be a Jetty 11 artifact for `websocket-client` yet. All other Jetty deps within `weld-se-coop` are for 11 version.

I have managed to first gut the dependency and replace most of its transitive deps with version for 11 which works.
However, I then found that they have released artifacts for websockets under new GAV, so I should instead be looking into what new artifact I should depend on and exclude the old one from htmlunit...

PR #2087, opened 2020-09-07
Remove portlet support.

manovotn:
I probably forgot to do this as part of the previous removal which consisted of mortbay and other old deps.
@mkouba do you agree that these bits can now be removed from Weld 4?


mkouba:
+1 if there are no "portlet consumers" ;-)

PR #2152, opened 2020-10-02
Docs - add an execution that generates multiple HTMLs; add index page…

manovotn:
…; make sure links are generated conditionally based on whether its single or multi HTML.

Summary:
* Re-introduces mutli HTML docs so that existing links (out in the wild) won't stop working
* Adds index page
* Each HTML now has its TOC and a link to the index page at its top
* All links between varying HTMLs has to be remade to be conditional in order to work correctly for both, single and multi HTML outputs

### DIFF
--- a/docs/reference/pom.xml
+++ b/docs/reference/pom.xml
@@ -59,6 +59,7 @@
                         </goals>
                         <configuration>
                             <backend>pdf</backend>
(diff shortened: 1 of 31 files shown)


manovotn:
Master branch version of this coming up soon.

PR #2153, opened 2020-10-02
Docs - add an execution that generates multiple HTMLs; add index page…

manovotn:
…; make sure links are generated conditionally based on whether its single or multi HTML.

Master branch counterpart of https://github.com/weld/core/pull/2152


mkouba (approved):
Looks really good!

PR #2159, opened 2020-10-12
WELD-2645 Change OSGi export statement for modules.

manovotn:


manovotn:
@lukasj this should fix it for you, can you verify it please?

Also, ignore any non-Travis CI results for the time being, it's being re-visited and will likely fail.

PR #2166, opened 2020-10-21
WELD-2567 WeldSEProvider should throw ISE instead of returning null i…

manovotn:
…f no container was found.


mkouba (approved):
I wonder what was the original reason to return null here?

manovotn:
@mkouba Previous CDI provider lookup couldn't properly deal with exceptions - it would just blow up and stop searching. So my guess is we did that to be CDI-friendlier despite not doing exactly what the spec defined.

PR #2167, opened 2020-10-21
WELD-2639 Allow to expand set of BDA via generic properties.

manovotn:
An impl for WELD-2639. If approved, similar change should go to 3.1 branch.


manovotn:
TODO - in order to be consistent, we should also add the ability to register this as a system property. E.g. `-Dorg.jboss.weld.se.additionalBeanDefiningAnnotations=org.my.SomeAnnotations,org.another.AnotherAnnotation`

manovotn:
@mkouba how about now?

mkouba (approved):
Looks great!

manovotn:
Rebased to trigger tests.

PR #2178, opened 2020-11-03
Try executing CDI TCK with Travis.

manovotn:


manovotn:
Hmm, the build is still running on `travic-ci.org`. I have tried to shift it to `travis-ci.com` now, so let's see if a force push does anything with it...

PR #2179, opened 2020-11-03
Try executing CDI TCK with Travis.

manovotn:
Second service - let's see if Travis chews on this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,26 @@
 language: java
+os: linux
 jdk:
   - openjdk8
   - openjdk11
-script: "mvn verify"
-sudo: false
+
+# Download latest WFLY and set JBOSS_HOME
+before_install:
+  # TODO use the following once it stops returning 404
+  #- wget https://ci.wildfly.org/guestAuth/repository/download/WF_Nightly/latest.lastFinished/wildfly-latest-SNAPSHOT.zip -o wildfly.zip
+  - wget -O wildfly.zip https://download.jboss.org/wildfly/21.0.0.Final/wildfly-21.0.0.Final.zip
+  - mkdir container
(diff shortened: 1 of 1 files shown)


manovotn:
To sum up:

Travis has a [new pricing plan](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) which basically cuts down free project builds down to 1000 minutes/month on Linux machines.

This PR build passes with just the in-container testing on WFLY. Overall it takes 30-35 minutes per build and its times 2 because we use two JDKs.
Realistically we also need to add non-container tests, examples build and javadoc build to have it complete - I estimate that would be shorter, but would still cost us at least 15 minutes per build per JDK rounding up at, let's say, **100 minutes raw execution time per PR build execution** (some of it would be parallel, but that's irrelevant now). That's basically 1/10 of what Travis allows you monthly.

Given that you can re-build the same PR several times due to literally any change you push into it, it can grow up pretty fast.
Then there is build on PR and on master branch or tags when you do release - I suppose those can be turned off.
However, there is also Dependabot which is a force to be reckoned with - this bot alone (while being useful) would be able to chew on whole Travis allowance on its own, every month.

We can also leverage GH actions and possibly combine them together. Although [GH actions also has a limit of 2000 action minutes per month](https://github.com/pricing). Even combined, that would total to 30 builds (assuming 100min builds) per month, which is very low.

Tomorrow I'll try to play with GH actions and see how it fares there (with incontainer tests for instance, because those should take the longest chunk of time), but I strongly suspect the outcome I outlined above...

manovotn:
Closing this as the resource limit Travis provides is ridiculously low. I'll remove Travis access to this project and try with GH actions instead.

PR #2180, opened 2020-11-04
Create GH Actions config file

manovotn:
First and probably wrong shot at GH actions file, supposedly for 3.1 branch.

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -1,0 +1,352 @@
+name: Weld CI
+
+on:
+  pull_request:
+    branches: [ 3.1 ]
+    # Do not run for non-code changes
+    paths-ignore:
+      - '.gitignore'
+      - '*.md'
+      - '*.adoc'
+      - '*.txt'
+
+jobs:
+  # builds Weld snapshot, downloads WFLY and upgrades it, prepares ENV variable
+  build-jdk11:
+    name: "Initial JDK 11 Weld Build + WildFly patch"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
+        uses: actions/setup-java@v1.4.3
+        with:
+          java-version: 11
+      - name: Download WildFly
+        run: |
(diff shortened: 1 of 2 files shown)


manovotn:
I've rebased all the work into one commit.

What's included:
* Junit tests
* SE tests
* Servlet tests
* SE + servlet coop tests
* CDI TCK (in container and SE)
* CDI TCK 1.2 against Weld 3 _(not sure we want to keep this?)_
* Weld integration tests (incontainer and without container)
* Examples build and test _(this is extra compared to what previous CI testing did)_

What it doesn't do (and what was in release job workflow):
* Javadoc build which happens anyway on release dry run
* Probe ftests against Firefox - I have yet to look into this but in theory, we should be able to perform it assuming those machines can have FF installed and we can download the driver manually...

The workflow is basically two sequential jobs, you first you need to build initial Weld SNAPSHOT that caches Maven repo and also build, patches and uploads WFLY for other jobs to share. After that, all remaining jobs are parallel and run pretty fast.
All of it is free to use and I don't see any monthly limitations to it so far.

@mkouba WDYT about this? If we agree it's good, I'll do similar thing for master branch (we might need to temp. skip incontainer testing there or build some EE 9 custom WFLY snapshot), Weld API (which will be very simple in comparison) and lastly for Weld parent in order to be able to test dependency upgrades there.

mkouba:
> CDI TCK 1.2 against Weld 3 (not sure we want to keep this?)

Why do we acutally run this? +1 for removal.

> Probe ftests against Firefox...

I don't think that we need to test this in a PR job anyway. The code is pretty stable and rarely changes, so I think it's OK to only run the tests as a part of a release job.

mkouba (approved):
Looks pretty good! ;-)

manovotn:
> I don't think that we need to test this in a PR job anyway. The code is pretty stable and rarely changes, so I think it's OK to only run the tests as a part of a release job.

@mkouba Well, the problem is, we now don't have a release job now. At least not yet, anyway. But I think GH actions allow this concept, I am planning to ask Guillaume about it :-)

mkouba:
> Well, the problem is, we now don't have a release job now

release job = you running all the test suites manually ;-)

manovotn:
Removed TCK 1.2 execution, added exclusion for changes to doc or readme files and cleaned up remaining TODO.

PR #2183, opened 2020-11-11
Create GH Actions config file

manovotn:
This is a draft and will not yet pass because it doesn't yet use EE 9 version of WFLY. I have talked to Brian and there is a good chance there will be nightly builds of those later this week.
So instead of creating a hacky build that needs several snapshots, I'd rather consume their nightly.

The testing setup is otherwise identical to that of 3.1 branch with the exception of different cache setup so that both branches use their own cache.

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -1,0 +1,352 @@
+name: Weld CI
+
+on:
+  pull_request:
+    branches: [ master ]
+    # Do not run for non-code changes
+    paths-ignore:
+      - '.gitignore'
+      - '*.md'
+      - '*.adoc'
+      - '*.txt'
+
+jobs:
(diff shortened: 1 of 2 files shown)


manovotn:
Since WFLY didn't yet create their EE 9 nightly stable build, I'm trying to create an extra step to build it ourselves. There should also be caching so while the first build might take a bit due to deps downloading but it should speed up afterwards. That is, after I correct all the YAML mistakes :-)

manovotn:
Ok, this seems to work just fine. We are building our own WFLY snapshot which adds about 10 mins of execution time that other jobs have to wait for.
However, we can easily remove this soon as we have a WFLY nightly link. I'll merge this and perform the same tweak in https://github.com/weld/parent/pull/43

PR #2207, opened 2020-12-03
WELD-2652 Do not try to load classes from META-INF/versions.

manovotn:
I think something as simple as this might be able to solve the problem.
There is no automated test attached because there is failure, just WARN logging happening. To test it, I used a reproducer on my branch - https://github.com/manovotn/core/tree/weld2652_reproducer

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/AnnotatedTypeLoader.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/AnnotatedTypeLoader.java
@@ -42,6 +42,7 @@ class AnnotatedTypeLoader {
     final ContainerLifecycleEvents containerLifecycleEvents;
 
     static final String MODULEINFO_CLASS_NAME = "module-info";
+    static final String MULTI_RELEASE_JAR_PATH = "META-INF.versions";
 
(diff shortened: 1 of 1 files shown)


impl/src/main/java/org/jboss/weld/bootstrap/AnnotatedTypeLoader.java
```diff
@@ -42,6 +42,7 @@
     final ContainerLifecycleEvents containerLifecycleEvents;
 
     static final String MODULEINFO_CLASS_NAME = "module-info";
+    static final String MULTI_RELEASE_JAR_PATH = "META-INF.versions";
```
mkouba, line 45: Hm, so a class from `META-INF/versions` has a name like `META-INF.versions.9.org.acme.Foo`?

manovotn: Forgot to mention that, sorry. Yeah, within the `loadClass(String className)` method, you will see that class as something like this: `META-INF.versions.12.io.smallrye.context.impl.JdkSpecificImpl`

manovotn: @mkouba are you OK with this PR or should I change something?

mkouba (approved):
LGTM

PR #2209, opened 2020-12-04
Upgrade Weld API to 4.0.CR1.

manovotn:


manovotn:
Turned out to need more than just API bump.
By upgrading CDI, there is a failing `CustomCDIProviderTest` in our TS and in TCK - https://github.com/eclipse-ee4j/cdi-tck/issues/217.
I've issues a fix for our test, TCK test and added an exclusion on our side until there is a release with it.

PR #2210, opened 2020-12-07
WELD-2650 Use Phaser to implement a barrier system and prevent race c…

manovotn:
…ondition.

The issue here is that sometimes the second assertion (== second https request) gets completed before the first one is done with bean destructions. Hence we need some sort of barrier approach to make sure we give it enough time.


manovotn:
@pkremens this is my take on it, opinions and review are welcome :-)
With this approach, I didn't manage to reproduce the failure, even if I try to artificially delay bean destructions.

pkremens:
Looking good :+1: this one should fix it, thanks!

manovotn:
Although the test failure was unrelated (failed to download artifact), I've restarted the jobs.

PR #2236, opened 2021-01-08
WELD-2618 Make proxy name creation more deterministic.

manovotn:
See https://issues.redhat.com/browse/WELD-2618 for more information on what is the current "issue" with the creation process and for reproducer.

I did use the same test the aforementioned issue has to confirm results although I chose not to add it into this PR as it uses not-so-nice reflections to invoke the method because whole `ProxyFactory` isn't public.

I do not have a string opinion about the current versus proposed proxy names, but I don't see a reason to make it more deterministic and readable.
@mkouba WDYT?

### DIFF
--- a/impl/src/main/java/org/jboss/weld/util/Proxies.java
+++ b/impl/src/main/java/org/jboss/weld/util/Proxies.java
@@ -21,10 +21,13 @@ import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/jboss/weld/util/Proxies.java
```diff
@@ -51,19 +54,28 @@
 
     public static class TypeInfo {
 
-        private final Set<Class<?>> interfaces;
-        private final Set<Class<?>> classes;
+        private final List<Class<?>> interfaces;
+        private final List<Class<?>> classes;
+        private final Comparator<Class<?>> classComparator = Comparator.comparing(Class::getName);
+        private boolean interfacesSorted;
```
mkouba, line 60: I don't think it makes sense to keep this class mutable and to introduce these flags. `TypeInfo.add(Type)` is only called in the `Proxies.TypeInfo.of(Set<? extends Type>)` method. So it would be probably reasonable to make this class immutable, i.e. create the sorted and unmodifiable lists in the constructor.

manovotn: This should now be addressed.

PR #2237, opened 2021-01-08
WELD-2618 Make proxy name creation more deterministic.

manovotn:
3.1 counterpart of https://github.com/weld/core/pull/2236

### DIFF
--- a/impl/src/main/java/org/jboss/weld/util/Proxies.java
+++ b/impl/src/main/java/org/jboss/weld/util/Proxies.java
@@ -21,10 +21,13 @@ import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


impl/src/main/java/org/jboss/weld/util/Proxies.java
```diff
@@ -51,13 +54,38 @@
...
-        private TypeInfo() {
-            super();
-            this.interfaces = new LinkedHashSet<Class<?>>();
-            this.classes = new LinkedHashSet<Class<?>>();
+        private TypeInfo(Set<? extends Type> types) {
+            Comparator<Class<?>> classComparator = Comparator.comparing(Class::getName);
+            List<Class<?>> foundInterfaces = new ArrayList<>();
+            List<Class<?>> foundClasses = new ArrayList<>();
+
+            types.stream().forEach(type -> add(type, foundInterfaces, foundClasses));
```
mkouba, line 65: Just out of curiosity, what's the reason to use a stream and lamda here? ;-)

mkouba (approved):
Looks good!

impl/src/main/java/org/jboss/weld/util/Proxies.java
manovotn: Just to see if you are going to comment on it :-P

PR #2238, opened 2021-01-11
Attempt at a job that creates WFLY patches and Weld distribution.

manovotn:
On PR trigger is to be removed once I can see that the file is OK and generates expected result.
The tag-creation trigger I am not sure works yet, but guess that's to be resolved once I try to make a tag along with next release.

### DIFF
--- a/.github/workflows/ci-release-job.yml
+++ b/.github/workflows/ci-release-job.yml
@@ -1,0 +1,79 @@
+# This job triggers automatically when a tag is created from the branch
+# Its purpose is to create and archive WFLY patch as zip file and to build and archive Weld distribution
+name: Weld Release
+
+on:
+  push:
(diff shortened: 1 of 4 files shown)


manovotn:
Turns out the gradle build script wasn't allowing builds from SNAPSHOT (it expected all artifacts to be at least staged).
I've changed that, but in order to pass this build, we'll also need to produce javadocs - will try to fix that shortly.

manovotn:
Seems to work fine, example of artifact outputs (WFLY patch, dist zip and docs zip) can be seen here - https://github.com/weld/core/pull/2238/checks?check_run_id=1682592919
I've removed the on-PR trigger. Tomorrow during release I'll check whether it correctly triggers on tag creation.

PR #2240, opened 2021-01-21
WELD-2581 Rework class defining for Weld in JDK 11+

manovotn:
**This is work in progress.**

First draft of how we may solve this issue - it combines `MethodHandles.Lookup`, which is the intended way, with class loader wrapping which we still need for classes where we define new (non-existent) package.

Changes concern mostly SE and there are now two test failures which I am trying to understand.

CI failures are expected because the project will now need to build with JDK 11 at all times (which CI setup doesn't take into account). It should however be testable with JDK 8 as well as 11.


manovotn:
NOTE - this PR requires Weld parent version 43 which I have released into central today but apparently it hasn't synced yet (OSS was really sluggish when I did the release).

manovotn:
> I don't know if this idea would work

Care to elaborate? What exactly wouldn't work?

About the new module - we cannot use `java.langProxy` to define our proxies. We have custom with with certain functionality to enable CDI capabilities.
That being said, we need to be able to define classes with arbitrary made-up packages on the fly. I am not sure a custom module would allow that? Besides custom module might present further issues with access to/from it to other classes in modular system, right?

Thihup:
> > I don't know if this idea would work
> 
> Care to elaborate? What exactly wouldn't work?
> 
Sorry about the confusion, I mean to say that my idea could not work, not this PR. 

> About the new module - we cannot use `java.langProxy` to define our proxies. We have custom with with certain functionality to enable CDI capabilities.
> That being said, we need to be able to define classes with arbitrary made-up packages on the fly. I am not sure a custom module would allow that? Besides custom module might present further issues with access to/from it to other classes in modular system, right?
> 

Actually I don't have a lot of experience with Proxies, but I guess the JDK had similar problems implementing Proxy classes. 

However I guess it wouldn't require a new classloader. And about the access I guess you're right. I don't know if the JDK proxies have this problem, but we might have. 

I actually have to see what exactly the Synthetic modifier means and how it relates to the other modules, but I've thought about this new module being an open module by default.

manovotn:
The more I think about it, the more I see the issue in the fact that we can declare arbitrary amount of new packages.
I was thinking we could to private lookup in our own module and define it there, but it still won't work because the lookup is based on a package and can only define classes into that package.

Obviously, we can create a dummy package with dummy class in and and ditch the proxies there, but I think that's error prone. One of the issues would be that it will blow up the moment we try to define two proxies for bean two beans based on same interfaces into equal package (say two beans for `List` where we change the package names) - although I haven't tried it locally to see if this scenario is realistic.

And I don't know of any way to programmatically "enlist" new packages into our own module. If we could do that, then we could keep using purely `MethodHandles.Lookup` too.

Thihup:
Do you know if it would be possible to use hidden classes?
Maybe it would be more feasible to create a dummy package and create the classes there.

https://openjdk.java.net/jeps/371

manovotn:
That's interesting feature but I don't think it anyhow helps our situation - the class would still mean we will pour everything into one one package. The package of the hidden class. Or did I miss anything?

Thihup:
> That's interesting feature but I don't think it anyhow helps our situation - the class would still mean we will pour everything into one one package. The package of the hidden class. Or did I miss anything?

Exactly, however, the JVM will create a unique name for the class, so from my understanding, it would be possible to have two proxies with the same name but the JVM will create a unique name for the classes.
I guess this is how the Lambdas works in the newer versions, there are just a few interfaces but a lot of classes with the same name and the only difference is the unique identifier the JVM gives.
I will give it a try and see if it works.

> BTW Jdk 15+ isn't really such blocker. We just need a solution before we hit the hard illegal access wall which i think was jdk 16 (not sure from the top of my head).

OK. Yeah, it is the JDK 16 that will block the access by default ;)

Thihup:
I have looked again at how the JDK Proxy implementation works and looks like they create a fake package (_com.sun.proxy_).
I have searched in the entire JDK codebase and couldn't find any reference to that package, so I guess it is still possible to create arbitrary packages if you create a module.

manovotn:
Not sure I am following you - we cannot use `java.lang.reflect.Proxy` and I haven't really seen a manual way to declare new `Module` at runtime yet. Maybe I just missed it somewhere?
I do recall being able to declare whole module layers, but that looks like an overkill and TBF I don't quite yet understand the full extent of what it does just yet.

Besides, even in the javadoc of `Module` (whether dynamically creatd one or not) I don't see a way to declare packages. Or do they use some subclass of Module that allows it?

If you have a code to point at, that would be great.
And of course if you want to contribute (based on this PR or in completely new one) that's an option as well :-)

Thihup:
Let's recap:
My initial idea was to see how the JDK has implemented the Proxies to give us some idea of how to proceed. My goal wasn't to use the Proxy class but only to have a look at how they overcome this problem.
I have linked the source code of the JDK where they create a module. [1]

However, with my tests I found the following:
- They use special access to the _Modules.defineModule_. [2]
- They use _JLA.defineClass_ to actually define the class. [3]

We can still create a module in runtime, it's very bloated but it is possible (you need to create a ModuleFinder, a ModuleDescriptor, and a ModuleReference).
But we still cannot use this newly created module to allow us to create a class because we still need some class in that package :|

Sorry about all the confusion. It wasn't all clear to me what exactly would be needed to help fix this problem until I actually had started to create some experiments :)

[1] https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/reflect/Proxy.java#L900
[2] https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/reflect/Proxy.java#L910
[3] https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/reflect/Proxy.java#L558

manovotn:
> Sorry about all the confusion.

Nothing to be sorry about. In fact, thanks for digging into this.
I am also just trying to get the hang of it :-)

> But we still cannot use this newly created module to allow us to create a class because we still need some class in that package :|

That's shame. Means we are back to square one and using class loaders I suppose. Though I expected they will be using internal APIs to achieve that. Seems like I have to get back to figuring out why the remaining tests fail with current CL solution.

manovotn:
Added some changes:
* weld-parent dep is back to SNAPSHOT because there are more changes needed
  * enforcer plugin declaration is too restrictive not allowing JDK 8 testing
* Added enforcer rule for JDK 11 just inside `core-impl` module
* Fixed `CreateAnnotatedTypeWithIdTest` by declaring the new annotated in independent class loader to avoid duplicate declaration issue
* Fixed the serialization test by deserializing it with same CL that defined the proxy instance (which is our own)
  * I am not sure such thing is an "expected behaviour" if you are implementing (de)serialization in your SE app

This means I can get this PR locally to build and pass tests.
It will still fail here because of SNAPSHOT parent dep as well as CI misconfiguration that will try to build it with JDK 8.
I can fix those in upcoming days to get a clearer picture.

Though the question whether it is a correct solution still stands :-)

manovotn:
Email discussion with David confirmed my suspicion that we cannot really avoid dual approach that is suggested in this PR as there is no way to define classes into non-existing packages with modular approach.
There are probably things we could do for (se)serialization when using our CL in SE but I would rather leave that for later if we see this is an issue.

@mkouba I'd like your review for this. If you agree it's OK then I'll create a similar PR for master. Also, do you think this change is worth a minor version bump (e.g. from 3.1 to 3.2)? On one hand it can be a bigger deal, OTOH it's just in SE and all other environments won't be affected which makes me think the minor could stay as is (which will also be easier for WFLY/EAP and others to chew on).

mkouba:
> I'd like your review for this. If you agree it's OK then I'll create a similar PR for master.

Looks good but my knowledge of class loading in JPMS is very limited.

> Also, do you think this change is worth a minor version bump (e.g. from 3.1 to 3.2)?

I don't think so.

mkouba (approved):

manovotn:
I have pushed some code format changes and removal of TODOs.

Other things I am now looking into:
* adding a caching for the custom CL so we avoid creating new ones if we can
* turning the implementation into our own default `ProxyServices` impl which would be used if there is no other
* Looking into shaded JARs not using the correct implementation (and hence still throwing warning)

manovotn:
(Although GH doesn't seem to acknowledge it yet,) I've pushed another commit that transforms current impl into our own impl of `ProxyServices`.
Along with it we could remove some deprecated methods from the API and perhaps deprecate another one.
There are a few TODOs that I'll get back to tomorrow but overall I got the tests to pass with it on both JDKs which is a good start.

I have yet to look into why shaded jars don't respect alternative MR impls of classes.

manovotn:
(Although GH doesn't seem to acknowledge it yet,) I've pushed another commit that transforms current impl into our own impl of `ProxyServices`.
Along with it we could remove some deprecated methods from the API and perhaps deprecate another one.
There are a few TODOs that I'll get back to tomorrow but overall I got the tests to pass with it on both JDKs which is a good start.

I have yet to look into why shaded jars don't respect alternative MR impls of classes.

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -174,28 +167,15 @@ public ProxyFactory(String contextId, Class<?> proxiedBeanType, Set<? extends Ty
...
-            // No interface beans must use the bean impl as superclass
+            // No interface beans, must use the bean impl as superclass
             superClass = proxiedBeanType;
         }
         this.beanType = superClass;
 
         addDefaultAdditionalInterfaces();
         baseProxyName = proxyName;
+        // TODO can we expect that all services now support class defining? Throw an exception if they don't?
         proxyServices = Container.instance(contextId).services().get(ProxyServices.class);
```
manovotn, line 178: Depending on what we do in https://github.com/weld/api/pull/126, we might want to check this and throw in case someone provides custom `ProxyServices` that don't implement it. WDYT @mkouba ?

manovotn:
I have changed the implementation into our own `ProxyServices` impl that supports class defining in two variants.
Shaded artifacts not also contain the appropriate MR jar attribute in order to work properly which was causing the issues I mentioned earlier. This means we are free to delete `ClassFileUtils` as well as `SimpleProxyServices` and its extending class.
@mkouba please re-review if you have some time, I hope this is the final form  (apart from removing TODOs) in which I would like to merge it :)

manovotn:
Eliminated remaining TODOs accordingly to what we do in API (no removals, one additional deprecation).
Weld now checks that all implementations of `ProxyServices` support class defining and there is an automated test for it.

manovotn:
Master branch counterpart is available at https://github.com/weld/core/pull/2259

manovotn:
I am going to squash and merge this. Any further feedback is still welcome on this PR or via Weld JIRA.

PR #2265, opened 2021-02-23
WELD-2658 Fire ProcessBean event for enabled synthetic alternatives.

manovotn:


impl/src/main/java/org/jboss/weld/util/Beans.java
```diff
@@ -223,16 +224,22 @@ public static boolean containsAllInterceptionBindings(Set<Annotation> expectedBi
...
                     if (enabled.isEnabledAlternativeStereotype(stereotype)) {
-                        return true;
+                        isEnabled = true;
+                        break;
                     }
                 }
-                return false;
             }
+            // For synthetic enabled alternatives, the ModuleEnablement may not yet be aware of them
+            if (!isEnabled && bean instanceof WeldBean && ((WeldBean<?>) bean).getPriority() != null) {
```
mkouba, line 239: Hm, how about custom beans implementing `jakarta.enterprise.inject.spi.Prioritized`? I'm just curious ;-)

manovotn: Good point, it is probably affected as well. I'll check it

manovotn: @mkouba how about now? :)

mkouba (approved):
Looks good now!

PR #2289, opened 2021-03-17
Update WFLY version for patches, add on-release job to build and test…

manovotn:
… patch.

This should theoretically let us build patches for Weld 4 (which we didn't do until now since there was no WFLY release) but in practice the patch-gen util fails. I am trying to see why it happens and whether it's misconfiguration or just plain impossible to use that tool for WFLY preview ATM.


manovotn:
Turns out this isn't going to work, we have to revisit how we create wfly patches.

PR #2294, opened 2021-03-25
WELD-2662 Fix an issue where you couldn't create proxy for producer m…

manovotn:
…ethod with return type from another package.

The actual issue lies in `WeldDefaultProxyServices#defineWithMethodLookup` where we need to determine which module is the target of lookup. We do so by looking at earliest occurrence of `$` which was missing in this scenario.


manovotn:
OK, the PR should now account for the nested class scenario and the proxy name we create contains the declaring class name as well.

PR #2297, opened 2021-04-08
WELD-2666 - Reproducer and potential fix for proxy naming of hierarchical interfa…

manovotn:
…ce-based beans.

This PR attempts to address the issue that we sometimes create a non-existent combination of proxy package and it's classes. This leads to issues when doing JDK 11 lookup. The fix here lies in choosing the class sequence first and attaching the respective package of the first class chosen.

This should cover the two cases I and @darranl discovered.

However, there is one more TODO in the code where I would like @mkouba opinion.
It is a case of extending a `Principal` (or any other interface that comes from `java` package) - in current solution the `Principal.class.getName()` alphabetically precedes the extending interface which in proxy creation leads to a proxy such as `java.security.Principal$AMuchBetterPrincipal$....`.
This in turns means that we will have to use Weld custom class loader to define the class instead of `MethodHandles.Lookup`. So I was thinking if it's worth a shot at re-ordering the interfaces so that (if applicable) we always put a non `java.*` package classes first. WDYT Martin?

### DIFF
--- a/impl/src/main/java/org/jboss/weld/util/Proxies.java
+++ b/impl/src/main/java/org/jboss/weld/util/Proxies.java
@@ -24,10 +24,12 @@ import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import javax.enterprise.inject.spi.Bean;
@@ -39,6 +41,8 @@ import org.jboss.weld.exceptions.UnproxyableResolutionException;
 import org.jboss.weld.logging.UtilLogger;
 import org.jboss.weld.logging.ValidatorLogger;
 import org.jboss.weld.util.collections.Arrays2;
+import org.jboss.weld.util.collections.ImmutableList;
+import org.jboss.weld.util.collections.ImmutableMap;
 import org.jboss.weld.util.reflection.Reflections;
 
 /**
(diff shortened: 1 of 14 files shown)


impl/src/main/java/org/jboss/weld/util/Proxies.java
```diff
@@ -56,32 +58,37 @@
...
-            types.stream().forEach(type -> add(type, foundInterfaces, foundClasses));
+            types.stream().forEach(type -> add(type, foundInterfaces, foundClasses, classToPackage));
 
             // sort both collections and create unmodifiable lists
             Collections.sort(foundClasses, classComparator);
             Collections.sort(foundInterfaces, classComparator);
+            // TODO should we re-sort this in a way that non-java[x] class names are first?
             this.interfaces = Collections.unmodifiableList(foundInterfaces);
             this.classes = Collections.unmodifiableList(foundClasses);
+            this.classToPackageMap = Collections.unmodifiableMap(classToPackage);
```
mkouba, line 77: I think that using Weld immutables would be more effective here, e.g. the list of interfaces can often be empty.

impl/src/main/java/org/jboss/weld/util/Proxies.java
```diff
@@ -56,32 +58,37 @@
...
             List<Class<?>> foundClasses = new ArrayList<>();
+            Map<Class<?>, String> classToPackage = new HashMap<>();
 
-            types.stream().forEach(type -> add(type, foundInterfaces, foundClasses));
+            types.stream().forEach(type -> add(type, foundInterfaces, foundClasses, classToPackage));
 
             // sort both collections and create unmodifiable lists
             Collections.sort(foundClasses, classComparator);
             Collections.sort(foundInterfaces, classComparator);
+            // TODO should we re-sort this in a way that non-java[x] class names are first?
```
mkouba, line 74: I like the idea...

impl/src/main/java/org/jboss/weld/util/Proxies.java
```diff
@@ -56,32 +58,37 @@
 
         private final List<Class<?>> interfaces;
         private final List<Class<?>> classes;
+        private final Map<Class<?>, String> classToPackageMap;
```
mkouba, line 61: BTW you're aware of the fact that `Class` does not implement equals()/hashCode() so the object identity will be used instead, right? ;-)

manovotn: Good point, that slipped me. I can actually easily make it `Map<String, String>` as further down the line we use the string representation anyway.

impl/src/main/java/org/jboss/weld/util/Proxies.java
manovotn: Ok, will add something. Theoretically a smarter comparator should be able to achieve this.
But I think we should only sort interfaces in this case, otherwise we might break something.

manovotn:
Updated.
Tests now assert that the proxy class has certain format.
Added custom comparator for *interfaces only* that sorts them alphabetically but tries to always put anything not starting with `java.` or `javax.` first. This means the `Principal` test now uses a non-java package proxy. The classes in `TypeInfo` are still sorted just plainly alphabetically.

@mkouba please re-review when you get the chance :-)

impl/src/main/java/org/jboss/weld/util/Proxies.java
```diff
@@ -56,32 +59,61 @@
...
+            Map<String, String> classToPackage = new HashMap<>();
 
-            types.stream().forEach(type -> add(type, foundInterfaces, foundClasses));
+            types.stream().forEach(type -> add(type, foundInterfaces, foundClasses, classToPackage));
 
             // sort both collections and create unmodifiable lists
-            Collections.sort(foundClasses, classComparator);
-            Collections.sort(foundInterfaces, classComparator);
+            Collections.sort(foundClasses, Comparator.comparing(Class::getName));
+            Collections.sort(foundInterfaces, (o1, o2) -> {
```
mkouba, line 73: It might make sense to turn the comparator into a class and write a unit test for it ;-).

manovotn: The tests in this PR actually assert the ordering (e.g. the resulting class name).
I wanted to keep the comparator isolated in the class but I can extract it...

impl/src/main/java/org/jboss/weld/util/Proxies.java
```diff
@@ -56,32 +59,61 @@
...
+                        // first class is not prefixed, second class is
+                        return -1;
+                    } else {
+                        // neither class is prefixed
+                        return firstClassName.compareTo(secondClassName);
+                    }
+                }
+            });
             this.interfaces = Collections.unmodifiableList(foundInterfaces);
             this.classes = Collections.unmodifiableList(foundClasses);
```
mkouba, line 101: I know that it's not related to this PR but it would not hurt to use ``ImmutableList``s here as well...

impl/src/main/java/org/jboss/weld/util/Proxies.java
mkouba: It could be a nested class...

mkouba: > The tests in this PR actually assert the ordering (e.g. the resulting class name).

Yeah, I know. But a simple unit test is IMO always a good start for similar stuff. An arquillian test should always follow of course ;-).

manovotn:
@mkouba third time's the charm? ;-)

Extracted the comparator class and added unit test for it.

mkouba (approved):

manovotn:
Hmm, I am not sure what's with the CI in the past two days.
Those failures are pretty unrelated and different tests crash every time...

PR #2299, opened 2021-04-08
[WELD-2664] Don't wrap ImmediateInstanceFactory in WeldInstanceFactory.

darranl:
Undertow uses the ImmediateInstanceFactory where a pre-instantiated instance of the Object should be used so this should not be wrapped in a WeldInstanceFactory as that attempts to create a second instance.

https://issues.redhat.com/browse/WELD-2664

This is the master PR for https://github.com/weld/core/pull/2296

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .project
+.checkstyle
 .classpath
 .settings
 .factorypath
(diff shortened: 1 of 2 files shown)


manovotn:
The CI failures are unrelated; something's off probably due to some infrastructure update or just GH service awkwardness. I have yet to investigate that.

PR #2360, opened 2021-05-17
WELD-2672 Use package name of bean class for proxies

beikov:
https://issues.redhat.com/browse/WELD-2672

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
@@ -51,6 +51,7 @@ import org.jboss.classfilewriter.code.CodeAttribute;
 import org.jboss.classfilewriter.util.Boxing;
 import org.jboss.classfilewriter.util.DescriptorUtils;
 import org.jboss.weld.Container;
+import org.jboss.weld.bean.RIBean;
 import org.jboss.weld.bean.builtin.AbstractBuiltInBean;
 import org.jboss.weld.config.WeldConfiguration;
 import org.jboss.weld.exceptions.DefinitionException;
@@ -240,7 +241,12 @@ public class ProxyFactory<T> implements PrivilegedAction<T> {
 
     private static ProxyNameHolder createCompoundProxyName(String contextId, Bean<?> bean, TypeInfo typeInfo, StringBuilder name) {
         String className;
-        String proxyPackage = null;
+        String proxyPackage;
+        if (bean instanceof RIBean) {
+            proxyPackage = typeInfo.getPackageNameForClass(((RIBean) bean).getType());
+        } else {
+            proxyPackage = typeInfo.getPackageNameForClass(bean.getBeanClass());
+        }
         final List<String> interfaces = new ArrayList<String>();
         final Set<String> declaringClasses = new HashSet<>();
         for (Class<?> type : typeInfo.getInterfaces()) {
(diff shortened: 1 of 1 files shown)


impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -240,7 +240,7 @@ static String getProxyName(String contextId, Class<?> proxiedBeanType, Set<? ext
 
     private static ProxyNameHolder createCompoundProxyName(String contextId, Bean<?> bean, TypeInfo typeInfo, StringBuilder name) {
         String className;
-        String proxyPackage = null;
+        String proxyPackage = typeInfo.getPackageNameForClass(bean.getBeanClass());
```
manovotn: Hmm, I think this can result in a mismatch of proxy package name with the first class name which will further down the line lead to what was fixed in https://issues.redhat.com/browse/WELD-2666

This is the reason why we had the current `if` condition inside the loop.

Also note that `bean.getBeanClass()` for producer method will, by CDI definition, return the class of the bean that declares the producer method. See https://github.com/eclipse-ee4j/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/Bean.java#L37-L43
This is something you don't want because the class of the declaring bean can be in a different module than where the proxy needs to be created.

manovotn (changes requested):
Thanks for the PR but there are test failures because this re-introduces problems described in [WELD-2666](https://issues.redhat.com/browse/WELD-2666).

To solve this, it would be nice to have an automated test, or at least a reproducer project that we can use. Do you have either?

I don't yet know how to solve this exactly, but we'd need to make sure that proxy package and first declared class align together plus we'd need to be able to always deterministically tell which interface to use (e.g. in your example from JIRA, `B` should precede `A` which will not be the case now).
However, by the time we invoke `createCompoundProxyName(...)`, we don't really have information on which class is which. We would either need to restructure it, or do some assignability magic perhaps.

impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
beikov: I just updated the PR, could you please take another look?

beikov: Also, I'm not very familiar with the code base but I outlined the case in the issue: https://issues.redhat.com/browse/WELD-2672

manovotn: I think I understand the general underlying issue and can create a similar reproducer but it would be without actual modular structure because we don't have that in tests (or I just don't know how to do it alongside current test setup).
For that reason, having a standalone reproducer app with actual modular setup would be very valuable.

As for the PR, I think you will still have the same problem of mismatch between proxy package and the first class which should result in CI failure again...

manovotn: I have some idea that I can try to implement this afternoon and see if it helps here. I'll report back once I know more.
If you could cobble the reproducer app together, that would help greatly.

manovotn:
I have some WIP visible on this branch - https://github.com/manovotn/core/tree/weld2672_master

It should fix your case but I am having issue with interception factory tests which I am trying to fix now.
The branch also contains a test that should simulate what you are trying to achieve (e.g. to have the proxy always in the package of the most specific class we know of).

beikov:
Not sure about the implementation you did, but the test is doing what I described, yes. So if that test succeeds, this should be fine.

manovotn:
Ok, after a long time debugging I found the silly mistake I made and fixed that.
My branch is now passing the basic tests, so I have turned it into a PR - https://github.com/weld/core/pull/2362

Closing this PR in favor of the above mentioned one.

PR #2362, opened 2021-05-18
WELD-2672 Make sure proxies for producer always have the package and …

manovotn:
…class of the most specific type we know of.

Also add a test that simulates a situtation similar to the reported problem.


manovotn:
@beikov if you could confirm that the snapshot from this PR/branch fixes your problem, that would be great :-)

beikov:
Hey @manovotn, thanks for the fix. I just tried it and it works like a charm. Looking forward to use 3.1.8.Final :)

manovotn:
@mkouba could you review this, please?

mkouba (approved):
Looks good.

PR #2365, opened 2021-05-31
WELD-2673 @Decorated Bean<> injection point can get populated with wrong meta data

manovotn:
Automated test includes two decorators (which are the issue) and two interceptors (which seem to work fine), both of which use injected metadata, applied to a single bean method invocation.

The issue seems to be that decorators create a different creational context hierarchy than interceptors,
Interceptors always have a parent CC of the bean the intercept whereas with multiple decorators, that doesn't seem to be true.
With interceptors the CC hierarchy is: `ActualBean` -> [`DecoratorOne`, `DecoratorTwo`]
With decorators the CC hierarchy is: `ActualBean` -> `DecoratorOne` -> `DecoratorTwo`

The current "fix" here is really just a workaround. I need to figure out why are the CCs created differently for decorators in the first place.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/builtin/DecoratedBeanMetadataBean.java
+++ b/impl/src/main/java/org/jboss/weld/bean/builtin/DecoratedBeanMetadataBean.java
@@ -16,10 +16,13 @@
  */
 package org.jboss.weld.bean.builtin;
 
+import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.util.Collections;
 import java.util.Set;
 
+import jakarta.enterprise.context.spi.Contextual;
+import jakarta.enterprise.context.spi.CreationalContext;
 import jakarta.enterprise.inject.Decorated;
 import jakarta.enterprise.inject.spi.Bean;
 import jakarta.enterprise.inject.spi.Decorator;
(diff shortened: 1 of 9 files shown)


impl/src/main/java/org/jboss/weld/bean/builtin/DecoratedBeanMetadataBean.java
```diff
@@ -50,6 +56,32 @@ protected void checkInjectionPoint(InjectionPoint ip) {
...
     }
 
+    @Override
+    protected Bean<?> newInstance(InjectionPoint ip, CreationalContext<Bean<?>> ctx) {
+        checkInjectionPoint(ip);
+
+        WeldCreationalContext<?> decoratorContext = getParentCreationalContext(ctx);
+        WeldCreationalContext<?> beanContext = getParentCreationalContext(decoratorContext);
+        // in case of multiple decorators, the parent context can be another decorator
+        // TODO this is hack; instead I need to find where we create the CC hierarchy and why it is like this
```
mkouba, line 66: This "hack" seems reasonable to me ;-)

manovotn: Agreed, after some more investigation and discussion, I think this first approach I tried is actually a correct solution.
I will remove the TODO and keep just a comment explaining the need for it.

PR #2379, opened 2021-06-02
Update for the CDI-TCK 4.0 master changes to split the test suite int…

starksm64:
…o standalone and web profile tests

Update the arquillian protocol handler to servlet 5.0 for the wildfly integration

### DIFF
--- a/jboss-tck-runner/pom.xml
+++ b/jboss-tck-runner/pom.xml
@@ -501,6 +501,17 @@
                         <version>${surefire.version}</version>
                         <configuration>
                             <excludedGroups>${excluded.groups}</excludedGroups>
+
+                            <argLine>
+                                --add-exports=java.desktop/sun.awt=ALL-UNNAMED
+                                --add-opens=java.base/java.io=ALL-UNNAMED
+                                --add-opens=java.base/java.lang=ALL-UNNAMED
+                                --add-opens=java.base/java.lang.invoke=ALL-UNNAMED
+                                --add-opens=java.base/java.security=ALL-UNNAMED
+                                --add-opens=java.base/java.util=ALL-UNNAMED
+                                --add-opens=java.management/javax.management=ALL-UNNAMED
+                                --add-opens=java.naming/javax.naming=ALL-UNNAMED
+                            </argLine>
                             <properties>
                                 <property>
                                     <!-- Note a 7.4.0 issue results in NPEs and useless output if you go higher than 1 here -->
@@ -519,7 +530,19 @@
                             </systemPropertyVariables>
                             <systemProperties>
                                 <jacoco.agent>${jacoco.agent}</jacoco.agent>
-                                <additional.vm.args>-Dtest=org.jboss.cdi.tck.tests.deployment.packaging.installedLibrary.InstalledLibraryEarTest</additional.vm.args>
+                                <!-- Specifiy any garbage collection related args here -->
+                                <gc.args></gc.args>
+                                <!-- This property propagates to the javaVmArguments in arquillian.xml -->
+                                <additional.vm.args>
+                                    --add-exports=java.desktop/sun.awt=ALL-UNNAMED
+                                    --add-opens=java.base/java.io=ALL-UNNAMED
+                                    --add-opens=java.base/java.lang=ALL-UNNAMED
+                                    --add-opens=java.base/java.lang.invoke=ALL-UNNAMED
+                                    --add-opens=java.base/java.security=ALL-UNNAMED
+                                    --add-opens=java.base/java.util=ALL-UNNAMED
+                                    --add-opens=java.management/javax.management=ALL-UNNAMED
+                                    --add-opens=java.naming/javax.naming=ALL-UNNAMED
+                                </additional.vm.args>
                                 <illegal.access.option>${surefire.plugin.jdk9.args}</illegal.access.option>
                             </systemProperties>
                             <suiteXmlFiles>
--- a/jboss-tck-runner/pom.xml
+++ b/jboss-tck-runner/pom.xml
@@ -23,8 +23,6 @@
     <properties>
         <!-- This matches the htmlunit version in TCK -->
         <htmlunit.version>2.50.0</htmlunit.version>
-        <!-- Make the surefire plugin version explicit -->
-        <surefire.version>3.0.0-M5</surefire.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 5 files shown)


jboss-tck-runner/pom.xml
```diff
@@ -21,8 +21,10 @@
     <url>http://weld.cdi-spec.org</url>
 
     <properties>
-        <htmlunit.version>2.9</htmlunit.version>
-        <shrinkwrap.version>1.0.1</shrinkwrap.version>
+        <!-- This matches the htmlunit version in TCK -->
+        <htmlunit.version>2.50.0</htmlunit.version>
```
manovotn, line 25: This should be inherited from core POM, see https://github.com/weld/core/blob/master/pom.xml#L72
Therefore, this shouldn't be needed (probably a leftover from some temporary issues).

jboss-tck-runner/pom.xml
```diff
@@ -21,8 +21,10 @@
...
+        <!-- Make the surefire plugin version explicit -->
+        <surefire.version>3.0.0-M5</surefire.version>
```
manovotn, line 27: The very same version is declared by weld-parent project that is a `<parent>` of this one. So you should be already getting this version.
See https://github.com/weld/parent/blob/44/pom.xml#L112

jboss-tck-runner/pom.xml
```diff
@@ -236,17 +247,46 @@
                 <plugins>
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
-                        <artifactId>maven-surefire-plugin</artifactId>
+                        <artifactId>maven-dependency-plugin</artifactId>
+                        <version>3.1.2</version>
```
manovotn, line 251: This is also inherited from weld-parent POM and thereby redundant.
See https://github.com/weld/parent/blob/master/pom.xml#L90

jboss-tck-runner/pom.xml
```diff
@@ -417,22 +501,33 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-surefire-plugin</artifactId>
+                        <version>${surefire.version}</version>
                         <configuration>
                             <excludedGroups>${excluded.groups}</excludedGroups>
+                            <properties>
+                                <property>
+                                    <name>surefire.testng.verbose</name>
+                                    <value>1</value>
```
manovotn, line 510: Does this level of verbosity have any specific value for us (as opposed to default 0?)

jboss-tck-runner/pom.xml
```diff
@@ -417,22 +501,33 @@
...
+                                <arquillian.launch>wildfly-23</arquillian.launch>
                                 <!-- Specified system properties are not
                                     available during in-container test execution -->
                                 <org.jboss.cdi.tck.libraryDirectory>target/dependency/lib</org.jboss.cdi.tck.libraryDirectory>
                                 <org.jboss.cdi.tck.testDataSource>java:jboss/datasources/ExampleDS</org.jboss.cdi.tck.testDataSource>
                             </systemPropertyVariables>
                             <systemProperties>
                                 <jacoco.agent>${jacoco.agent}</jacoco.agent>
-                                <additional.vm.args>${additional.vm.args}</additional.vm.args>
+                                <additional.vm.args>-Dtest=org.jboss.cdi.tck.tests.deployment.packaging.installedLibrary.InstalledLibraryEarTest</additional.vm.args>
```
manovotn, line 524: Why do you keep the `-Dtest` here? You want to run all tests.

jboss-tck-runner/pom.xml
```diff
@@ -417,22 +501,33 @@
...
                                 <illegal.access.option>${surefire.plugin.jdk9.args}</illegal.access.option>
+                                <test>org.jboss.cdi.tck.tests.deployment.packaging.installedLibrary.InstalledLibraryEarTest</test>
```
manovotn, line 526: Similar as above, this is probably a leftover from your testing?

pom.xml
```diff
@@ -86,7 +87,7 @@
         <junit.version>4.13.2</junit.version>
         <log4j.version>1.2.17</log4j.version>
         <shrinkwrap.version>1.2.6</shrinkwrap.version>
-        <shrinkwrap.descriptors.version>1.1.0-beta-1</shrinkwrap.descriptors.version>
+        <shrinkwrap.descriptors.version>2.0.0</shrinkwrap.descriptors.version>
```
manovotn, line 90: I think even this version is still not jakarta-friendly (and the project is pretty much dead). But that's something we will need to eventually solve on the TCK level.

jboss-tck-runner/pom.xml
```diff
@@ -133,6 +135,12 @@
             <artifactId>jakarta.faces-api</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
```
manovotn, line 140: Isn't this something that should be inherited from TCKs if they require it?

manovotn (changes requested):
Dropped some comments.

jboss-tck-runner/pom.xml
starksm64: The problem was that weld was using a much newer version of htmlunit than the TCK was, so it had been overriden to 2.9 at this level previously. I was expecting that these versions might diverge again in the future and an explicit override would still be needed here.

pom.xml
starksm64: 2.0.0 is the current version of the TCK, but if it needs to be removed from the TCK I can do that.

jboss-tck-runner/pom.xml
starksm64: There is a problem with testng 7.4.0 that when you go above 1 in the verbose level and a test fails in a certain way, the verbose logging listener starts throwing NPEs and makes all output useless. I meant to add a comment here about 1 being the highest level and then was going to look at creating a testng patch to fix the issue. Either that, or the TCK needs to revert to 7.3.0.

pom.xml
manovotn: We have no replacement ATM. Even if it creates outdates descriptors (w/o jakarta namespace) it still works.
The only other way is to go and create all the descriptions manually under TCK which I don't think is worth it.
So, let's just keep it this way for now.

manovotn:
@starksm64 Seeing all those JDK 8 jobs fail in CI (`class file has wrong version 55.0, should be 52.0`) - am I right to assume we can count on JDK 11+ being used for CDI 4?

starksm64:
Yes, the baseline for anything in EE10 is going to be Java SE 11+

manovotn:
These commits are now squashed into https://github.com/weld/core/pull/2383 as part of the work to make master work with latest CDI APIs.
Closing this PR.

PR #2380, opened 2021-06-02
[WELD-2674] NPE when deploying an EJB bean in default package

spyrkob:
Issue: https://issues.redhat.com/browse/WELD-2674

### DIFF
--- a/impl/src/main/java/org/jboss/weld/util/Proxies.java
+++ b/impl/src/main/java/org/jboss/weld/util/Proxies.java
@@ -58,6 +58,8 @@ public class Proxies {
 
     public static class TypeInfo {
 
+        private static final String DEFAULT_PACKAGE = "";
+
         private final List<Class<?>> interfaces;
         private final List<Class<?>> classes;
         private final Map<String, String> classToPackageMap;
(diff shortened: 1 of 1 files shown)


impl/src/main/java/org/jboss/weld/util/Proxies.java
```diff
@@ -81,7 +81,7 @@ private TypeInfo(Set<? extends Type> types) {
         private TypeInfo add(Type type, List<Class<?>> foundInterfaces, List<Class<?>> foundClasses, Map<String, String> classToPackageMap) {
             if (type instanceof Class<?>) {
                 Class<?> clazz = (Class<?>) type;
-                classToPackageMap.put(clazz.getName(), clazz.getPackage().getName());
+                classToPackageMap.put(clazz.getName(), clazz.getPackage()==null?null:clazz.getPackage().getName());
```
manovotn, line 84: While this should be technically safe (since the package should then be replaced), I'd rather we store empty String in the map instead of just `null`.

manovotn (changes requested):
Just one small comment, otherwise this looks good.

manovotn (approved):
Awesome, thanks for contributing!
I'll merge it once CI run passes.

manovotn:
@spyrkob thanks for contributing the fix!

PR #2383, opened 2021-06-07
Update default branch to new version, add CDI 4.0 bits

manovotn:
~~*WORK IN PROGRESS*~~
**This PR should now be ready for review.**

This PR will contain necessary adjustments to switch the default branch to CDI 4.0.
Basically, I would like to get the PR to a buildable/testable state and from there we can once more do once PR at a time.

TODO list for this PR:

- [x] CI needs to use JDK 11+ (probably JDK 11 + JDK 16?); JDK 8 will no longer be supported
- [x] Update Weld API to a released Alpha for 5.x (currently uses snapshot, see https://github.com/weld/api/pull/131)
- [x] Implement `BeanContainer` within `BeanManagerImpl` and remove deprecated methods
- [x] Several more items were deprecated in CDI 4.0, check for usage in Weld
- [ ]  ~~CDI 4.0 aims to redefine bean discovery mode for empty `beans.xml`; we need to adapt and see what tests will be broken by this change~~
  - Created a tracking issue to include this later - https://issues.redhat.com/browse/WELD-2677 
- [x] Include https://github.com/weld/core/pull/2379 as part of this PR
- [ ] ~~Decide on how we include support for build compatible extensions; e.g. how we include https://github.com/Ladicek/StillDI/tree/main/impl inside Weld~~
  - Created a tracking issue to include this later - https://issues.redhat.com/browse/WELD-2678
- [x] This PR will also require new version of Arquillian tooling because they are using removed parts of API
  - [x] Arq. Weld container needs an update to remove usage of deprecated APIs - see https://github.com/arquillian/arquillian-container-weld/issues/103
  - [x] Actually, even Arq. core will need an update since it uses deprecated APIs - see https://github.com/arquillian/arquillian-core/issues/338
- [x] Due to CDI API changes, CDI TCK will need to be updated to remove deprecated parts and released, see https://github.com/eclipse-ee4j/cdi-tck/issues/239
- [x] Change CI to exclude all incontainer testing because the current WFLY cannot be easily patched - this is due to other specs using now removed deprecated API parts


inject-tck-runner/src/test/java/org/jboss/weld/atinject/tck/Producers.java
```diff
@@ -63,8 +65,10 @@ public Seat produceAtDriversSeat(@New DriversSeat driversSeat) {
      */
     @Produces
     @Typed(DriversSeat.class)
-    public DriversSeat produceDriversSeat(@New DriversSeat driversSeat) {
-        return driversSeat;
+    public DriversSeat produceDriversSeat(BeanManager manager) {
+        // TODO this used to use @New to create the instance, do this in a cleaner way
```
manovotn, line 69: I am not even sure how this worked with `@New` in the first place since this was basically a circular dependency.

While the `Unmanaged` approach here would probably work, it is meant just as a hack to get a compile-able code.
It needs to be rewritten to something reasonable - maybe creating the instance via BM and injecting into it or an extension.
Same scenario applies on L#100 of this class.

I'd be curious to hear how this worked and what  you suggest instead @mkouba :)

manovotn:
Work done in https://github.com/weld/core/pull/2379 is now incorporated here.

manovotn:
With the current PR state, if I update to snapshot version of:
* CDI TCK
* Arq. core
* Arq. weld container

I can locally run build with tests (without container) and it all passes.
Incontainer tests are trickier because several other EE components are using now removed CDI APIs which crashes tests on WFLY.
I have to look into how much effort it is to make it work with WFLY and what EE components are affected - first that I bumped into was JSF.

manovotn:
So, after quite some fighting, I managed to run TCKs against WFLY as well.
It requires a custom jsf-impl as well as hibernate-validator artifacts as both of them currently deprecated CDI APIs. I have submitted PRs to both specs to let them know of these changes.

That being said, with all the snapshots needed, I don't think it is worth the hassle to try and setup CI so that it can run in-container tests. I will probably just adapt it so that we run all tests that don't need a container which should give us enough confidence for the time being.

And for that we need:
* Arq. core release
* Followed by Arq. weld container release
* CDI TCK release

manovotn:
Almost there - incontainer tests are disabled until we at least have a WFLY version that can use bumped CDI API version. After that, we can patch the EE 9 WFLY to use newer CDI and it should work.

Two more items from the original list remain:
* Change of default behavior of `beans.xml`
* Incorporating support for build compatible extensions

Both of the above are IMO good candidates for separate JIRAs and I wouldn't want to delay this PR any further.
Other than that, there are JDK 16 tests failing that need investigation.

PR #2413, opened 2021-07-13
WELD-2660 - Remove WFLY patch creation utils from code.

manovotn:
Removal of patch creation tooling for Weld 3.1.x


manovotn:
Those juicy CI errors seem are coming from WFLY CI (where we download the server from) and not even manual attempts to download it will work. If not fixed later today, I will contact someone in WFLY team.

PR #2492, opened 2021-09-06
WELD-2677 - Implement a backward compatibility option that allows to …

manovotn:
…treat empty beans.xml as discovery mode ALL.

EDIT: note that TCK tests will currently blow up as we are running with the empty beans.xml changes against a TCK that wasn't yet adapted.

### DIFF
--- a/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
+++ b/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
@@ -32,6 +32,7 @@ import java.util.Set;
 import jakarta.annotation.Priority;
 
 import org.jboss.weld.bootstrap.api.Bootstrap;
+import org.jboss.weld.bootstrap.spi.BeanDiscoveryMode;
 import org.jboss.weld.bootstrap.spi.BeansXml;
 import org.jboss.weld.bootstrap.spi.Metadata;
 import org.jboss.weld.environment.deployment.WeldBeanDeploymentArchive;
@@ -62,8 +63,11 @@ public abstract class AbstractDiscoveryStrategy implements DiscoveryStrategy {
 
     private final List<BeanArchiveHandler> handlers;
 
+    private final BeanDiscoveryMode emptyBeansXmlDiscoveryMode;
+
     protected AbstractDiscoveryStrategy() {
         handlers = new LinkedList<BeanArchiveHandler>();
+        this.emptyBeansXmlDiscoveryMode = BeanDiscoveryMode.ANNOTATED;
     }
 
     /**
@@ -71,12 +75,16 @@ public abstract class AbstractDiscoveryStrategy implements DiscoveryStrategy {
      * @param resourceLoader
      * @param bootstrap
      * @param initialBeanDefiningAnnotations
+     * @param emptyBeansXmlDiscoveryMode
      */
-    public AbstractDiscoveryStrategy(ResourceLoader resourceLoader, Bootstrap bootstrap, Set<Class<? extends Annotation>> initialBeanDefiningAnnotations) {
+    public AbstractDiscoveryStrategy(ResourceLoader resourceLoader, Bootstrap bootstrap,
+                                     Set<Class<? extends Annotation>> initialBeanDefiningAnnotations,
+                                     BeanDiscoveryMode emptyBeansXmlDiscoveryMode) {
         handlers = new LinkedList<BeanArchiveHandler>();
         setResourceLoader(resourceLoader);
         setBootstrap(bootstrap);
         setInitialBeanDefiningAnnotations(initialBeanDefiningAnnotations);
+        this.emptyBeansXmlDiscoveryMode = emptyBeansXmlDiscoveryMode;
     }
 
     @Override
@@ -102,7 +110,7 @@ public abstract class AbstractDiscoveryStrategy implements DiscoveryStrategy {
     public Set<WeldBeanDeploymentArchive> performDiscovery() {
 
         if (scanner == null) {
-            scanner = new DefaultBeanArchiveScanner(resourceLoader, bootstrap);
+            scanner = new DefaultBeanArchiveScanner(resourceLoader, bootstrap, emptyBeansXmlDiscoveryMode);
         }
 
         final List<BeanArchiveBuilder> beanArchiveBuilders = new ArrayList<BeanArchiveBuilder>();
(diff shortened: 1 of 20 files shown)


manovotn:
@mkouba please review this when you have a moment.

Few notes:
* This option should work in all environment (SE, EE, servlet) but users will always have to find a way to pass in a system property.
  * I don't really know of better way to achieve this across environments
* The test I wrote is for SE, I am not sure if we can test this with any test within `tests-arquillian` as those start Weld before we can hope to set up any property from within test class
  * Furthermore, EE testing won't currently fly anyway as there is no EE 10 container to start with :-/

impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java
```diff
@@ -136,6 +136,18 @@
...
+     * Standard behavior is that empty {@code beans.xml} is treated as discovery mode {@code annotated}.
+     * This option allows to change the behavor to discovery mode {@code all} which is how it used to work prior to
+     * CDI 4.0.
+     * <p/>
+     * Note that this option is temporary and will eventually be removed.
+     * <p/>
+     * This option can only be set via system config properties.
+     */
+    @Description("By default, empty beans.xml are treated as discovery mode annotated.")
+        LEGACY_EMPTY_BEANS_XML("org.jboss.weld.discovery.emptyBeansXmlModeAll", false),
```
Ladicek, line 149: Indentation :-)

mkouba:
Does it really have to be a system property? For weld-se the `Weld.property()` method can be used, or? And in WildFly, a value specified in a config xml file could be used, or?

manovotn:
@mkouba well, it was you who pointed me to that config when I last asked :D 
We need this information before we start creating any deployments and we need to cover SE, EE and servlet.
And if it is not a system property, then we will have to do something akin to what I described [in this JIRA comment](https://issues.redhat.com/browse/WELD-2677?focusedCommentId=18807871&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18807871).
We'd basically have to have three different ways to declare this property (SE `Weld.property()`, servlet context param and WFLY subsystem config) which then all have to somehow proliferate down to `BeansXmlStreamParser`.

I am I am not against it, just discussing options :)

mkouba:
> well, it was you who pointed me to that config when I last asked

Well, we could have both - an enum value in the `ConfigurationKey` and a "special handling", e.g. something like https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/config/ConfigurationKey.java#L275 and https://github.com/weld/core/blob/master/environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java#L228. But I agree that it's not very nice and definitely an additional maintenance burden. It might be worth the effort in the WildFly though...

manovotn:
I will try to adapt this PR so that:
* Weld SE can use its properties and we can then check their presence via `Weld.isEnabled()` (which checks the property map and has fallback for system property checking
* WFLY should be able to use its subsystem configuration
  * We cannot currently contribute this as a PR because it is EE 10 change and it's too early for that; we'll have to have a tracking issue for it 
* Servlet is of minor importance and will be resolved ad hoc if needed, the system property should still be an option there

manovotn:
@mkouba I have reworked this PR. It now requires API modifications from https://github.com/weld/api/pull/138 and removes the system property approach, Instead, it allows to set the config option via Weld SE property and in servlet via context param (this is untested ATM).

Is this what you had in mind? If so, I'll polish this and make a draft PR for what is needed on WFLY side.

manovotn:
Added one more minor change needed for WFLY.
On WFLY side, I think we will need something like this - https://github.com/wildfly/wildfly/compare/main...manovotn:weld2677?expand=1
The solution is untested ATM due to inability to run WFLY with EE 10 dependencies but at least we will have a PoC draft once they get there.

mkouba:
> Is this what you had in mind?

Yes, exactly. But maybe we could change the `boolean`  flag to the default `BeanDiscoveryMode` used for empty `beans.xml`. My problem with `legacyEmptyBeansXmlTreatment` is that you need to know what "legacy" means in order to immediately understand how to use it...

impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
```diff
@@ -44,17 +44,28 @@
...
 
     private final BeansXmlValidator beansXmlValidator;
+    // having this information here allows WFLY to create appropriate parser
+    private final BeanDiscoveryMode emptyBeansXmlDiscoveryMode;
 
     public BeansXmlParser() {
+        this(false);
+    }
+
+    public BeansXmlParser(boolean legacyEmptyBeansXmlTreatment) {
```
manovotn, line 54: I left this as boolean since it is coming from WFLY but it could be changed as well.

manovotn:
@mkouba updated per suggestion

environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/AbstractDiscoveryStrategy.java
```diff
@@ -62,6 +63,8 @@
 
     private final List<BeanArchiveHandler> handlers;
 
+    private BeanDiscoveryMode emptyBeansXmlDiscoveryMode = BeanDiscoveryMode.ANNOTATED;
```
mkouba, line 66: `private final`?

environments/se/core/src/main/java/org/jboss/weld/environment/se/Weld.java
```diff
@@ -199,6 +199,15 @@
...
     public static final String DEV_MODE_SYSTEM_PROPERTY = "org.jboss.weld.development";
 
+    /**
+     * Standard behavior is that empty {@code beans.xml} is treated as discovery mode {@code annotated}.
+     * This configuration property allows to change the behavior to discovery mode {@code all} which is how it used to work prior to
+     * CDI 4.0.
+     * <p/>
+     * Note that this option is temporary and servers to easy migration. As such, it will be eventually removed.
+     */
+    public static final String LEGACY_EMPTY_BEANS_XML_TREATMENT = "org.jboss.weld.se.discovery.emptyBeansXmlModeAll";
```
mkouba, line 209: I like the name used in the key more, e.g. something like `EMPTY_BEANS_XML_DISCOVERY_MODE_ALL`?

impl/src/main/java/org/jboss/weld/xml/BeansXmlParser.java
mkouba: `legacyEmptyBeansXmlTreatment` -> `emptyBeansXmlDiscoveryModeAll`?

manovotn:
Updated. I will get to releasing API and upgrading it here some time later today or tomorrow based on other workload.

manovotn:
Weld API is released and updated in this PR.

manovotn:
FTR - I have created [WELD-2680](https://issues.redhat.com/browse/WELD-2680) to keep track of the changes needed on WFLY side once we can contribute them.

PR #2495, opened 2021-09-27
Embedded TCK execution should use suite XML file.

manovotn:
Fixes an issue where TCK tests that weren't properly named weren't executed in embedded container. See https://github.com/eclipse-ee4j/cdi-tck/issues/273#issuecomment-927090230


manovotn:
Also updated TCK to newly released version which should, at last, be rewritten to acknowledge empty `beans.xml` changes.

manovotn:
Yay, we are back to green builds!

PR #2529, opened 2021-10-27
WELD-2685 Implement Instance.Handle interface.

manovotn:
Requires new Weld API release with updated CDI API version.
See also https://github.com/weld/api/pull/140

### DIFF
--- a/tests-arquillian/src/test/java/org/jboss/weld/tests/instance/enhanced/InstanceHandleTest.java
+++ b/tests-arquillian/src/test/java/org/jboss/weld/tests/instance/enhanced/InstanceHandleTest.java
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source
- * Copyright 2016, Red Hat, Inc., and individual contributors
+ * Copyright 2021, Red Hat, Inc., and individual contributors
  * by the @authors tag. See the copyright.txt in the distribution for a
  * full listing of individual contributors.
  *
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.jboss.weld.tests.instance.enhanced;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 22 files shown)


tests-arquillian/src/test/java/org/jboss/weld/tests/instance/enhanced/InstanceHandleTest.java
```diff
@@ -0,0 +1,129 @@
...
+        assertEquals(1, sequence.size());
+        assertEquals(alpha2Id, sequence.get(0));
+
+        alpha1.destroy();
+        // Subsequent invocations are no-op
+        alpha1.destroy();
+
+        sequence = ActionSequence.getSequenceData();
+        assertEquals(2, sequence.size());
+        assertEquals(alpha1.get().getId(), sequence.get(1));
```
manovotn, line 92: This is actually wrong - we cannot assert the ID at this point as the instance was already destroyed. We need to change the behavior to instead throw an exception when you perform `Handle.get()` in a `Handle` that destroyed its reference. Plus we need to test that (in both test variants).

tests-arquillian/src/test/java/org/jboss/weld/tests/instance/enhanced/WeldInstanceTest.java
```diff
@@ -50,20 +49,22 @@
...
+    public void testIsResolvable(WeldClient client) {
         ActionSequence.reset();
         assertNotNull(client);
         assertTrue(client.getAlphaInstance().isResolvable());
         assertFalse(client.getBigDecimalInstance().isResolvable());
     }
 
     @Test
-    public void testGetHandler(Client client, BeanManager beanManager) {
+    public void testGetHandler(WeldClient client, BeanManager beanManager) {
```
manovotn, line 67: Same as the other comment, this needs changing.

manovotn (commented):
Found an issue with the behavior and testing; I'll fix it later today.

mkouba (commented):
Hm, there is also `org.jboss.weld.environment.se.test.instance.enhanced.WeldInstanceTest` that needs to be fixed.

manovotn:
@mkouba yeah I noticed, already fixing that locally, will be pushed soon as I get green test result :)
It is actually the only test that also asserts on the ordering so I am creating a dual variant that will test both, Weld and CDI APIs.

manovotn:
@mkouba should be all good now

mkouba (approved):
Looks good.

PR #2559, opened 2021-11-04
WELD-2688 Allow usage of @Priority on stereotypes, add tests.

manovotn:
Allows usage of `@Priority` on stereotypes - implements a recursive search of stereotypes for priority annotation in order to register globally enabled alternatives.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/BeanDeployer.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/BeanDeployer.java
@@ -22,7 +22,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import jakarta.annotation.Priority;
 import jakarta.decorator.Decorator;
+import jakarta.enterprise.inject.Stereotype;
 import jakarta.enterprise.inject.spi.AnnotatedType;
 import jakarta.enterprise.inject.spi.Bean;
 import jakarta.enterprise.inject.spi.Extension;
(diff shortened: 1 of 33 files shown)


impl/src/main/java/org/jboss/weld/logging/BootstrapLogger.java
```diff
@@ -342,4 +342,7 @@
     @Message(id = 182, value= "Provided implementation of org.jboss.weld.serialization.spi.ProxyServices ({0}) does not support class defining. This functionality is required in order to be able to define proxy classes.", format = Format.MESSAGE_FORMAT)
     IllegalStateException proxyServicesWithoutClassDefining(Object services);
 
+    @Message(id = 183, value= "Multiple different @Priority values derived from stereotype annotations for annotated type - {0}", format = Format.MESSAGE_FORMAT)
+    DeploymentException multiplePriorityValuesDeclared(Object annotatedType);
```
Ladicek, line 346: It should be a definition error, not a deployment problem, I think?

Ladicek: (I'm not exactly sure what's the difference. Feel free to ignore :rofl:)

manovotn: Good point, thanks!
Fixed.

mkouba: A definition error may be usually detected at development time, e.g. by tooling. Deployment problems are detected at initialization time, i.e. during bootstrap.

impl/src/main/java/org/jboss/weld/bootstrap/BeanDeployer.java
```diff
@@ -98,22 +100,77 @@ public BeanDeployer addClass(Class<?> clazz, AnnotatedTypeLoader loader) {
         return ctx;
     }
 
+    /**
+     * Attempts to find a {@code @Priority} annotation declared on a stereotype of given {@code AnnotatedType}.
+     *
+     * If there is multiple {@code @Priority} annotations within hierarchy of stereotypes, the first found is used.
```
mkouba, line 106: Hm, shouldn't both situations result in a deployment exception? What do we do if multiple interceptor bindings of the same type but with different member values are applied? I think it's an error, or?

manovotn: Maybe I poorly worded it, what I meant by it is a single stereotype which might have several priorities in its hierarchy. Imagine you have two stereotypes:

```
@Stereotype
@Priority(200)
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
public @interface MyStereotype1
```

```
@Stereotype
@Priority(100)
@MyStereotype1 // declares another stereotype which has priority
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
public @interface MyStereotype2
```

And then you have a bean such as:

```
@Dependent
@Alternative
@MyStereotype2
public class Foo
```

Since `@Priority` is a standard non-repeatable annotation, I think it would be natural for `MyStereotype2` to "override" the value that `MyStereotype1` brings in. Besides, it allows us to end the recursive search sooner.
As for specification, I don't think we say anything about this case in particular (correct me if I am wrong @Ladicek).It is basically the same as for multiple scopes on stereotype. However, in the TCKs I've added a test that works this way - see https://github.com/eclipse-ee4j/cdi-tck/pull/311.
I am not opposed to changing it, this is just what made sense to me.

Ladicek: Isn't this the exact situation we discussed privately yesterday? My _feeling_ is that this situation should result in an error, but I'll admit that your reasoning makes a lot of sense.

mkouba: (1) Yes, in this particular case it makes sense. But what about this use case:
```java
@Stereotype
@MyStereotype11
public @interface MyStereotype1 {}

@Stereotype
@Priority(1000)
public @interface MyStereotype11 {}

@Stereotype
@MyStereotype21
public @interface MyStereotype2 {}

@Stereotype
@Priority(2000)
public @interface MyStereotype21 {}

@MyStereotype1
@MyStereotype2
public class Foo {}
```
An error, right?

(2) What do you think about the consistency with interceptor bindings?

manovotn: @mkouba 

> (1) Yes, in this particular case it makes sense. But what about this use case:

Ok, I have reworded the javadoc to try and make it clearer
Your presented case will be an error. This is specified and there is a test for it - `ConflictingPriorityStereotypesTest`. Albeit without inheritance but the result is the same.

> (2) What do you think about the consistency with interceptor bindings?

What do you mean?

manovotn: @Ladicek yes, that is what we discussed and this is what I arrived at later on when I gave it some more thought :-)

mkouba: > What do you mean?

Well, for interceptor bindings a similar use case should IMO result in an error because the spec states: _"If the set of interceptor bindings of a component class or interceptor, including bindings inherited from stereotypes and other interceptor bindings, has two instances of a certain interceptor binding type and the instances have different values of some annotation member, the container automatically detects the problem and treats it as a definition error."_

manovotn: Fair point, I haven't tested it but the wording is pretty clear.

However, I am not fully convinced that `@Priority` and interceptor bindings with values are so similar they should be treated the same way.

manovotn:
I have added a commit that introduced two tests for inherited stereotypes - one that should succeed and one that introduces two inherited and conflicting stereotypes. I'll add the same to the TCK PR.

manovotn:
@mkouba @Ladicek this PR is now modified so that the scenario we discussed results in a `DefinitionException` because it has been well specified via https://github.com/eclipse-ee4j/cdi/pull/557.
I have adapted the implementation accordingly and added a test for this case.

Ladicek (approved):
Took me a little while to realize that CDI disallows transforming annotations on annotations and hence it's fine to look directly at the `Class` objects :-)

PR #2560, opened 2021-11-07
WELD-2690 Replace usages of getParameterTypes().length with getParame…

clara0:
…terCount()

https://issues.redhat.com/browse/WELD-2690


manovotn:
I see no reason not to accept this change. However, I wonder what is so devilish about `getParameterTypes().length` that it needs changing throughout Weld and whole WFLY @clara0? :-)

manovotn (approved):

clara0:
Hi @manovotn, `getParameterTypes().length` ends up creating unnecessary clones, which take up memory and adds to the burden of the GC. Furthermore, replacing it with `getParameterCount()` increases the readability.

PR #2563, opened 2021-11-12
WELD-2692 Execute and adjust tests-integration for incontainer testin…

manovotn:
…g with WFLY.


manovotn:
Note that the CI won't be able to check this yet because there is no easy way to automatically obtain WFLY with CDI for EE 10 - I had to glue that together locally.

PR #2564, opened 2021-11-18
WELD-2605 Ignore non-existing class path entries

Vampire:

PR #2565, opened 2021-11-18
Do not include same class path entries from class path property and manifest entries

Vampire:


manovotn:
@Vampire this commit could probably be appended to https://github.com/weld/core/pull/2564?

Vampire:
I would have rebased it after that was merged, but I can of course also incorporate it, however you like

manovotn:
@Vampire incorporate it please, it is very much related to fixing the same class and will be easier for future tracking :)

Vampire:
Done

PR #2568, opened 2021-11-24
WELD-2678 Add implementation of Lite extension translator

manovotn:
A draft of the PR needed to include support for build compatible extensions.
First commit is just adding the impl as-is from Ladislav's StillDI implementation.
Next ones are minor adjustments are integration into SE and Servlet in order to allow testing of TCKs via embedded container.

I will keep adding more modifications as there are multiple TODOs in the impl.
From the top of my head, we'll also need to:
* Change logging to jboss logging (which is what rest of Weld code uses)
* Remove any deprecated methods used in impl (ATM I know of `Class.newInstance()`)
* Use security actions for any operation with CL and possibly others place?
* ?? probably more things that I will update as I go...

### DIFF
--- a/environments/se/core/pom.xml
+++ b/environments/se/core/pom.xml
@@ -45,6 +45,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.jboss.weld</groupId>
+            <artifactId>weld-lite-extension-translator</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.jboss.weld.probe</groupId>
             <artifactId>weld-probe-core</artifactId>
         </dependency>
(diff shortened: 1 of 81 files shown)


manovotn:
I think the way to do it is to quickly accept these changes and cut a release to have Alpha out.
Then we can also get a release of Arq. container (see https://github.com/arquillian/arquillian-container-weld/issues/113) which is needed to be able to test Lite extensions in embedded mode.

We should create a new JIRA and track the changes needed for the adaptation of StillDI.

manovotn:
I have created https://issues.redhat.com/browse/WELD-2694 to track further improvements to this module and I have marked that for the next release.
I am aiming to merge this and stage Alpha1 version so that we can improve Arq. container as well.

PR #2605, opened 2021-12-07
Fix how Lite extension translator distinguishes synthetic observers

Ladicek:


manovotn (commented):
Since getting a `Bean` from OM is now supported, we'll also need to implement it under `ObserverInfoImpl`, see https://github.com/weld/core/blob/master/weld-lite-extension-translator/src/main/java/org/jboss/weld/lite/extension/translator/ObserverInfoImpl.java#L77-L78

We can either do it in a temporary way (casting it impl-specific class), or wait until CDI API releases. I'd say we do it in Weld specific way for now so that we can get this merged.

Ladicek:
I'd say this PR itself can be merged regardless, because it doesn't touch that code. Otherwise, I don't really mind.

PR #2606, opened 2021-12-08
Adapt the Lite extensions translator to CDI 4.0.0.Beta3 changes

Ladicek:


weld-lite-extension-translator/src/main/java/org/jboss/weld/lite/extension/translator/ExtensionPhaseEnhancement.java
```diff
@@ -38,11 +39,11 @@ void runExtensionMethod(java.lang.reflect.Method method) {
             String errorMsg = " of type ClassInfo, MethodInfo, FieldInfo, ClassConfig, MethodConfig," +
                     " or FieldConfig for method " + method + " @ " + method.getDeclaringClass();
             if (numQueryParameters == 0) {
-                throw new IllegalArgumentException("No parameter" + errorMsg);
+                throw new DefinitionException("No parameter" + errorMsg);
```
manovotn, line 42: Interestingly, even when this was throwing IAE, the newly added TCK tests passed. I suppose there is some extension transformer in play...

Ladicek: I did the same change in ArC, so I figured it would make sense here too. I guess Weld handles exceptions thrown from extensions in some clever way, or there's an exception transformer indeed.

manovotn: I totally agree, I was just surprised that this exception is wrapped somehow :)

PR #2607, opened 2021-12-09
Update lite extension translator

manovotn:
Updates Weld API to new version.
Updates Lite extension translator to be able to initialize it with `ClassLoader` which is needed for WFLY integration.
Updated `jboss-as/pom.xml` to be able to update extension translator version when swapping out artifacts inside WFLY dist.


manovotn:
This build is going to fail until I manage to release Weld Parent and yet another Weld API version.
I'll bump the CI once that is done.

PR #2617, opened 2021-12-12
WELD-2699 Add support for execution of CDI lang model TCKs.

manovotn:


manovotn:
Hmm, realized that I cannot really add it into `weld-lite-extension-translator` module as it would create a circular dep if used with Weld embedded Arq. container but also with just pure Weld SE as they both depend on this module.

We'll need to either use something similar to atinject TCK setup to avoid circularity, or move it into some other module (existing testing module or new).

manovotn:
I've added it as a separate module that executes any time tests are executed during build (for instance, as part of plain `mvn clean install` on the project). This way we should keep track of whether we are compliant via every PR sent.
Test itself is done via embedded Arq. container and one build compatible extension.

manovotn:
Ooof, there is a failure with JDK 16 but not with JDK 11.
It's coming from this line - https://github.com/eclipse-ee4j/cdi-tck/blob/master/lang-model/src/main/java/org/jboss/cdi/lang/model/tck/AnnotatedTypes.java#L633

Apparently, with JDK 11 reflection won't see any annotation whereas with 16, it does.

manovotn:
FTR, I have created a TCK issue to address these failures - https://github.com/eclipse-ee4j/cdi-tck/issues/330

PR #2618, opened 2021-12-14
Fix special handling of constructor params for enums with JDK 17.

manovotn:
I found out that reflection access to parameters of enum constructor is different in JDK 11 and 17.
We have a hack in place for Lang model to account for JDK 11 but this doesn't work with JDK 17.

The difference lays in JDK 11 seeing any user-defined parameter annotations as belonging to the first *synthetic* parameter whereas JDK 17 correctly attributes those annotations to user-defined param. It also correctly accounts for synth parameters when invoking `getAnnotatedParameters()` for the constructor method.

I have cobbled together what worked for me but wanted to ask @Ladicek for a review.


Ladicek:
If it's fixed in later JDK, that's great! I even had to add a special case to the lang model TCK for JDK <15, this is just so much pain.

If this passes, then it's good.

manovotn:
> If it's fixed in later JDK, that's great! I even had to add a special case to the lang model TCK for JDK <15, this is just so much pain.
> 
> If this passes, then it's good.

This does pass for me locally :-)
Although we still need to have the hack in place because of the synth parameters and because of compat with anything lower than JDK 17.

As for special cases with other JDKs, I sadly found another as described in https://github.com/eclipse-ee4j/cdi-tck/pull/331

PR #2634, opened 2022-01-04
WELD-2695 Create smoke test for coexistence of portable extensions with build c…

manovotn:
…ompatible extensions.

### DIFF
--- a/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/extension/build/compatible/OverridenBuildCompatibleExtension.java
+++ b/environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/extension/build/compatible/OverridenBuildCompatibleExtension.java
@@ -1,0 +1,63 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2022, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


manovotn:
Note that these are just smoke tests and are deliberately running in SE. This is because ATM we rely on CDI TCK to contain tests verifying actual functionality of BCE and those tests are executed via Arq. embedded container where the inclusion of `weld-lite-extension-translator` is done slightly differently.

manovotn:
@Ladicek BTW writing this I realized we don't really have `@SkipIfPortableExtensionPresent` covered in TCKs. Might want to add it as a Full-only test in some simple form as well.

environments/se/tests/src/test/java/org/jboss/weld/environment/se/test/extension/build/compatible/OverridenBuildCompatibleExtension.java
```diff
@@ -0,0 +1,63 @@
...
+import jakarta.enterprise.lang.model.declarations.ClassInfo;
+
+/**
+ * This BCE is overriden and should never be invoked
+ */
+// TODO change to OverridingPortableExtension.class once https://github.com/eclipse-ee4j/cdi/issues/585 is resolved
+@SkipIfPortableExtensionPresent("org.jboss.weld.environment.se.test.extension.build.compatible.OverridingPortableExtension")
+public class OverridenBuildCompatibleExtension implements BuildCompatibleExtension {
+
+    public static int TIMES_INVOKED = 0;
```
mkouba, line 37: Just nitpicking over tiny details but are we 100% sure this static variable is only accessed from a single thread?

manovotn: Yea, we are executing all extensions in a synchronous manner from a single thread, so I'd say it's safe.

mkouba: If that's the case then we're good to go ;-)

mkouba (approved):

Ladicek (approved):
LGTM, I'd just rename `StandardBCE` to `StandardBuildCompatibleExtension`, because all other extensions have their names expanded :-)

PR #2648, opened 2022-01-12
WELD-2694 Introduce jboss logging into extension translator. Replace deprecated Class.newInstance() calls.

manovotn:


manovotn:
Added initial commit that removes deprecated calls to `Class.newInstance` and replaces them with `getDeclaredConstructor().newInstance()`.
Added jboss logging and replaced all exception throwing with calls to the newly added logger.

Note that any non-exceptional logging (such as `MessagesImpl`) is still a TODO that's coming in the next commit.

manovotn:
The failures were because checkstyle was checking generated logger files. By moving `Logger` under `logging` package, it passes our checkstyle suppression list and should pass.

manovotn:
Additional commits now:
* Make sure `MessagesImpl` now use JBoss logging as well.
* Remove the `BeanManagerAccess` hack and replace it with BM reference propagation throughout most classes.
* Suppress logging a warning for portable extension with `PAT<?>` in case it is `LiteExtensionTranslator`.


There is still few more thing to do. I am aware of two `setAccessible()` invocations which I think should be replaced by security actions.
Ideally we also want to have some detection mechanism that will allow us to not register `LiteExtensionTranslator` at all if there are no build compatible extensions found during bootstrap. At least for SE/servlet, that could be doable within this PR. WFLY will have to take it into consideration once it can be implemented there.

manovotn:
Next batch of commits fixes:
* `setAccessible()` methods
* Allows to detect any `BuildCompatibleExtension` impls before registering `LiteExtensionTranslator` to process them
* Addresses TODOs in this module

manovotn:
The PR now covers the scope I intended for it, hence I will turn it from draft to PR.

@mkouba, I going to have to ask you for a review here. The PR has several commits which might be easier to review that the PR as a whole.

Also CCing @Ladicek in case he wants to see how I mangled his work :-D

weld-lite-extension-translator/src/main/java/org/jboss/weld/lite/extension/translator/MessagesImpl.java
```diff
@@ -4,83 +4,81 @@
...
 
     private final String AT = " at ";
 
     private final SharedErrors errors;
-    private final Logger logger;
 
-    MessagesImpl(java.lang.reflect.Method method, SharedErrors errors) {
+    MessagesImpl(SharedErrors errors) {
         this.errors = errors;
-        this.logger = Logger.getLogger(method.getDeclaringClass().getName());
```
Ladicek, line 19: This is a little unfortunate. Previously, each BCExtension got its own logger category (the extension class), but with this change, they are all lumped together (under the translator package).

Ladicek:
Looks fine, added just 1 minor comment.

weld-lite-extension-translator/src/main/java/org/jboss/weld/lite/extension/translator/MessagesImpl.java
manovotn: @Ladicek Does that make much difference it terms of the message user gets? Do you think we should somehow preserve the notion of declaring class the logged message?

Ladicek: I don't think it's _that_ big of a deal, but AFAIK JBoss Logging would easily allow retaining the original behavior, which I'd find nice.

manovotn: Yea, I could leave this as it was but I wanted to align with the rest of the logging done in this module. Personally, I didn't see any difference in it...

manovotn:
Rebased onto master and applied Martin's suggestion WRT to thread safety of the extension loader class.

PR #2652, opened 2022-01-20
WELD-2704 Apply the change about Class.getPackage() to ProxyFactory.createProxyClass().

kaido207:
* Related to https://github.com/weld/core/pull/2264  

With JDK 11, Class.getPackage no longer returns null for classes in default packages.   
The following conditional branch also uses Class.getPackage() to determine the ProtectionDomain, so it is necessary to apply the above fix.  
https://github.com/weld/core/blob/f504fdf1cfc45a98d01c35e9ec294ec6ffa42ec9/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java#L489-L490

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
@@ -486,7 +486,7 @@ public class ProxyFactory<T> implements PrivilegedAction<T> {
 
         ProtectionDomain domain = AccessController.doPrivileged(new GetProtectionDomainAction(proxiedBeanType));
 
-        if (proxiedBeanType.getPackage() == null || proxiedBeanType.equals(Object.class)) {
+        if (proxiedBeanType.getPackage() == null || proxiedBeanType.getPackage().getName().isEmpty() || proxiedBeanType.equals(Object.class)) {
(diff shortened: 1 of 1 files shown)


manovotn:
Hi, @kaido207 and thanks for the PR.

Do  you have any specific test case where this was a problem?
What version of Weld did you encounter this with?

kaido207:
@manovotn thank you for your comment.
In general use, this can not be a problem, so it is difficult to give a specific test case but I think this change is necessary for compatibility.
Without this fix, the returned proxyClass with default package is created by the different ProtectionDomain(*) than JDK 8 or earlier.

(*) ProtectionDomain$CodeSource$locationNoFragString
  - JDK8 : file:///`glassfish version`/glassfish/modules/weld-osgi-bundle.jar
  - JDK11: file:///`glassfish version`/glassfish/domains/domain1/applications/`application name`/

// using Weld 3.0.0.Final on GlassFish 5.1.0 with JDK 11.

manovotn:
I see, in that case we'll need to make sure this PR goes to all active Weld versions.

I've created JIRA issue to track it - https://issues.redhat.com/browse/WELD-2704
I will pick it up from here and issue remaining PRs based on this one.

Thanks for pointing this out!

manovotn (approved):

PR #2662, opened 2022-02-01
Added support to read beans.xml and Jandex-Index from JRT-images

1tchy:
JRT-images are modular run-time images from JEP 220.

### DIFF
--- a/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/DefaultBeanArchiveScanner.java
+++ b/environments/common/src/main/java/org/jboss/weld/environment/deployment/discovery/DefaultBeanArchiveScanner.java
@@ -21,6 +21,7 @@ import static org.jboss.weld.environment.util.URLUtils.PROCOTOL_FILE;
 import static org.jboss.weld.environment.util.URLUtils.PROCOTOL_HTTP;
 import static org.jboss.weld.environment.util.URLUtils.PROCOTOL_HTTPS;
 import static org.jboss.weld.environment.util.URLUtils.PROCOTOL_JAR;
+import static org.jboss.weld.environment.util.URLUtils.PROCOTOL_JRT;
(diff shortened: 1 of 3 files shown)


manovotn:
@1tchy hello and thanks for the PR.

I have to read up on JEP-220 to understand the case but I have two questions anyway:
* What Weld version should support this? Since this PR is against `master` which is Weld 5/CDI 4
* Do you have any reproducer with which we could verify that its working? I don't suppose we will be able to easily create an automated test but I would still like to at least manually try it

1tchy:
I'm not too deep into this myself. I'm using it with Helidon and I unfortunatly did not fully brought it to work. But it did load the beans.xml and Jandex-index in my local project.
I opened the pull request in the hope it will help someone else, and under the assumtion, it will not break anything existing.
So to answer your actual questions:

- I have no opinion on a version.
- Unfortunatly I have no example project to share. And I support your opinion, that it would be difficult to write an automated test.

manovotn:
> I unfortunatly did not fully brought it to work.

If it's not yet ready, it might be better to turn this PR into a draft.

> I have no opinion on a version.

Alright, I think Weld 4 and 5 might make sense here.

> Unfortunatly I have no example project to share.

Ok, I have to admit that I am not to keen on accepting a PR that isn't fully working and that we don't have a test/use case for :)

I'll look into the JEP and see if I can come up with something.

1tchy:
Feel free to do whatever you want. 😀
**I think this is an improvement for Weld and not a risk at all.** But I'm unfortunatly not able to invest further. You can also close it, if you feel like it will not be merged as it is now and no one else wants to extend it.

manovotn:
Closing as stale.
There wasn't any other ask for this over the course of time and no scenario to test it with.

We can always reopen is need be.

PR #2712, opened 2022-03-21
WELD-2709 implement Startup and Shutdown event firing as per CDI 4.0 specification

manovotn:


impl/src/main/java/org/jboss/weld/bootstrap/WeldStartup.java
```diff
@@ -574,6 +576,7 @@ public void endInitialization() {
             for (BeanDeploymentModule module : modules) {
                 if (!module.isWebModule()) {
                     module.fireEvent(Object.class, ContextEvent.APPLICATION_INITIALIZED, Initialized.Literal.APPLICATION);
+                    module.fireEvent(Startup.class, new Startup(), Any.Literal.INSTANCE);
```
mkouba, line 579: Hm, would it make sense to fire the `Startup` event for all modules here? You don't need to fire the event from within  `HttpContextLifecycle` as you don't need to access the `ServletContext`... Actually, it might make sense to create an SPI so that an integrator can decide when to fire this event because otherwise I see no benefit compared to `@Initialized(ApplicationScoped.class)`. WDYT?

manovotn: > Hm, would it make sense to fire the Startup event for all modules here?

It should be fired for all modules - the `HttpContextLifecycle` contains the other part. Or did you mean something else?

> Actually, it might make sense to create an SPI so that an integrator can decide when to fire this event because otherwise I see no benefit compared to @Initialized(ApplicationScoped.class). WDYT?

I tend to disagree. For a runtime implementation such as Weld, it is very much identical to initialization of app scope and that's IMO fine.

These events were added to CDI for two reasons from what I can recall:
* For any build time implementation to allow them to differentiate initialization of app scope from state of actual container readiness as the two might differ. We already saw this in Quarkus.
* To create separate standard events that stand for the init and shutdown of container and don't require preliminary knowledge of other components to make use of them. I.e. you can now observe an event that directly represent container ready state as opposed to observing a context init event where you need to know that it happens as a first thing after container startup

mkouba: > It should be fired for all modules - the HttpContextLifecycle contains the other part. Or did you mean something else?

I thought that you don't need to fire this even from the `HttpContextLifecycle` but instead fire it from this loop outside the `if (!module.isWebModule())` block.

manovotn: Oh I see; sure, can do that.

tevans78: +1 for an SPI. The events need to be fired at slightly different times to the actual init and shutdown of the CDI container. Really these are "post-Startup" and "pre-Shutdown" events.

manovotn: @mkouba @tevans78 how do you imagine this SPI should look like?

Since it is a plain CDI event and integrators have the complete knowledge on when to fire it, do we really need an SPI to achieve that?

Ladicek: Technically we tried to formulate the spec of the `Startup` and `Shutdown` events in a way that it allows firing them at the same time as `@Initialized|BeforeDestroyed(ApplicationScoped.class) Object` for the purely runtime implementations like Weld. At the same time, I agree it would be great for integrators to be able to indicate to Weld that it should not fire these events, because the integrator will do it itself (ideally at a better point in time :-) ).

tevans78: @manovotn I was thinking so something really simple like `WeldBootstrap.fireStartupEvent() / fireShutdownEvent()`. However, If the web modules are not processed separately via the HttpContextLifecycle, and there is a way for us to prevent Weld from firing the events, then perhaps you are correct that we could fire these events independently. I will experiment with that.

Regardless of how it is done, getting the timing right could be difficult/unclear. Just because the CDI container has been started, should that guarantee that all other JEE services are ready and started... but no requests will be processed yet? On Shutdown, should everything else still be available... but no more requests be processed?

manovotn: > I was thinking so something really simple like `WeldBootstrap.fireStartupEvent()` / `fireShutdownEvent()`...

One thing I don't know from the top of my head is if you can fire that event and be able to observe it globally or whether you need to fire it in some per-module manner. I'd need to check tomorrow. Do let me know if you experiment with firing the event yourself please.

> Regardless of how it is done, getting the timing right could be difficult/unclear. Just because the CDI container has been started, should that guarantee that all other JEE services are ready and started... but no requests will be processed yet? On Shutdown, should everything else still be available... but no more requests be processed?

This is hard to answer because there are several perspectives. Note that CDI exists outside of JEE so this event makes sense even there. However, being able to fire this event yourself, you can use that to indicate whenever you think the container is ready (i.e. when other non-CDI services have also started?). Having it fired directly from Weld means we can only indicate when the CDI container started which isn't really that useful.

To me, the `Startup` event is ideally an entry point into the earliest time in the container lifecycle when you are safe to play around with all of its features. And this point can be very different in SE (where a bootstrap of CDI container is a job done) and in EE (where you also need to start other services and possibly only do that after CDI bootstraps).
However, this is not exactly what the javadoc states, it is purposefully very vague - https://github.com/eclipse-ee4j/cdi/blob/master/api/src/main/java/jakarta/enterprise/event/Startup.java

manovotn:
@mkouba @tevans78 I have updated the PR and would love to hear your thoughts on it.

There is now an API PR - https://github.com/weld/api/pull/155
And then this PR for Weld core.

The idea is that every implementation of `org.jboss.weld.bootstrap.api.Environment` can state whether they want Weld to automatically handle firing `Startup`/`Shutdown` events or not. By default, Weld keeps firing them because that way it is IMO smoother for all integrators (be it SE, Servlet or EE).
If integrators want to fire these events themselves, then that should be easily doable via `BeanManager.getEvent()`.

Side note - the PR build now fails because it requires new Weld API version (which is why I turned it into draft).

tevans78:
@manovotn I did experiment with this and I was able to fire the events myself. However, if I fired the Startup event for all modules at once then web modules received the Startup event _before_ `@Initialized(ApplicationScoped.class)`. While I could probably find a way to handle web modules differently, it would be more or less the same as how you coded it originally and does not leave me much room for changing the timing.

I agree with Emily that the use cases need to be clearer. Otherwise it is not possible to be sure that the spec is correct and that we have implemented it correctly.

manovotn:
> For the Shutdown event, since the CDI container is shuting down, the event might not be observed by the applications since CDI container is shutting down. 

This is why spec requires this event to happen prior to `@BeforeDestroyed` for application scope - all application scoped beans should be able to observe it still.

> Even though the event is observed, not much things the applications can perform at the containers are shutting down immeniently

I agree, this is the same as `@BeforeDestroyed` and `@Destroyed` events which have a very similar problem. The `Shutdown` is there mainly as a counterpart to `Startup` I guess. You could still do some cleanup, but not much in terms of beans and CDI as such.

> Can you clarify the use case you are trying to address?

I don't personally have a use case for this, I am just trying to implement the specification here :-D
I *guess* use cases are similar to those for `@BeforeDestroyed(ApplicationScoped.class)`?

Ladicek:
I'd just point out that the specification of `Startup` and `Shutdown` is [meant to be] vague enough to allow purely runtime implementations such as Weld to fire them at the same time as `@Initialized|BeforeDestroyed(ApplicationScoped.class) Object`. Assuming purely runtime implementations, these 2 events basically serve the same purpose (and the same set of use cases, so not sure why anyone wants to debate use cases again), except they can actually be intuitively understood by normal people.

(As @manovotn says, in certain build-time oriented implementations, application context initialization may happen during application build, and suddenly the difference matters. But I don't think that Weld or any of its consumers needs to care.)

manovotn:
Ok, at this point I think I will keep this solution because it does give some leeway to integrators but also provides a sensible default.
I have also issued a TCK PR with tests so once that is settles, I will do TCK RC followed by Weld API and core RC releases and so that we can restart the ballot ASAP.

Emily-Jiang:
> I'd just point out that the specification of `Startup` and `Shutdown` is [meant to be] vague enough to allow purely runtime implementations such as Weld to fire them at the same time as `@Initialized|BeforeDestroyed(ApplicationScoped.class) Object`. Assuming purely runtime implementations, these 2 events basically serve the same purpose (and the same set of use cases, so not sure why anyone wants to debate use cases again), except they can actually be intuitively understood by normal people.
> 
The confusion is that two events are firing for both startup and shutdown. I am not sure whether normal people can actually find easier to understand or even more confused. If we want to push for the new events, it might make more sense to deprecate the `@Initialized|BeforeDestroyed(ApplicationScoped.class) Object`.
> (As @manovotn says, in certain build-time oriented implementations, application context initialization may happen during application build, and suddenly the difference matters. But I don't think that Weld or any of its consumers needs to care.)

manovotn:
> If we want to push for the new events, it might make more sense to deprecate the @Initialized|BeforeDestroyed(ApplicationScoped.class) Object.

Definitely not. Those events were never meant to be used to indicate the startup of application but are instead meant as an initialization of a scope. And they exist for all other scopes as well.

> Besides the implementation for the new events are different. For normal applications with multiple web modules, for application scoped initialization/destroy events, it would be fired once. However, for the startup/shutdown, multiple events will be fired as the event is per web module.

This shouldn't be the case. Both events are delivered only once to their respective observers.

Emily-Jiang:
> > If we want to push for the new events, it might make more sense to deprecate the @initialized|BeforeDestroyed(ApplicationScoped.class) Object.
> 
> Definitely not. Those events were never meant to be used to indicate the startup of application but are instead meant as an initialization of a scope. And they exist for all other scopes as well.

Well, I commented on the earlier comments regarding more understandable events.
> 
> > Besides the implementation for the new events are different. For normal applications with multiple web modules, for application scoped initialization/destroy events, it would be fired once. However, for the startup/shutdown, multiple events will be fired as the event is per web module.
> 
> This shouldn't be the case. Both events are delivered only once to their respective observers.

What I meant is that the startup event is fired for each web modules. If a shared lib (accessible to multiple web modules) observes this startup event, it will receive multiple events per web module.

manovotn:
> What I meant is that the startup event is fired for each web modules. If a shared lib (accessible to multiple web modules) observes this startup event, it will receive multiple events per web module.

Sorry but I still don't understand. Such library should only get one notification. Do you have a reproducer which would show otherwise?

mkouba:
> What I meant is that the startup event is fired for each web modules. If a shared lib (accessible to multiple web modules) observes this startup event, it will receive multiple events per web module.

I don't think it does, AFAIR a shared lib which is not part of a web module should not receive the event with `ServletContext` payload but `java.lang.Object` and should be delivered once.

impl/src/main/java/org/jboss/weld/bootstrap/WeldRuntime.java
```diff
@@ -55,6 +57,8 @@ public BeanManagerImpl getManager(BeanDeploymentArchive beanDeploymentArchive) {
 
     public void shutdown() {
         try {
+            // fire Shutdown event for all modules first
+            fireEventForAllModules(Shutdown.class, new Shutdown(), Any.Literal.INSTANCE);
```
tevans78, line 61: Shouldn't there be a guard here?
```suggestion
if (environment.automaticallyHandleStartupShutdownEvents()){
            fireEventForAllModules(Shutdown.class, new Shutdown(), Any.Literal.INSTANCE);
}
```

manovotn: Yeah, you are right. That's what I get for context switching :)
https://github.com/weld/core/pull/2713 fixes it

Emily-Jiang:
> > What I meant is that the startup event is fired for each web modules. If a shared lib (accessible to multiple web modules) observes this startup event, it will receive multiple events per web module.
> 
> I don't think it does, AFAIR a shared lib which is not part of a web module should not receive the event with `ServletContext` payload but `java.lang.Object` and should be delivered once.

Thank you @mkouba for explaining this!

Emily-Jiang:
> > What I meant is that the startup event is fired for each web modules. If a shared lib (accessible to multiple web modules) observes this startup event, it will receive multiple events per web module.
> 
> Sorry but I still don't understand. Such library should only get one notification. Do you have a reproducer which would show otherwise?

I think I have misunderstood this. Based on what you and Martin explained, it sounds okay. We will try it out to see whether we have any issues.

PR #2722, opened 2022-04-05
WELD-2712 Write more detailed tests for self-interception and allow p…

manovotn:
…rotected access modifier on method and its return type/params

### DIFF
--- a/tests-arquillian/src/test/java/org/jboss/weld/tests/resolution/circular/self/modifiers/SelfInterceptor.java
+++ b/tests-arquillian/src/test/java/org/jboss/weld/tests/resolution/circular/self/modifiers/SelfInterceptor.java
@@ -1,0 +1,40 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2022, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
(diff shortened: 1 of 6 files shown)


tests-arquillian/src/test/java/org/jboss/weld/tests/resolution/circular/self/modifiers/SelfInterceptor.java
```diff
@@ -0,0 +1,40 @@
...
+import jakarta.interceptor.AroundInvoke;
+import jakarta.interceptor.Interceptor;
+import jakarta.interceptor.InvocationContext;
+
+@Interceptor
+@SelfIntercept
+@Priority(1)
+public class SelfInterceptor {
+
+    public static String INTERCEPTED_METHOD_NAME = null;
```
mkouba, line 29: It's not a constant so why the uppercase letters? Just nitpicking ;-)

mkouba (commented):
Looks good!

mkouba (approved):

tests-arquillian/src/test/java/org/jboss/weld/tests/resolution/circular/self/modifiers/SelfInterceptor.java
manovotn: TBF, I don't remember :-D

PR #2725, opened 2022-04-21
Propagate causes in Lite Extension Translator

Azquelt:
Ensure that when we're wrapping an exception using the translation
logger, we retain the proper cause.

Also, if the cause is an InvocationTargetException, it won't have a
useful toString, so use the toString of its cause in the exception
message instead.

Fixes [WELD-2710](https://issues.redhat.com/browse/WELD-2710)


manovotn (approved):
Looks good, thanks @Azquelt!

PR #2727, opened 2022-04-25
WELD-2696 Update documentation.

manovotn:


docs/reference/src/main/asciidoc/ri-spi.asciidoc
```diff
@@ -903,237 +900,103 @@ ifdef::generate-index-link[link:configure.html#external_config[Defining external
 
 === Migration notes
 
-This part of the appendix documents the changes in Weld across major and
-minor releases that an integrator should be aware of. These changes
-mostly touch changes in the SPI or in the container contract.
-
-==== Migration from Weld 1.x to 2.0
```
manovotn, line 910: I've deleted all migration notes and added a new section for Weld 4 -> Weld 5 because I thought the section is getting messy and you can always browse older documentation for information on older migrations.
If you disagree, we can keep this (or keep last two versions perhaps?).

manovotn:
I tried mentioning the new parts and then adjusted other bits I semi-randomly came across.

Few other things to consider:
* Tomcat servlet is untested and documentation is outdated. I noticed they have some Alpha for EE 10 now, see https://tomcat.apache.org/tomcat-10.1-doc/index.html
* We have quite a few Jetty mentions in the documentation but as far as I know, we aren't really testing the integration (and weren't even for Weld 4!). Maybe we should drop all mentions of Jetty from documentation?

docs/reference/src/main/asciidoc/ri-spi.asciidoc
mkouba: +1

manovotn:
> Hm, so we don't run the tests from https://github.com/weld/core/tree/master/environments/servlet/tests/tomcat and https://github.com/weld/core/tree/master/environments/servlet/tests/jetty in CI?

Exactly, at the time of writing, there was no servlet version for EE 10 in case of Tomcat and in case of Jetty we also lacked the Arq. adapter to start up embedded Jetty.

docs/reference/src/main/asciidoc/extend.asciidoc
```diff
@@ -717,3 +717,22 @@ context.
 
 For more information on implementing a custom context see this
 http://in.relation.to/Bloggers/CreatingACustomScope[blog post] or the https://github.com/weld/command-context-example[Command context example].
+
+[[extend_lite]]
+== Build Compatible extensions
```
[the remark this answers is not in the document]
manovotn: Ok, the link is present so I'll this as-is.

manovotn:
I have created https://issues.redhat.com/browse/WELD-2716 to track the servlet-related changes for next version so that it doesn't block the Final release.
I'll take a look at it when I have some free cycles

PR #2728, opened 2022-04-26
WELD-2717 Only fire Startup event right after Initialized(AppScoped) …

manovotn:
…event.

I have seen a failure on WFLY where, for a web module, the `@Initialized(AppScoped) event can happen *after* `Startup` which is against specification. Therefore, we should move the logic to always follow up on `@Initialized` firing.

CC @tevans78 since you reviewed the original PR.
I am fairly surprised the current release worked for you on Open Liberty, I'd expect a similar issue.
Could you verify this doesn't break anything for you, please?

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/WeldRuntime.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/WeldRuntime.java
@@ -58,10 +58,10 @@ public class WeldRuntime {
 
     public void shutdown() {
         try {
-            // fire Shutdown event for all modules first
+            // fire Shutdown event for all non-web modules first
             Environment env = Container.getEnvironment();
             if (env != null && env.automaticallyHandleStartupShutdownEvents()) {
-                fireEventForAllModules(Shutdown.class, new Shutdown(), Any.Literal.INSTANCE);
+                fireEventForNonWebModules(Shutdown.class, new Shutdown(), Any.Literal.INSTANCE);
             }
             // The container must destroy all contexts.
             // For non-web modules, fire @BeforeDestroyed event
--- a/impl/src/main/java/org/jboss/weld/bootstrap/WeldStartup.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/WeldStartup.java
@@ -576,10 +576,11 @@ public class WeldStartup {
             for (BeanDeploymentModule module : modules) {
                 if (!module.isWebModule()) {
                     module.fireEvent(Object.class, ContextEvent.APPLICATION_INITIALIZED, Initialized.Literal.APPLICATION);
-                }
-                // Fire Startup event for all modules if required
-                if (environment.automaticallyHandleStartupShutdownEvents()) {
-                    module.fireEvent(Startup.class, new Startup(), Any.Literal.INSTANCE);
+                    // Fire Startup event for all non-web modules if required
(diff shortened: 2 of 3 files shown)


manovotn:
> Hm, so we effectively fire one Startup event for all non-web modules and one for every web module.

We fire one for every module, just do so from different places so that for each given module, you get `@Initialized` in it fired sooner than you get `Startup`.

> I.e. a bean in an EJB module can be notified before the @Initialized(ApplicationScoped.class) in a web module.

I mean yes, but that's the same as it was up until now with just `@Initialized(ApplicationScoped.class)` - you could have a non-web module notified some time before a web-module. That's why we fire it per-module and not globally.

> In other words, it's still a spec violation, although not tested by the TCK. I do understand why we do it this way but I can't say I like it...

If you have a better option, I am all ears :]

manovotn:
> It is likely that we are not seeing a problem because we're firing the Startup and Shutdown events ourselves. We are not letting Weld do it. 

Ok, that explains it, thanks.

> In addition, some other people have raised questions about the timing of the CDI Startup event vs the execution of a Startup EJB. Something else I need to investigate.

CDI startup and EJB startup are completely unrelated and independent of each other.
Especially given that the main usage of CDI `@Startup` lies in CDI Lite which has no EJB support at all :)

mkouba:
> > Hm, so we effectively fire one Startup event for all non-web modules and one for every web module.
> 
> We fire one for every module, just do so from different places so that for each given module, you get `@Initialized` in it fired sooner than you get `Startup`.
> 
> > I.e. a bean in an EJB module can be notified before the @initialized(ApplicationScoped.class) in a web module.
> 
> I mean yes, but that's the same as it was up until now with just `@Initialized(ApplicationScoped.class)` - you could have a non-web module notified some time before a web-module. That's why we fire it per-module and not globally.

I completely understand (and I still remember a little how we implemented it back then ;-). But the main reason for this approach for `@Initialized(ApplicationScoped.class)` was the payload (`ServletContext` for web modules). Here the main reason is timing...

> 
> > In other words, it's still a spec violation, although not tested by the TCK. I do understand why we do it this way but I can't say I like it...
> 
> If you have a better option, I am all ears :]

I don't have a better option. I'm just grumbling ;-)

tevans78:
> CDI startup and EJB startup are completely unrelated and independent of each other. Especially given that the main usage of CDI `@Startup` lies in CDI Lite which has no EJB support at all :)

While this is a little off topic for this PR, I remind you that CDI Lite is part of CDI Full, which does support EJBs. The spec for the CDI Startup event says it must be fired "before processing requests". In this context, does a "request" include one which was originally driven as part of EJB Startup processing?

manovotn:
> While this is a little off topic for this PR, I remind you that CDI Lite is part of CDI Full, which does support EJBs. The spec for the CDI Startup event says it must be fired "before processing requests". In this context, does a "request" include one which was originally driven as part of EJB Startup processing?

I am aware of that but not every feature within CDI has to be designed with EJB in mind whether they happen to be present or not (even in full EE, you can have applications with CDI and without EJB). The two specs are, after all, independent specifications that happen to have certain level of integration.

BTW the vague formula "processing requests" is the very same one that has been present in CDI for a long time. For instance, the [`AfterDeploymentValidation` chapter](https://jakarta.ee/specifications/cdi/2.0/cdi-spec-2.0.html#after_deployment_validation) states:

> The container must fire an event after it has validated that there are no deployment problems and before creating contexts or processing requests.

Which is equally vague and ambiguous.

PR #2750, opened 2022-06-13
Example of getting embedded tomcat test passing

starksm64:


manovotn:
This is now incorporated in the original PR, thanks!

PR #2763, opened 2022-06-29
Update to Jandex 2.4.3 and avoid deprecated methods

Ladicek:


Ladicek:
Hmm, 2 tests from `AroundTimeoutInterceptorTest` failing with a NPE, I wonder if that could possibly be intermittent... (?)

Ladicek:
Locally, using `mvn clean verify -f jboss-tck-runner/pom.xml`, all tests pass. Which of course uses the embedded Weld container instead of WildFly.

PR #2768, opened 2022-07-07
add webprofileincontainer for running Web Profile in specified JBOSS_HOME

scottmarlow:
Follow up to my email question about ^

Signed-off-by: Scott Marlow <smarlow@redhat.com>

### DIFF
--- a/jboss-tck-runner/pom.xml
+++ b/jboss-tck-runner/pom.xml
@@ -357,7 +357,7 @@
             </build>
         </profile>
 
-        <!-- Run the TCK in the managed container (full Java EE platform) -->
+        <!-- Run the TCK in the managed container (Jakarta EE Full Platform) -->
         <profile>
             <id>incontainer</id>
             <activation>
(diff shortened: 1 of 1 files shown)


scottmarlow:
We may also want a way to specify command line options to pass when starting WildFly, like `./standalone.sh -c standalone-full.xml`

scottmarlow:
> We may also want a way to specify command line options to pass when starting WildFly, like `./standalone.sh -c standalone-full.xml`

I'll check if JAVA_OPTS are passed through to the started WildFly

scottmarlow:
> > We may also want a way to specify command line options to pass when starting WildFly, like `./standalone.sh -c standalone-full.xml`
> 
> I'll check if JAVA_OPTS are passed through to the started WildFly

I tried `export JAVA_OPTS=-c standalone-full.xml` and could tell from the resulting (wildfly) server.log that JAVA_OPTS were not passed into the launched (by Weld TCK Runner).  So it would be nice to have another way to support that.  This doesn't block this pr though, so we can discuss that later on mailing list or future pull request.

manovotn (changes requested):
@scottmarlow what's the need for this profile (didn't read the email, on PTO now, will check on Mon).
The tests it runs are a subset of what's already in `incontainer` profile so passing that is sufficient, especially for testing located in Weld project (all of these tests are done for each PR).

Also, CI tests are currently executed (for WFLY) by running `mvn clean verify -Dincontainer` which will trigger both profiles and their configuration will clash. See https://github.com/weld/core/blob/master/.github/workflows/ci-actions.yml#L182
If you inspect result of the CI job on this PR, you will see that it executed only one run for TCKs and it contains less tests than it should.

manovotn:
@scottmarlow so, in reply to your email...

In the current setup, `JBOSS_HOME` is only used when you run tests with `-Dincontainer`.
The `webprofile` seems to be a subset of the full EE testing, basically the exact same thing minus single test group 'javaee-full`, right?That's what the current profile seems to do.

Setting `JBOSS_HOME` and executing the following command from `jboss-tck-runner` directory:
* `mvn clean verify -Dincontainer`
  * runs all tests, total of 1,831 tests
* `mvn clean verify -Dincontainer -Dwebprofile=true` 
  * runs the webprofile tests, total of 1,701 tests

Therefore, I think the setup works as-is now. Am I missing something?

manovotn:
As for the server configuration file, this is controlled by the Arq. setup and possibly could be altered but IIRC, the standalone-full was insufficient.
ATM we seem to use `standalone-activemq-embedded.xml` because of the JMS broker needed for some of the TCK tests.
See https://github.com/weld/core/blob/master/jboss-tck-runner/src/test/wildfly8/arquillian.xml#L13-L14

scottmarlow:
> @scottmarlow so, in reply to your email...
> 
> In the current setup, `JBOSS_HOME` is only used when you run tests with `-Dincontainer`. The `webprofile` seems to be a subset of the full EE testing, basically the exact same thing minus single test group 'javaee-full`, right?That's what the current profile seems to do.
> 
> Setting `JBOSS_HOME` and executing the following command from `jboss-tck-runner` directory:
> 
>     * `mvn clean verify -Dincontainer`
>       
>       * runs all tests, total of 1,831 tests
Tests run: 1831, Failures: 0, Errors: 0, Skipped: 0

Agreed.

> 
>     * `mvn clean verify -Dincontainer -Dwebprofile=true` 
>       
>       * runs the webprofile tests, total of 1,701 tests

Thanks, I see the same now as well.  So, no need for this pr, closing.

> 
> 
> Therefore, I think the setup works as-is now. Am I missing something?

Agreed.

PR #2779, opened 2022-08-04
WELD-2727 Upgrade wildfly.arquillian.version, fix examples affected by latest WFLY changes

scottmarlow:
Signed-off-by: Scott Marlow <smarlow@redhat.com>

WildFly 27 no longer needs `WildFly Preview` since the EE 10 changes have been merged into the root WildFly `main` branch.  

WildFly 27 updates to a newer WildFly Arquilliant release `5.0.0.Alpha3` so, this pr changes to wildfly.arquillian.version `5.0.0.Alpha3`.

Since WildFly 27 removed the standalone-activemq-embedded.xml, we switched to running the EE 10 Full Platform CDI TCK tests with `standalone-full.xml` and that passes.  Currently, we are doing this by passing properties to maven:

```
# TODO: remove the following copy to standalone-activemq-embedded.xml which works around `standalone-activemq-embedded.xml` being removed from WildFly 27
cp $JBOSS_DIR/standalone/configuration/standalone-full.xml $JBOSS_DIR/docs/examples/configs/standalone-activemq-embedded.xml
export TCK_VERSION=4.0.5
mvn clean verify -D$useProperty -Dcdi.tck-4-0.version=${TCK_VERSION} -Dmaven.test.failure.ignore=true -Dwildfly.arquillian.version=5.0.0.Alpha3 
```

Thanks to @bstansberry for figuring out/suggesting that we need to use the above workaround to pass the CDI TCK with WildFly 27 and providing the [comparison of WildFly Preview 27.0.0.Alpha3 standalone-activemq-embedded.xml vs main's standalone-full.xml](https://gist.github.com/bstansberry/0fb0190da8cdcbd6b31590993b03c627) which I am referencing here just to show the differences of this change to use WildFly 27 standalone-full.xml.


jboss-tck-runner/pom.xml
```diff
@@ -490,8 +490,8 @@
                                     <outputDirectory>${JBOSS_HOME}/standalone/configuration</outputDirectory>
                                     <resources>
                                         <resource>
-                                            <directory>${JBOSS_HOME}/docs/examples/configs</directory>
-                                            <includes>standalone-activemq-embedded.xml</includes>
+                                            <directory>${JBOSS_HOME}/standalone/configuration</directory>
```
jamezp, line 493: We should just delete this whole plugin execution. There is no need to copy the same file to itself.

tests-arquillian/pom.xml
```diff
@@ -314,8 +314,8 @@
                                     <outputDirectory>${JBOSS_HOME}/standalone/configuration</outputDirectory>
                                     <resources>
                                         <resource>
-                                            <directory>${JBOSS_HOME}/docs/examples/configs</directory>
-                                            <includes>standalone-activemq-embedded.xml</includes>
+                                            <directory>${JBOSS_HOME}/standalone/configuration</directory>
```
jamezp, line 317: Same here, just delete the full execution.

jamezp (changes requested):

jboss-tck-runner/pom.xml
scottmarlow: That makes sense, I thought there was some magic somewhere that was causing this to copy to `standalone.xml` but removing the copy makes more sense.

scottmarlow: Done

tests-arquillian/pom.xml
scottmarlow: Done

jamezp (approved):

manovotn (dismissed):
LGTM, let's wait for CI results

jamezp:
> > Since WildFly 27 removed the standalone-activemq-embedded.xml, we switched to running the EE 10 Full Platform CDI TCK tests with standalone-full.xml and that passes. Currently, we are doing this by passing properties to maven:
> 
> So the `full` variant now contains the missing bits again? I recall it was deliberately extracted from there hence why we needed to use this specific config file.

Correct. `standalone-full.xml` now contains the AMQ embedded server configuration. One thing we probably should note is this ONLY works on WildFly upstream and not a released version to my knowledge.

scottmarlow:
> One thing we probably should note is this ONLY works on WildFly upstream and not a released version to my knowledge.

Once we have a WildFly 27 release that can be referenced, this pr should work.  

We probably should also update https://github.com/weld/core/blob/master/.github/workflows/ci-actions.yml#L26 to reference `WildFly` instead of `WildFly Preview`.

scottmarlow:
test results were better than expected for CDI TCK testing, I think we only failed:
```
 Error:  Tests run: 1831, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1,513.716 s <<< FAILURE! - in TestSuite
Error:  org.jboss.cdi.tck.interceptors.tests.contract.aroundTimeout.AroundTimeoutInterceptorTest.testSecurityContext[null](1)  Time elapsed: 5.052 s  <<< FAILURE!
java.lang.NullPointerException
	at deployment.AroundTimeoutInterceptorTest88617be40e944f3b93f2b1d9aec41c8ac137a1.war//org.jboss.cdi.tck.interceptors.tests.contract.aroundTimeout.AroundTimeoutInterceptorTest.testSecurityContext(AroundTimeoutInterceptorTest.java:79)
```

bstansberry:
@manovotn @scottmarlow @jamezp WildFly Preview removed the embedded broker, but std WF did not. It got restored in WFP pretty recently as really just a workaround to making our work moving std WF to EE 10 easier. It's possible we'll remove it again in WFP.

https://issues.redhat.com/browse/WFLY-16714

WildFly Preview has a general purpose beyond just the javax->jakarta transition that's been its main purpose up to now. Looking at things like dropping the embedded broker is one of those. We decided not to do that in stand WF yet, but it's still the right thing to do in the long run, IMHO, so I filed WFLY-16714 to track thinking about that.

One of the things doing things like that in WFP does is force us to deal with the impacts that would have, which can be annoying, but also helps us identify and clear up problems; e.g. how to efficiently test with the broker and without, how to make it easy to provision with a broker or without (see https://github.com/wildfly/wildfly/pull/15413) etc.

manovotn:
> We probably should also update https://github.com/weld/core/blob/master/.github/workflows/ci-actions.yml#L26 to reference WildFly instead of WildFly Preview.

@scottmarlow Yes, I was using preview since some EE 10 parts were still missing in classic WF so please feel free to change that to nightly classic WF.

> test results were better than expected for CDI TCK testing, I think we only failed:

That's a flaky test (which I have no idea how to fix and which never ever failed for me locally :D), there are about 3 such tests that tend to fail in CI and nowhere else.

> @manovotn @scottmarlow @jamezp WildFly Preview removed the embedded broker, but std WF did not. It got restored in WFP pretty recently as really just a workaround to making our work moving std WF to EE 10 easier. It's possible we'll remove it again in WFP.

Got it, thanks for info.

manovotn:
There are also failures in example testing; might be due to different Arq. version

scottmarlow:
The `Weld Examples build and test` failures were:
```
Error:    CommonLoginTest.loginTest:92 » NoSuchElement Unable to locate a node using //l...
Error:    CommonLoginTest.logoutTest:107 » NoSuchElement Unable to locate a node using /...
```

manovotn:
BTW I will be gone for about a week now, so either @mkouba needs to take this over, or I can take a look once I am back :)

manovotn:
I am back and will look back into this today or tomorrow.
One more thing I thought about though is that we had special exclusions plus direct dep in `jakarta` arq. enrichers which will no longer be needed. These should be changed as part of this PR. 
For instance https://github.com/weld/core/blob/master/tests-arquillian/pom.xml#L272-L276
and https://github.com/weld/core/blob/master/tests-arquillian/pom.xml#L138-L143
There is more of these but a simple `grep` should expose them.

@scottmarlow will you add that or should I?

manovotn:
I've added the aforementioned change.
The failure in `examples` seems to be some underlying error with hibernate query and ATM I don't know what's the cause (looks like `import.sql` might not be doing its initial build of the DB?). I am investigating further.

manovotn:
I am not sure what the issue is with `examples` but it looks like some config mismatch between the app and WFLY with its in-memory DB (H2) setup.
On JSF login app startup, WFLY console shows:
```
10:46:44,712 ERROR [org.hibernate.tool.schema.internal.SchemaDropperImpl$DelayedDropActionImpl] (ServerService Thread Pool -- 90) HHH000478: Unsuccessful: drop table if exists User cascade
... // some more bootstrap related lines
10:46:44,968 WARN  [org.hibernate.tool.schema.internal.ExceptionHandlerLoggedImpl] (ServerService Thread Pool -- 90) GenerationTarget encountered exception accepting command : Error executing DDL "drop table if exists User cascade " via JDBC Statement: org.hibernate.tool.schema.spi.CommandAcceptanceException: Error executing DDL "drop table if exists User cascade " via JDBC Statement
	at org.hibernate@6.0.2.Final//org.hibernate.tool.schema.internal.exec.GenerationTargetToDatabase.accept(GenerationTargetToDatabase.java:67)
```

All the app has is a trivial [`import.sql`](https://github.com/weld/core/blob/master/examples/jsf/login/src/main/resources/import.sql) script and [`persistence.xml`](https://github.com/weld/core/blob/master/examples/jsf/login/src/main/resources/META-INF/persistence.xml), none of which seems to have wrong setup (and nothing has changed there).
@scottmarlow does this ring a bell?

manovotn:
The examples should be passing now. The failures were related to changes in WFLY (not to this PR), namely hibernate update and the in memory DB. 
I have removed the `pastecode` example altogether as it was very outdated and would need a complete rework. Besides, it was showcasing complex setup with EJB which isn't something I would recommend and certainly not as a starter example for CDI.

Let's see what CI has to say.

PR #2793, opened 2022-09-29
WELD-2726 Add means to obtain all contextual instances via BM

manovotn:
This PR starts as a draft as it requires new Weld API that will contains changes done in https://github.com/weld/api/pull/172


manovotn:
PR now contains a commit that upgrades Weld API version

PR #2826, opened 2023-01-05
WELD-2738 Bean type assignability for recursive generic types

manovotn:
A draft of how to handle recursive generic bean types.

The approach lies in special casing this scenario once we get to covariant type assignability checks.
We need to detect the recursion and then we assume assignability.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/resolution/AbstractAssignabilityRules.java
+++ b/impl/src/main/java/org/jboss/weld/resolution/AbstractAssignabilityRules.java
@@ -54,7 +54,7 @@ public abstract class AbstractAssignabilityRules implements AssignabilityRules {
      * Standard Java covariant assignability rules are applied to all other types of bounds.
      * This is not explicitly mentioned in the specification but is implied.
      */
-    protected Type[] getUppermostTypeVariableBounds(TypeVariable<?> bound) {
+    static Type[] getUppermostTypeVariableBounds(TypeVariable<?> bound) {
         if (bound.getBounds()[0] instanceof TypeVariable<?>) {
             return getUppermostTypeVariableBounds((TypeVariable<?>) bound.getBounds()[0]);
         }
--- a/impl/src/main/java/org/jboss/weld/resolution/CovariantTypes.java
+++ b/impl/src/main/java/org/jboss/weld/resolution/CovariantTypes.java
@@ -226,12 +226,38 @@ public class CovariantTypes {
             throw ReflectionLogger.LOG.invalidTypeArgumentCombination(type1, type2);
         }
         for (int i = 0; i < type1.getActualTypeArguments().length; i++) {
+            // if it's recursive generics, we treat them as assignable
+            if (isResursiveGenericType(type1, types1[i])) {
+                return true;
+            }
             // Generics are invariant
             if (!InvariantTypes.isAssignableFrom(types1[i], types2[i])) {
                 return false;
             }
         }
         return true;
+    }
+
+    // used to try and detect recursive generic types such as <T extends Comparable<T> List<T>
+    private static boolean isResursiveGenericType(ParameterizedType originalType, Type typeToSearch) {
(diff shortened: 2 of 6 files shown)


manovotn:
So, this passes all the tests in Weld and CDI TCK.

I also looked into Quarkus/Arc and the runtime resolution there is basically a copy of what's here so if we find a solution here, it's mostly copy paste. FTR here is the Quarkus class - https://github.com/quarkusio/quarkus/blob/main/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/CovariantTypes.java

I'd be curious to hear from @mkouba and/or @Ladicek in terms of what do you think about it and whether you can think of a scenario where this wouldn't work.

mkouba (commented):
Looks good but I think that we really need input from our generics (and JLS/JVMS) guru @Ladicek here...

Ladicek:
Sorry it took me so long. I tried adding a few more tests (notably a test with mutually recursive type parameters) and found an issue with the algorithm here (which is pretty easy to fix). This code:

```java
                        // recursive search through all found type args
                        for (Type typeArg : ((ParameterizedType) singleType).getActualTypeArguments()) {
                            if (isResursiveGenericType(originalType, typeArg)) {
                                return true;
                            }
                        }
```

does a recursive call to `isResursiveGenericType` (has typo!) with the _original_ `originalType`, which is wrong; it should be `singleType`.

I took the liberty to rename a few things in this too and pushed a commit in my branch: https://github.com/Ladicek/weld-core/commits/recursive-generics

manovotn:
Thanks @Ladicek, your commit is now part of this PR.
I went through the changes and I think it makes sense now. We can do similar changes for Quarkus/Arc/

impl/src/main/java/org/jboss/weld/resolution/AbstractAssignabilityRules.java
```diff
@@ -54,7 +54,7 @@ public boolean matches(Type requiredType, Set<? extends Type> beanTypes) {
      * Standard Java covariant assignability rules are applied to all other types of bounds.
      * This is not explicitly mentioned in the specification but is implied.
      */
-    protected Type[] getUppermostTypeVariableBounds(TypeVariable<?> bound) {
+    static Type[] getUppermostTypeVariableBounds(TypeVariable<?> bound) {
```
ljnelson, line 57: (That's a backwards-incompatible change, right? Is that OK?)

manovotn: For an internal Weld class, I'd say yes, this is fine.

ljnelson: OK. Are there any conclusions to draw from this? i.e. anything in `impl` can change in any way without prior notice, even if it's `public`? (Does `impl` mean "internal Weld class"? Is that defined/specified/documented anywhere?)

Ladicek: The class is public though? But if it's not supposed to be used or extended, not even by integrators, then I agree it's fine. Restoring compatibility should be pretty straightforward if required.

mkouba: IIRC only the `weld-api` and `weld-spi` artifacts are considered a public API in Weld. The first for users and the latter [for intergrators](https://docs.jboss.org/weld/reference/latest/en-US/html_single/#ri-spi). I don't think it's explicitly defined somewhere but the very existence of these artifacts indicates what the intention was.

manovotn: I agree with Martin, there is a reason why we have API and SPI artifacts as well.
You can technically depend on/use any internal part of Weld but those are subject to change, albeit very rarely these days.

PR #2832, opened 2023-04-14
WELD-2741 Detecting a passivating non-normal scope should lead to deployment exception instead of definition

manovotn:
JIRA - https://issues.redhat.com/browse/WELD-2741

Also bumps the TCK version.

The TCK test asserting this is currently excluded but was already changed to expect this behavior, see https://github.com/jakartaee/cdi-tck/issues/431 and [code](https://github.com/jakartaee/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/full/extensions/lifecycle/bbd/broken/passivatingScope/AddingPassivatingScopeTest.java)

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bootstrap/events/AbstractDefinitionContainerEvent.java
+++ b/impl/src/main/java/org/jboss/weld/bootstrap/events/AbstractDefinitionContainerEvent.java
@@ -17,6 +17,7 @@
 package org.jboss.weld.bootstrap.events;
 
 import org.jboss.weld.exceptions.DefinitionException;
+import org.jboss.weld.exceptions.DeploymentException;
 import org.jboss.weld.logging.BootstrapLogger;
 import org.jboss.weld.manager.BeanManagerImpl;
 import org.jboss.weld.util.Preconditions;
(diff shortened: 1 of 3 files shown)


impl/src/main/java/org/jboss/weld/bootstrap/events/AbstractDefinitionContainerEvent.java
```diff
@@ -43,7 +44,12 @@ public void addDefinitionError(Throwable t) {
     public void fire() {
         super.fire();
         if (!getErrors().isEmpty()) {
-            throw new DefinitionException(getErrors());
+            if (getErrors().size() == 1 && getErrors().get(0) instanceof DeploymentException) {
+                // if the throwable was deployment exception, rethrow that
+                throw new DeploymentException(getErrors().get(0));
```
mkouba, line 49: Hm, shouldn't this be more like `throw getErrors().get(0)`? I mean you don't really rethrow but wrap the original exception...

manovotn: Done

mkouba (approved):

PR #2834, opened 2023-04-18
Add relaxed mode testing into CI runs

manovotn:
Introduce POM property to be able to switch TCK suite XML file.

Introduce a separate XML file for TCKs with web profile and without it.
This is because an XML file with exclusion not present in given TCK profile results in an error (there are two such files already in the CDI TCK anyway).

Add relaxed mode testing into CI - this isn't a complete matrix. Inatead, it is a TCK run against WFLY and a full run of our TS w/o container.


manovotn:
Rebased to resolve conflict in `pom.xml`.

PR #2835, opened 2023-04-21
WELD-2737 Correct wording for what classes are beans in accordance with CDI specification

manovotn:
JIRA - https://issues.redhat.com/browse/WELD-2737

The notion of `@ManagedBean` should not be present as that's deprecated and not a bean defining annotation.
Also update wording to be the same as what CDI spec uses.

### DIFF
--- a/docs/reference/src/main/asciidoc/beans.asciidoc
+++ b/docs/reference/src/main/asciidoc/beans.asciidoc
@@ -407,20 +407,22 @@ out-of-the-box.
 ==== Managed beans
 
 A managed bean is a Java class. The basic lifecycle and semantics of a
-managed bean are defined by the Managed Beans specification. You can
-explicitly declare a managed bean by annotating the bean class
-`@ManagedBean`, but in CDI you don't need to. According to the
-specification, the CDI container treats any class that satisfies the
-following conditions as a managed bean:
+managed bean are defined by the Managed Beans specification.
+According to the specification, the CDI container treats any class that satisfies the following conditions as a managed bean:
 
-* It is not a non-static inner class.
-* It is a concrete class, or is annotated `@Decorator`.
-* It is not annotated with an EJB component-defining annotation or
-declared as an EJB bean class in `ejb-jar.xml`.
-* It does not implement `jakarta.enterprise.inject.spi.Extension`.
-* It has an appropriate constructor—either:
-** the class has a constructor with no parameters, or
-** the class declares a constructor annotated `@Inject`.
+* It is either:
+** an abstract or non-abstract class annotated @Decorator, or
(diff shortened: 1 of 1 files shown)


docs/reference/src/main/asciidoc/beans.asciidoc
```diff
@@ -407,20 +407,19 @@ out-of-the-box.
...
-following conditions as a managed bean:
-
-* It is not a non-static inner class.
-* It is a concrete class, or is annotated `@Decorator`.
+managed bean are defined by the Managed Beans specification.
+According to the specification, the CDI container treats any class that satisfies the following conditions as a managed bean:
+
+* It is not an inner class.
+* It is a non-abstract class.
+* It is an abstract or non-abstract class annotated `@Decorator`.
```
ljnelson, line 415: (As written, this means all managed beans must be annotated with `@Decorator`. Obviously that's not what you mean.)

manovotn: Hm, I suppose you could say the same issue exists in https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#what_classes_are_beans
What wording would you suggest?

ljnelson: I would list "positive" conditions first ("it is…"), then "negatives" ("it is not…"), and be very careful with my language. Here's a draft using mostly what is already written, but tightened up a little bit:

According to the specification, the CDI container treats any class that satisfies all of the following conditions as a managed bean:

* It is either:
** an abstract or non-abstract class annotated `@Decorator`, or
** a non-abstract class.
* It declares either:
** a constructor with no parameters, or
** a constructor annotated `@Inject`
* It is not an inner class.
* It does not implement `jakarta.enterprise.inject.spi.Extension`
* It does not implement `jakarta.enterprise.inject.build.compatible.spi.BuildCompatibleExtension`.
* It is not annotated `@Vetoed`
* It is not in a package annotated `@Vetoed`
* It is not annotated with an EJB component-defining annotation
* It is not declared as an EJB bean class in `ejb-jar.xml`.

mkouba: I would just use the wording from CDI 2.0; i.e. negative conditions only plus _"- It is a non-abstract class, or is annotated `@Decorator`."_ (you don't need to distinguish CDI Full and Lite in weld docs).

The wording in CDI 4.0 is IMO ok.

manovotn: I like @ljnelson's idea; it is easy to read and understand so I've added it here

PR #2847, opened 2023-06-07
WELD-2744 CFW update and needed changes to keep operating correctly on JDK 17+

manovotn:
CFW in version 1.3.0.Final includes following commit - https://github.com/jbossas/jboss-classfilewriter/commit/4028c251ecfdc9ded83e508b013aa306276057a3

This means as soon as `DefaultClassFactory.INSTANCE` gets initialized, you need `--add-opens` to be able to execute the code in there - else you're getting an exception.
Even though we never use CFW to define classes, this still affects us because of the way we instantiate `ClassFile` in our `ProxyFactory` - we use a deprecated method that now gets re-routed to a default implementation of the class factory.

In order to bypass this, we need to use a variant of the `ClassFile` method where we specify CL as well as `ClassFactory` instance. This PR uses dummy variant to do just that.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
@@ -499,7 +499,10 @@ public class ProxyFactory<T> implements PrivilegedAction<T> {
 
     private ClassFile newClassFile(String name, int accessFlags, String superclass, String... interfaces) {
         try {
-            return new ClassFile(name, accessFlags, superclass, interfaces);
+            // We need to use a (non-deprecated) method that avoids instantiating DefaultClassFactory.INSTANCE
+            // If that happens, we will have module accessibility issues and the need to use --add-opens clausules
+            // NOTE: the CL and ClassFactory are never really used to define the class, see WeldDefaultProxyServices
+            return new ClassFile(name, accessFlags, superclass, ProxyFactory.class.getClassLoader(), DummyClassFactoryImpl.INSTANCE, interfaces);
         } catch (Exception e) {
             throw BeanLogger.LOG.unableToCreateClassFile(name, e.getCause());
         }
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/util/WeldDefaultProxyServices.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/util/WeldDefaultProxyServices.java
@@ -46,7 +46,7 @@ public class WeldDefaultProxyServices implements ProxyServices {
         ClassLoader originalLoader = originalClass.getClassLoader();
         if (originalLoader == null) {
             originalLoader = Thread.currentThread().getContextClassLoader();
-            // is it's still null we cannot solve this issue and we need to throw an exception
+            // if it's still null we cannot solve this issue, and we need to throw an exception
             if (originalLoader == null) {
                 throw BeanLogger.LOG.cannotDetermineClassLoader(className, originalClass);
             }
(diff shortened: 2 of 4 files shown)


impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -499,7 +500,10 @@ private Class<T> createProxyClass(Class<?> originalClass, String proxyClassName)
 
     private ClassFile newClassFile(String name, int accessFlags, String superclass, String... interfaces) {
         try {
-            return new ClassFile(name, accessFlags, superclass, interfaces);
+            // We need to use a (non-deprecated) method that avoids instantiating DefaultClassFactory.INSTANCE
+            // If that happens, we will have module accessibility issues and the need to use --add-opens clausules
+            // NOTE: the CL and ClassFactory are never really used to define the class, see WeldDefaultProxyServices
+            return new ClassFile(name, accessFlags, superclass, ProxyFactory.class.getClassLoader(), WeldDefaultProxyServices.DummyClassFactoryImpl.INSTANCE, interfaces);
```
manovotn, line 506: I am not sure `ProxyFactory.class.getClassLoader()` is the best choice here but we need literally *any* non-null CL impl - it never gets used anyway.
We could use TCCL as well but I know that one can be set to `null` in some cases. I am open to better ideas :)

impl/src/main/java/org/jboss/weld/bean/proxy/util/WeldDefaultProxyServices.java
```diff
@@ -183,4 +184,25 @@ public final Class<?> publicDefineClass(String name, byte[] b, int off, int len,
...
+     *
+     * See {@link WeldDefaultProxyServices#defineClass(Class, String, byte[], int, int)} for details on how we define
+     * classes.
+     */
+    public static class DummyClassFactoryImpl implements ClassFactory {
+
+        private DummyClassFactoryImpl() {}
+
+        // volatile because of possible multiple tread defining classes simultaneously
+        public static volatile DummyClassFactoryImpl INSTANCE = new DummyClassFactoryImpl();
```
manovotn, line 201: I am not 100% sure this needs to be `volatile` but I thought it's better safe than sorry case.
Weld can use multiple threads during bootstrap and proxy definition and there is no need to have multiple references of this dummy factory being passes around.

mkouba: Why not `final`?

manovotn: Ah, good point, that's way simpler. Done!

impl/src/main/java/org/jboss/weld/bean/proxy/util/WeldDefaultProxyServices.java
```diff
@@ -183,4 +184,25 @@ public final Class<?> publicDefineClass(String name, byte[] b, int off, int len,
...
+
+    /**
+     * A dummy implementation which has only one purpose - to avoid instantiating {@code DefaultClassFactory.INSTANCE}.
+     * The sole method in this class is never used as we define classes using different means that further vary
+     * between in-container (such as WildFly) and SE setups.
+     *
+     * See {@link WeldDefaultProxyServices#defineClass(Class, String, byte[], int, int)} for details on how we define
+     * classes.
+     */
+    public static class DummyClassFactoryImpl implements ClassFactory {
```
mkouba, line 196: I wonder if it would make sense to move this class into `ProxyFactory.java` or even the `org.jboss.weld.bean.proxy` package so that it does not have to be `public`?

manovotn: Sure, I was considering the same. Current location is more or less a remnant of other approaches I tried to fix this issue.
I will make it a separate class under `org.jboss.weld.bean.proxy` and remove the `public` modifier.

manovotn: Done

manovotn:
All tests passes, going to merge this.

PR #2848, opened 2023-07-01
WELD 2745 Restore proper getReference behavior from 5.1.0.Final and repair CreationalContextImpl destruction and releasing logic

ljnelson:
This PR:
* removes the `createChildCc` parameter from `BeanManagerImpl#getReference` and
* repairs the destruction and releasing logic in `CreationalContextImpl` while
* still allowing `DependentContextTest` to pass

Reference: https://issues.redhat.com/browse/WELD-2745

### DIFF
--- a/impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
+++ b/impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
@@ -21,10 +21,12 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import jakarta.enterprise.context.spi.Contextual;
 import jakarta.enterprise.context.spi.CreationalContext;
@@ -59,6 +61,9 @@ public class CreationalContextImpl<T> implements CreationalContext<T>, WeldCreat
     private final List<ContextualInstance<?>> parentDependentInstances;
 
     private final CreationalContextImpl<?> parentCreationalContext;
+
+    // Precondition for access when non-null: synchronized (dependentInstances)
+    private transient Set<ContextualInstance<?>> destroyed;
 
     private transient List<ResourceReference<?>> resourceReferences;
 
@@ -126,9 +131,12 @@ public class CreationalContextImpl<T> implements CreationalContext<T>, WeldCreat
     public void release(Contextual<T> contextual, T instance) {
         synchronized (dependentInstances) {
             for (ContextualInstance<?> dependentInstance : dependentInstances) {
-                // do not destroy contextual again, since it's just being destroyed
                 if (contextual == null || !(dependentInstance.getContextual().equals(contextual))) {
                     destroy(dependentInstance);
+                } else {
+                    // do not destroy contextual again, since it's just being destroyed, but make sure its dependencies
+                    // are destroyed
+                    release(dependentInstance);
                 }
             }
         }
@@ -139,8 +147,29 @@ public class CreationalContextImpl<T> implements CreationalContext<T>, WeldCreat
         }
     }
 
-    private static <T> void destroy(ContextualInstance<T> beanInstance) {
-        beanInstance.getContextual().destroy(beanInstance.getInstance(), beanInstance.getCreationalContext());
+    private <T> void destroy(ContextualInstance<T> beanInstance) {
+        // Precondition: synchronized (dependentInstances)
+        if (this.destroyed == null) {
+            this.destroyed = new HashSet<>();
+        }
+        if (this.destroyed.add(beanInstance)) {
+            beanInstance.getContextual().destroy(beanInstance.getInstance(), beanInstance.getCreationalContext());
+        }
+    }
+
+    private <T> void release(ContextualInstance<T> beanInstance) {
+        // Precondition: synchronized (dependentInstances)
+        if (this.destroyed == null) {
+            this.destroyed = new HashSet<>();
+        }
+        if (this.destroyed.add(beanInstance)) {
+            CreationalContext<T> cc = beanInstance.getCreationalContext();
+            if (cc instanceof CreationalContextImpl<?>) {
+                ((CreationalContextImpl<T>) cc).release(beanInstance.getContextual(), beanInstance.getInstance());
+            } else {
+                cc.release();
+            }
+        }
     }
 
     /**
--- a/impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
+++ b/impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
@@ -663,11 +663,7 @@ public class BeanManagerImpl implements WeldManager, Serializable {
     }
 
     public Object getReference(Bean<?> bean, Type requestedType, CreationalContext<?> creationalContext, boolean noProxy) {
-        return getReference(bean, requestedType, creationalContext, noProxy, true);
-    }
-
-    private Object getReference(Bean<?> bean, Type requestedType, CreationalContext<?> creationalContext, boolean noProxy, boolean createChildCc) {
-        if (creationalContext instanceof CreationalContextImpl<?> && createChildCc) {
+        if (creationalContext instanceof CreationalContextImpl<?>) {
             creationalContext = ((CreationalContextImpl<?>) creationalContext).getCreationalContext(bean);
         }
         if (!noProxy && isProxyRequired(bean)) {
@@ -704,7 +700,7 @@ public class BeanManagerImpl implements WeldManager, Serializable {
         // Ensure that there is no injection point associated
         final ThreadLocalStackReference<InjectionPoint> stack = currentInjectionPoint.push(EmptyInjectionPoint.INSTANCE);
         try {
-            return getReference(bean, requestedType, creationalContext, false, false);
+            return getReference(bean, requestedType, creationalContext, false);
         } finally {
             stack.pop();
         }
--- a/tests-arquillian/src/test/java/org/jboss/weld/tests/beanManager/getReference/interceptor/ManualInterceptorInstanceRetrievalTest.java
+++ b/tests-arquillian/src/test/java/org/jboss/weld/tests/beanManager/getReference/interceptor/ManualInterceptorInstanceRetrievalTest.java
@@ -1,0 +1,46 @@
+package org.jboss.weld.tests.beanManager.getReference.interceptor;
+
+import jakarta.enterprise.context.spi.CreationalContext;
+import jakarta.enterprise.inject.spi.BeanManager;
+import jakarta.enterprise.inject.spi.InterceptionType;
+import jakarta.enterprise.inject.spi.Interceptor;
+import jakarta.inject.Inject;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.BeanArchive;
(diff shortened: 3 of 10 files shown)


impl/nb-configuration.xml
```diff
@@ -0,0 +1,18 @@
...
+    <properties xmlns="http://www.netbeans.org/ns/maven-properties-data/1">
+        <!--
+Properties that influence various parts of the IDE, especially code formatting and the like. 
+You can copy and paste the single properties, into the pom.xml file and the IDE will pick them up.
+That way multiple projects can share the same settings (useful for formatting rules for example).
+Any value defined here will override the pom.xml file value but is only applicable to the current project.
+-->
+        <netbeans.hint.jdkPlatform>JDK_11</netbeans.hint.jdkPlatform>
+    </properties>
+</project-shared-configuration>
```
ljnelson, line 18: (My bad; not sure how this got in here.)

tests-arquillian/nb-configuration.xml
```diff
@@ -0,0 +1,18 @@
...
+    <properties xmlns="http://www.netbeans.org/ns/maven-properties-data/1">
+        <!--
+Properties that influence various parts of the IDE, especially code formatting and the like. 
+You can copy and paste the single properties, into the pom.xml file and the IDE will pick them up.
+That way multiple projects can share the same settings (useful for formatting rules for example).
+Any value defined here will override the pom.xml file value but is only applicable to the current project.
+-->
+        <netbeans.hint.jdkPlatform>JDK_11</netbeans.hint.jdkPlatform>
+    </properties>
+</project-shared-configuration>
```
ljnelson, line 18: (My bad; not sure how this got in here.)

ljnelson:
(I see a failure with `Weld1234Test.java`; not present when I run `mvn clean install` on my local `weld/core` installation. Investigating.)

ljnelson:
I followed https://github.com/weld/core#upgrading-weld-in-wildfly and https://github.com/weld/core#running-integration-tests-and-the-tck-on-wildfly and got:
```
java.lang.NoClassDefFoundError: org/jboss/cdi/tck/extlib/Translator
	at deployment.InstalledLibraryWarTesteac726309f6549cbb54715e112e86a7d4d9fa42.war//org.jboss.cdi.tck.tests.deployment.packaging.installedLibrary.InstalledLibraryWarTest.testInjection(InstalledLibraryWarTest.java:68)
Caused by: java.lang.ClassNotFoundException: org.jboss.cdi.tck.extlib.Translator from [Module "deployment.InstalledLibraryWarTesteac726309f6549cbb54715e112e86a7d4d9fa42.war" from Service Module Loader]
	at deployment.InstalledLibraryWarTesteac726309f6549cbb54715e112e86a7d4d9fa42.war//org.jboss.cdi.tck.tests.deployment.packaging.installedLibrary.InstalledLibraryWarTest.testInjection(InstalledLibraryWarTest.java:68)
```
I'll pause and wait for instructions.

manovotn:
> I followed https://github.com/weld/core#upgrading-weld-in-wildfly and https://github.com/weld/core#running-integration-tests-and-the-tck-on-wildfly and got:
> 
> ```
> java.lang.NoClassDefFoundError: org/jboss/cdi/tck/extlib/Translator
> 	at deployment.InstalledLibraryWarTesteac726309f6549cbb54715e112e86a7d4d9fa42.war//org.jboss.cdi.tck.tests.deployment.packaging.installedLibrary.InstalledLibraryWarTest.testInjection(InstalledLibraryWarTest.java:68)
> Caused by: java.lang.ClassNotFoundException: org.jboss.cdi.tck.extlib.Translator from [Module "deployment.InstalledLibraryWarTesteac726309f6549cbb54715e112e86a7d4d9fa42.war" from Service Module Loader]
> 	at deployment.InstalledLibraryWarTesteac726309f6549cbb54715e112e86a7d4d9fa42.war//org.jboss.cdi.tck.tests.deployment.packaging.installedLibrary.InstalledLibraryWarTest.testInjection(InstalledLibraryWarTest.java:68)
> ```
> 
> I'll pause and wait for instructions.

That's CDI TCK ext library - only needed for 3 (or 4?) specific TCK tests so not what's failing here.
The way you can prepare WFLY for running all tests is:
* Checkout Weld and build with `mvn clean install` (`-DskipTests`)
* `export JBOSS_HOME=/path/to/wfly //CI uses nightly build for instance but you could use last released too
* mvn clean package -Pupdate-jboss-as -Dtck -f jboss-as/pom.xml
  * This updates Weld inside given WFLY dist to what your current build is
  * `-Dtck` is what installs the TCK ext lib
* You can now run incontainer tests by `mvn clean verify -Dincontainer` - be it TCKs or internal tests 

And you're right that the `-Dtck` param is missing in the docs!

> (I see a failure with Weld1234Test.java; not present when I run mvn clean install on my local weld/core installation.

The test in question is WFLY-only (marked with `@Category(Integration.class)`) which if why you don't see it fail locally without server.

Note that the problem could also be in some WFLY change since CI uses latest nigthly build (although it happens rarely).
This can be verified by using latest released version of WFLY -  I can do that on Mon.

manovotn:
I'll also take a look at your proposal on Mon @ljnelson as I am not near my workstation now.
I'd also love to add some tests (runnable without container) that tests the scenario we have been discussing - after all, we got here because there was no coverage, so we want to make sure we can avoid repeating the same mistake :)

ljnelson:
With the updated instructions all tests pass on my machine using Wildfly 28.0.1.Final on my machine using `JAVA_HOME=$(/usr/libexec/java_home -v11) JBOSS_HOME=/Users/lairdnelson/Downloads/wildfly-28.0.1.Final mvn clean verify -Dincontainer -f tests-arquillian/pom.xml`.

manovotn:
The test failing here is definitely unrelated. Over the weekend dependabot created some dependency update PRs and they all fail with the same error.

That's likely a WFLY change, I'm looking into that as well.

impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
```diff
@@ -42,6 +44,7 @@
  * @author Pete Muir
  * @author Ales Justin
  * @author Jozef Hartinger
+ * @author <a href="https://about.me/lairdnelson" target="_parent">Laird Nelson</a>
```
manovotn, line 47: Please don't add yourself; we should probably remove these sections as they are outdated in many place anyway.
Plus git and GH are good enough tools to determine authors and contributors. In fact they are far better as they can show you exactly who authored any part of the code.

impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
```diff
@@ -139,8 +150,33 @@ public void release(Contextual<T> contextual, T instance) {
...
+            CreationalContext<T> cc = beanInstance.getCreationalContext();
+            if (cc instanceof CreationalContextImpl<?>) {
+                ((CreationalContextImpl<T>) cc).release(beanInstance.getContextual(), beanInstance.getInstance());
+            } else {
+                cc.release();
+            }
+        }
+    }
+
+    private boolean isDestroyed(Contextual<?> c) {
```
manovotn, line 172: IDE shows me this method is never really used - probably a leftover from your earlier iterations?

impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
```diff
@@ -192,6 +228,14 @@ protected Object writeReplace() throws ObjectStreamException {
         return this;
     }
 
+    // Serialization
+    protected Object readResolve() throws ObjectStreamException {
```
manovotn, line 232: Do we actually need this method?
I am struggling to see when are we going to use it.

impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
```diff
@@ -60,6 +63,9 @@ public class CreationalContextImpl<T> implements CreationalContext<T>, WeldCreat
 
     private final CreationalContextImpl<?> parentCreationalContext;
 
+    // Precondition for access when non-null: synchronized (dependentInstances)
```
manovotn, line 66: :+1: this is a smart way not to have to use another synchronized collection

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
```diff
@@ -193,6 +193,7 @@
  * @author Marius Bogoevici
  * @author Ales Justin
  * @author Jozef Hartinger
+ * @author <a href="https://about.me/lairdnelson" target="_parent">Laird Nelson</a>
```
manovotn, line 196: Same as above

manovotn (commented):
Added a few remarks.
The WFLY issue is being discussed elsewhere and the failing test can be ignored for now - if you want to see the whole CI run, feel free to mark that test with `@Ignore`.

impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
```diff
@@ -89,6 +96,7 @@ private CreationalContextImpl() {
         // We can't use immutable empty lists because of some rare scenarios with Instance.get()
         this.dependentInstances = Collections.synchronizedList(new ArrayList<ContextualInstance<?>>());
         this.parentDependentInstances = Collections.synchronizedList(new ArrayList<ContextualInstance<?>>());
+        this.destroyed = new HashSet<>();
```
manovotn, line 99: @ljnelson I think you might want to be lazy here and avoid the allocation if you can because many contexts won't need it. Start with `null` and, since you are only using this in a synchronized block, only create the `HashSet` when first needed.

impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
ljnelson: Done

impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
ljnelson: Done

impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
ljnelson: Removed

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
ljnelson: Done

impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
ljnelson: Done

manovotn:
I've forced push into your branch @ljnelson:

* Squashed your commits into one
* Rebased onto master which now has fix for the WFLY test failure you were seeing
* Added tests for `getReference` that are based on your example in the JIRA and then one more for the manual interceptor resolution

manovotn (approved):
I am good with this PR, although I'd like to hear Martin's opinion as well.

Thanks for deep diving into this @ljnelson :)
Once merged, I'll probably consider an SP release so this gets out as fast as possible.

mkouba (commented):
I'm +0 for this change.

Pros:
- compatibility with older versions of Weld
- test coverage

Cons:
- stick to the behavior that is not defined by the spec; and does not make a lot of sense IMO
- even more complex code and `CC#destroyed`

manovotn:
There were some bank holiday and then a PTO, hence my silence on this PR.
I am fine with its state and hoping to get back to this later this week in order to merge it and create an SP release containing this fix as well as https://github.com/weld/core/pull/2862 in it.
Thanks for your extensive research and PR @ljnelson!

Martin is right that the state isn't ideal and while I agree I would prefer not to create an extra empty parent CC in most cases, I think we might want to revisit this for some future release that's not a micro with just bug fixes.

PR #2862, opened 2023-07-03
WELD-2747 Avoid calling SessionObjectReference.isRemoved() on every invocation.

pferraro:
https://issues.redhat.com/browse/WELD-1234

SessionObjectReference.isRemoved() can be expensive.  Let's avoid calling this per invocation, and instead return the fabricated toString() if the EJB was not found.
See https://github.com/wildfly/wildfly/pull/16885 for context.

### DIFF
--- a/modules/ejb/src/main/java/org/jboss/weld/module/ejb/EnterpriseBeanProxyMethodHandler.java
+++ b/modules/ejb/src/main/java/org/jboss/weld/module/ejb/EnterpriseBeanProxyMethodHandler.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Set;
 
 import jakarta.ejb.EJBException;
+import jakarta.ejb.NoSuchEJBException;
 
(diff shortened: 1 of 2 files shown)


manovotn:
Hmm, this doesn't seem to cut it, the test is still failing @pferraro.
CI runs on latest WFLY patched to Weld in this PR, so the expectation is that is should succeed.

pferraro:
Ah - this is because the proxy invocation handler used by WildFly handles toString() via an earlier interceptor: https://github.com/wildfly/wildfly/blob/main/ejb3/src/main/java/org/jboss/as/ejb3/component/EJBComponentDescription.java#L972
which means that toString() will return successfully (i.e. it does not throw a NoSuchEJBException), even if a previous method threw an exception causing the bean instance to be discarded.

I'll fix the test assertion accordingly.

manovotn:
I've created a tracking issue for this - https://issues.redhat.com/browse/WELD-2747

Thanks for the PR @pferraro :)

manovotn (approved):

PR #2863, opened 2023-07-18
WELD-2756 Add support for InvocationContext.getInterceptorBindings()

Ladicek:


manovotn:
This PR requires Weld API release with https://github.com/weld/api/pull/177.
The TCK upgrade doesn't seem mandatory as there are no TCK tests for this.

Notes to self:
* We should also look at any internal usage of our own now-deprecated APIs and see if we can just defer to the standard way instead
* We also need to take a look at any doc references and correct them - probably keep relevant sections and mention that this is now standardized.

manovotn:
Rebased, added deprecation notes to docs and tests, updated deps (no longer depends on SNAPSHOT of TCKs and need 6.x for Weld API).

manovotn (approved):

manovotn:
Rebased onto master, the CI should now pass just fine.

manovotn (approved):
Thanks @Ladicek

PR #2864, opened 2023-07-25
WELD-2764 Method invokers, first impl

manovotn:
JIRA - [WELD-2764](https://issues.redhat.com/browse/WELD-2764)

Draft for the prototype of invokeable methods in Weld. 
~~CI failures are expected due to `SNAPSHOT` deps.~~
CI should now be fine , there are no longer any snapshot dependencies and there is added profile that allows CDI and interceptors API patching in WFLY. All tests should pass.

The impl is based on Java `MethodHandles`.

**How to try out:**

* Checkout and build (`mvn clean install`) the [CDI branch](https://github.com/jakartaee/cdi/pull/639) with invokeable methods API proposal
* Checkout and build this Weld branch (`mvn clean install -DskipTests`)
* You now have Weld `5.1.2-SNAPSHOT` with draft of the impl
  * If you wish to see use cases, take a look at the tests in this PR

**Looking at performance:**

The implementation uses `MethodHandles` and as such is very efficient. The difference in direct method invocation versus invocation via `Invoker` becomes negligible the moment method does anything non-trivial (basically anything else than returning a constant). Currently, I have been playing with some microbenchmarks using JMH to test this. A WIP can be seen over in [this repo and branch](https://github.com/weld/weld-core-benchmarks/tree/invokableMethods). If you want to run that:
* Check out the repo and switch to the `invokableMethods` branch
* `mvn clean install`
* `./run-benchmarks.sh -v "5.1.2-SNAPSHOT" -b "InvokableMethodBenchmarkNoLookup InvokableMethodBenchmarkSimpleBeanInvocation InvokableMethodBenchmarkLookup"`

**Known TODOs:**
* ~~Support for build compatible implementations (currently only works via portable extensions)~~
  * BCEs are now supported as well; there is basic test coverage as the impl is more or less a matter of delegation
* ~~Support exception transformers returning arbitrary value (which is something method handles don't allow out of the box)~~
  * You can now return any return value from exception transformers; this requires some juggling on Weld side, but ultimately works
* ~~Properly support lookup with qualifiers~~
  * The PR now supports this and has respective test coverage
* Proper exception handing setup
* Deployment validation of lookups
  * This should also include deployment resolution of lookups and storing them in some form that will speed up repeated resolutions when invoking the method
* Validation of invokable methods during invoker creation
  * After removal of `@Invokable` we now need to add verification that happens when you attempt to create an invoker for any given method; for instance validate that the method is non-private
* Tests/changes for invokeable methods with generics?


manovotn:
Added changes to support recent adjustment in CDI PR - removal of `@Invokable` and the need to hold metadata about which methods are invokable for each bean class. You can now attempt to create invoker for any method of given bean (and some should fail if you do)

This also reduces the volume of this PR from 96 changes files to just 68 :)

manovotn:
I guess all of the in-container tests fail because we aren't patching the CDI JAR in WFLY dist; will take a closer look on Mon as to whether we can improve that until we have WFLY version which consumes the expected artifact.

manovotn:
I've created a follow up [JIRA issue](https://issues.redhat.com/browse/WELD-2765) capturing leftover TODOs from this one and targetting Alpha2.

PR #2871, opened 2023-08-04
WELD-2750 Tweak how parent-child creational contexts are created. Use WeakReference to keep track of destroyed instance.

manovotn:
Attempts to solve a memory leak in dependent instance handling.
See the JIRA ([WELD-2750](https://issues.redhat.com/browse/WELD-2750)) for detailed description and a reproducer.

First commit:
* Makes sure the set of destroyed instance we keep in each CC is a `WeakReference` allowing GC to collect them. 
* Also changes how we create CC for non-dependent beans when obtaining injectable reference - we no longer unconditionally create child CC as that implies linked them together and potentially holding a reference to an already destroyed instance.

Second commit:
* An improvement on the second step of first commit - tries to avoid creating any superfluous CC when not needed. A side effect is that `BeanManagerImpl#getReference` can no longer return `null`. 
  * However, after browsing code I didn't see any such usage anyway - CI might still prove me wrong though as I only ran non-container tests locally :)

### DIFF
--- a/impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
+++ b/impl/src/main/java/org/jboss/weld/contexts/CreationalContextImpl.java
@@ -21,12 +21,12 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


manovotn:
CC @ljnelson as this affects the code you recently contributed.

ljnelson:
Haven't looked in detail but `WeakReference` was absolutely where I wanted to go initially but didn't think it was worth the complexity. Clearly I was wrong!

PR #2875, opened 2023-09-01
WELD-2752 WeldInitialListener will now only attempt context destruction if init was successful

manovotn:
JIRA - https://issues.redhat.com/browse/WELD-2752

Lifecycle can be `null` if deployed app failed to start. In such case we want to skip context destruction to avoid throwing NPE.


manovotn:
Cc @Azquelt

Azquelt (approved):

PR #2886, opened 2023-09-26
WELD-2758 Proxies for non-public class-based beans should reside in bean's package if possible

manovotn:
https://issues.redhat.com/browse/WELD-2758


mkouba (approved):
Looks good.

PR #2893, opened 2023-10-17
WELD-2762 Do not swallow exceptions occuring during HTTP session creation

manovotn:
I am not sure if this can lead to unexpected scenarios due to the fact that the bean store is lazy but at first sight it seems OK to propagate the exception as it is out of our control anyway.

### DIFF
--- a/modules/web/src/main/java/org/jboss/weld/module/web/context/beanstore/http/LazySessionBeanStore.java
+++ b/modules/web/src/main/java/org/jboss/weld/module/web/context/beanstore/http/LazySessionBeanStore.java
@@ -85,10 +85,10 @@ public class LazySessionBeanStore extends AbstractSessionBeanStore {
         try {
             return SessionHolder.getSession(request, create);
         } catch (IllegalStateException e) {
-            // If container can't create an underlying session, invalidate the
-            // current one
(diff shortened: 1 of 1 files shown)


modules/web/src/main/java/org/jboss/weld/module/web/context/beanstore/http/LazySessionBeanStore.java
```diff
@@ -85,9 +86,10 @@ protected HttpSession getSession(boolean create) {
         try {
             return SessionHolder.getSession(request, create);
         } catch (IllegalStateException e) {
-            // If container can't create an underlying session, invalidate the
-            // current one
+            // If container can't create an underlying session, invalidate the current one
             detach();
+            // re-throw the exception to properly show cause and message
+            Exceptions.rethrowException(e);
```
mkouba, line 92: I wonder if it wouldn't be safer (compatibility, etc.) to log an error instead?

manovotn: According to the issue (and the linked JBEAP issue, I figured having actual exception propagated would make more sense for clarity and debugging.

PR #2897, opened 2023-11-14
improvements to the method invokers implementation

Ladicek:
Includes one tiny improvement to the Language Model implementation that should be clarified in the next version of the specification.


Ladicek:
The improvements here should be non-controversial.

With these improvements, Weld passes most of the proposed TCK for method invokers (https://github.com/jakartaee/cdi-tck/pull/502), with the exception of cca 4 tests where the specification needs more discussion.

manovotn (approved):

PR #2898, opened 2023-11-30
WELD-2768 support explicitly declaring @Priority on producers

manovotn:
Weld implementation for https://github.com/jakartaee/cdi/issues/556
Tests for this are on a TCK PR - https://github.com/jakartaee/cdi-tck/pull/499 (and this PR passes them)
I don't mind adding tests here as well, but it would basically be a duplication of the same ones - WDYT @mkouba?

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/AbstractProducerBean.java
+++ b/impl/src/main/java/org/jboss/weld/bean/AbstractProducerBean.java
@@ -24,6 +24,7 @@ import java.lang.reflect.Member;
 import java.lang.reflect.Type;
 import java.util.Set;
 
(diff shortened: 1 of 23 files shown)


mkouba (approved):
Looks good.

manovotn:
I've updated the PR, notably I:
* Moved the processing of explicit priority into a protected method that subclasses need to invoke (because they also need to initialize their annotated references prior to this)
* Added test coverage equal to what's in the TCKs in case TCK changes or excludes something in the future
* Added extra testing for combination of locally enabled alternative (via `beans.xml`) declaring producers with explicit priority values

PR #2900, opened 2023-12-11
Upgrade Weld API to 6.0.Alpha2; make ForwardingBeanManager implement EL-aware BM

manovotn:


manovotn:
Looks like it works now :tada:

PR #2901, opened 2024-01-04
Update Arq. Weld container

manovotn:
https://github.com/arquillian/arquillian-container-weld/releases/tag/4.0.0.Final


manovotn:
We should also take a loot at other Arq. versions but I am not sure how many things will break WRT in-container testing.

manovotn:
Also updated shrinkwrap resolver version which was causing mayhem in the SE test department.

PR #2904, opened 2024-01-18
WELD-2773 Incorrect producer field inheritance

manovotn:
Related to https://github.com/weld/core/pull/2903

This is a minimal change which corrects how we identify declaring class for a non-discovered annotated type (one that was tempered with by an extension for instance).

### DIFF
--- a/tests-arquillian/src/test/java/org/jboss/weld/tests/producer/field/FieldProducerBean.java
+++ b/tests-arquillian/src/test/java/org/jboss/weld/tests/producer/field/FieldProducerBean.java
@@ -1,0 +1,13 @@
+package org.jboss.weld.tests.producer.field;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.context.Dependent;
+import jakarta.enterprise.inject.Produces;
+
+@Dependent
+public class FieldProducerBean {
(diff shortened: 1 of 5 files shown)


tests-arquillian/src/test/java/org/jboss/weld/tests/producer/field/FieldProducerBean.java
```diff
@@ -0,0 +1,11 @@
+package org.jboss.weld.tests.producer.field;
+
+import jakarta.enterprise.context.Dependent;
+import jakarta.enterprise.inject.Produces;
+
+@Dependent
+public class FieldProducerBean {
+
+    @Produces
+    String p = "42";
```
mkouba, line 10: I'd try to avoid using JDK types without qualifiers as bean types in the tests. It might happen that the container registers ambiguous beans...

mkouba (approved):

tests-arquillian/src/test/java/org/jboss/weld/tests/producer/field/FieldProducerBean.java
manovotn: Fair enough, I've changed that to a custom type.

PR #2906, opened 2024-01-23
WELD-2774 Implement BM methods for event/bean assignability

manovotn:
The PR is going to fail CI until we get a CDI API release and ideally a TCK release to have tests.

There are no Weld specific tests for this as Weld doesn't add anything beyond what's in the TCK coverage via https://github.com/jakartaee/cdi-tck/pull/526

### DIFF
--- a/impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
+++ b/impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
@@ -53,6 +53,8 @@ import jakarta.enterprise.context.spi.Context;
 import jakarta.enterprise.context.spi.Contextual;
 import jakarta.enterprise.context.spi.CreationalContext;
 import jakarta.enterprise.event.Event;
+import jakarta.enterprise.inject.Any;
+import jakarta.enterprise.inject.Default;
 import jakarta.enterprise.inject.Instance;
 import jakarta.enterprise.inject.spi.Annotated;
 import jakarta.enterprise.inject.spi.AnnotatedField;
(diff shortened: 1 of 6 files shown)


manovotn:
I've added some validations:
* Qualifier validation for all inputs (if the annotation is a CDI qualifier)
* Event type validation for unresolved type variable
* Filtering of provided bean types for legal bean types only

These are so far not covered by the linked TCK test.
If they won't be covered, I'll add Weld test cases.

impl/src/main/java/org/jboss/weld/manager/BeanManagerImpl.java
```diff
@@ -1563,6 +1568,73 @@ public WeldInstance<Object> createInstance() {
...
+        validateQualifiers(observedEventQualifiers, loggingMethodName);
+        // type check first
+        if (EventTypeAssignabilityRules.instance().matches(observedEventType,
+                HierarchyDiscovery.forNormalizedType(eventType).getTypeClosure())) {
+            // normalize qualifiers, adding built-ins where missing, then compare
+            Set<Annotation> normalizedEventQualifiers = new HashSet<>(eventQualifiers);
+            if (eventQualifiers.isEmpty()
+                    || (eventQualifiers.size() == 1 && eventQualifiers.iterator().next().equals(Any.Literal.INSTANCE))) {
+                normalizedEventQualifiers.add(Default.Literal.INSTANCE);
+            }
```
Ladicek, line 1619: With our recently acquired deep knowledge of the subject, should this be just `if (eventQualifiers.isEmpty())`?

manovotn: Yes, I believe it should.

manovotn: This should now be fixed

manovotn:
Superseeded by https://github.com/weld/core/pull/2911

PR #2907, opened 2024-01-26
WELD-2775 Make sure BM#getEvent() adds @Default qualifier correctly

manovotn:
Attempts to fix the missing `@Default` qualifier when firing events via `BeanManager.getEvent()` API.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/event/EventMetadataImpl.java
+++ b/impl/src/main/java/org/jboss/weld/event/EventMetadataImpl.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Type;
 import java.util.Set;
 
 import jakarta.enterprise.inject.Any;
+import jakarta.enterprise.inject.Default;
 import jakarta.enterprise.inject.spi.EventMetadata;
 import jakarta.enterprise.inject.spi.InjectionPoint;
 
@@ -59,6 +60,9 @@ public final class EventMetadataImpl implements EventMetadata {
     public Set<Annotation> getQualifiers() {
         ImmutableSet.Builder<Annotation> builder = ImmutableSet.<Annotation> builder();
         builder.add(Any.Literal.INSTANCE);
+        if ((qualifiers != null && qualifiers.isEmpty()) || (qualifierArray != null && qualifierArray.length == 0)) {
+            builder.add(Default.Literal.INSTANCE);
+        }
         if (qualifiers != null) {
             return builder.addAll(qualifiers).build();
         } else if (qualifierArray != null) {
--- a/impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
+++ b/impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
@@ -40,6 +40,7 @@ import java.util.function.Supplier;
 
 import jakarta.enterprise.event.NotificationOptions;
 import jakarta.enterprise.event.ObserverException;
+import jakarta.enterprise.inject.Default;
 import jakarta.enterprise.inject.spi.EventMetadata;
 import jakarta.enterprise.inject.spi.ObserverMethod;
 
@@ -120,7 +121,7 @@ public class ObserverNotifier {
      * Resolves observer methods based on the given event type and qualifiers. If strict checks are enabled the given type is
      * verified.
      *
-     * @param event the event object
+     * @param eventType the event object
      * @param qualifiers given event qualifiers
      * @return resolved observer methods
      */
@@ -133,7 +134,7 @@ public class ObserverNotifier {
      * Resolves observer methods based on the given event type and qualifiers. If strict checks are enabled the given type is
      * verified.
      *
-     * @param event the event object
+     * @param eventType the event object
      * @param qualifiers given event qualifiers
      * @return resolved observer methods
      */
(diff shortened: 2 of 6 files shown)


tests-arquillian/src/test/java/org/jboss/weld/tests/event/ObservingBean.java
```diff
@@ -0,0 +1,92 @@
...
+import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.event.ObservesAsync;
+import jakarta.enterprise.inject.Default;
+import jakarta.enterprise.inject.spi.EventMetadata;
+
+@ApplicationScoped
+public class ObservingBean {
+
+    private int defaultObjectNotified = 0;
+    private int defaultObjectAsyncNotified = 0;
```
mkouba, line 16: These fields should be `volatile`, at least for the async events where the observers are notified on a different thread.

mkouba (approved):
Looks good.

Although I don't think the requirement for `BM#getEvent()` is correct. It should be more like `BM#createInstance()`...

tests-arquillian/src/test/java/org/jboss/weld/tests/event/ObservingBean.java
manovotn: Good point, thanks

manovotn:
> Looks good.
> 
> Although I don't think the requirement for `BM#getEvent()` is correct. It should be more like `BM#createInstance()`...

I agree, but for now this is what we're stuck with :shrug:

impl/src/main/java/org/jboss/weld/event/EventMetadataImpl.java
```diff
@@ -59,6 +60,9 @@ private EventMetadataImpl(Type type, InjectionPoint injectionPoint, Set<Annotati
     public Set<Annotation> getQualifiers() {
         ImmutableSet.Builder<Annotation> builder = ImmutableSet.<Annotation> builder();
         builder.add(Any.Literal.INSTANCE);
+        if (qualifiers.isEmpty()) {
```
Ladicek, line 63: The code below suggests that you should also look at `qualifierArray` (and that exactly one of `qualifiers` and `qualifierArray` is non-null).

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -200,9 +201,15 @@ public void fireEvent(Object event, Resolvable resolvable) {
     protected Resolvable buildEventResolvable(Type eventType, Set<Annotation> qualifiers) {
         // We can always cache as this is only ever called by Weld where we avoid non-static inner classes for annotation literals
         Set<Type> typeClosure = sharedObjectCache.getTypeClosureHolder(eventType).get();
-        return new ResolvableBuilder(resolver.getMetaAnnotationStore()).addTypes(typeClosure).addType(Object.class)
+        ResolvableBuilder resolvableBuilder = new ResolvableBuilder(resolver.getMetaAnnotationStore()).addTypes(typeClosure)
+                .addType(Object.class)
                 .addQualifiers(qualifiers)
-                .addQualifierUnchecked(QualifierInstance.ANY).create();
+                .addQualifierUnchecked(QualifierInstance.ANY);
+        if (qualifiers.isEmpty()) {
```
Ladicek, line 208: I wonder (here and in `EventMetadataImpl`), what if `qualifiers` contains `@Any` (and nothing else)?

These are event qualifiers, so `@Any` is always present (that is, there's no difference between user specifying `@Any` and user _not_ specifying `@Any`), so perhaps `@Default` should be added too in this case?

mkouba: > I wonder (here and in `EventMetadataImpl`), what if `qualifiers` contains `@Any` (and nothing else)?
> 
> These are event qualifiers, so `@Any` is always present (that is, there's no difference between user specifying `@Any` and user _not_ specifying `@Any`), so perhaps `@Default` should be added too in this case?

Why? If you do `@Inject @Any Event<Foo> event` and then `event.fire(new Foo())` then the `@Default` should not be present, or?

manovotn: > I wonder (here and in `EventMetadataImpl`), what if `qualifiers` contains `@Any` (and nothing else)?

Had the same idea and my previous version of this code did exactly what you're suggesting but there were TCKs failures.
IIRC it was the `EventMetadataTest#testSimpleEvent` which doesn't expect `@Default` to be present when you use something like `@Inject @Any Event<SimpleEvent>` and fire event from there.

impl/src/main/java/org/jboss/weld/event/EventMetadataImpl.java
manovotn: Point taken, will do

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
Ladicek: OK, that makes sense. `@Inject Event<Foo>` is equivalent to `@Inject @Default Event<Foo>`, so the set of specified qualifiers is `{@Default}` and the set of event qualifiers is `{@Default, @Any}`.

So the qualifiers here are not _event qualifiers_, they are _specified qualifiers_.

The problem is that

> slightly different behaviour applies to observers with @Default qualifier
> ...
> Such observer will only be notified for events having either no qualifiers or only `@Default` qualifier

talks about _event qualifiers_ (which are never `{}` or `{@Default}`), even though it probably should talk about _specified qualifiers_ (which are also never `{}`, but may be `{@Default}`).

(Technically, that paragraph should probably never have been added. Looking at the history, this was an attempt to clarify that `@Inject Event<Foo>` is equivalent to `@Inject @Default Event<Foo>` due to how injection points work, but that attempt unintentionally added a new provision to the specification that triggers in some weird situation. I actually don't know what that situation is anymore, considering that `BM.getEvent()` should unconditionally add `@Default` to the set of specified qualifiers...)

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
```diff
@@ -200,9 +201,15 @@ public void fireEvent(Object event, Resolvable resolvable) {
...
+            resolvableBuilder.addQualifier(Default.Literal.INSTANCE);
+        }
+        return resolvableBuilder.create();
+
     }
 
     protected Resolvable buildEventResolvable(Type eventType, Annotation... qualifiers) {
```
Ladicek, line 215: I guess you also need to modify this method?

Ladicek:
Also maybe fix the comment at `BeanManagerImpl.EventInjectionPoint.getQualifiers()`, because for events, `ResolvableBuilder.create()` is never called with an empty set of qualifiers (hence this PR).

manovotn:
> Also maybe fix the comment at `BeanManagerImpl.EventInjectionPoint.getQualifiers()`, because for events, `ResolvableBuilder.create()` is never called with an empty set of qualifiers (hence this PR).

I still think the an empty set makes the most sense - this is just a representation of a synthetic IP in case it gets introspected (via `EventMetadata` I assume). The exact same happens for `InstanceInjectionPoint`.

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
manovotn: Hm, I wonder why we don't just delegate the method via `buildEventResolvable(eventType, Set.of(qualifiers));`.
We should probably do that.

manovotn:
> No, that's just an implementation comment. Why are you explicitly adding `@Default` to the `ResolvableBuilder` in this PR? Because we never call `ResolvableBuilder.create()` with an empty set of qualifiers when it comes to events.

Ok, I'll change it to:
> `@Default` and `@Any` are added as needed in ObserverQualifier#buildEventResolvable

impl/src/main/java/org/jboss/weld/event/ObserverNotifier.java
Ladicek: That makes sense to me :-)

Ladicek: Also, the reason why ... is probably to save a few allocations, I can't think of anything else.

Ladicek (approved):

PR #2910, opened 2024-02-13
Add support for generic field injection.

RussiaVk:
weld not support  generic field injection yet.
For example:
```
public interface IService {
	void print();
}
@ApplicationScoped
public class ServiceImpl1 implements IService {
	@Override
	public void print() {
		System.out.println("-------ServiceImpl1---------");
	}
}
public abstract class AbstractCDIClass<T extends IService> {
	@Inject
	private T service;
	
	@Test
	public void test() {
		service.print();
	}
}
@ApplicationScoped
public class CDIImpl1 extends AbstractCDIClass<ServiceImpl1> {
	@Deployment
	public static JavaArchive getJavaArchive() {
		return ShrinkWrap.create(JavaArchive.class).addAsManifestResource("META-INF/beans.xml").addClass(CDIImpl1.class)
				.addClass(ServiceImpl1.class);

	}
}
```
If we run CDIImpl1,it will throw following exception :
```
org.jboss.weld.exceptions.DefinitionException: WELD-001407: Cannot declare an injection point with a type variable: [BackedAnnotatedField] @Inject private AbstractCDIClass.service
	at AbstractCDIClass.service(AbstractCDIClass.java:0)
  StackTrace
	at org.jboss.weld.bootstrap.Validator.validateInjectionPointForDefinitionErrors(Validator.java:301)
	at org.jboss.weld.bootstrap.Validator.validateInjectionPoint(Validator.java:287)
	at org.jboss.weld.bootstrap.Validator.validateProducer(Validator.java:425)
	at org.jboss.weld.injection.producer.InjectionTargetService.validateProducer(InjectionTargetService.java:36)
	at org.jboss.weld.manager.InjectionTargetFactoryImpl.validate(InjectionTargetFactoryImpl.java:153)
	at org.jboss.weld.manager.InjectionTargetFactoryImpl.createInjectionTarget(InjectionTargetFactoryImpl.java:81)
```
Or refer to this [stackoverflow question](https://stackoverflow.com/questions/7856253/using-inject-with-generic-type).

Could you please share your thoughts about the potential significance and value of the changes proposed in my PR? I would greatly appreciate any feedback or insights you could provide.
Thank you for your attention and guidance.

### DIFF
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -85,7 +85,12 @@
             <artifactId>bcel</artifactId>
             <optional>true</optional>
         </dependency>
-
+        
+		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId> 
+		</dependency>
+        
         <!-- test dependencies -->
 
         <dependency>
--- a/impl/src/main/java/org/jboss/weld/injection/attributes/AbstractInferringInjectionPointAttributes.java
+++ b/impl/src/main/java/org/jboss/weld/injection/attributes/AbstractInferringInjectionPointAttributes.java
@@ -20,12 +20,14 @@ import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Member;
 import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
 import java.util.Set;
 
 import jakarta.decorator.Delegate;
 import jakarta.enterprise.inject.spi.Bean;
 
 import org.jboss.weld.annotated.enhanced.EnhancedAnnotated;
+import org.jboss.weld.manager.InjectionTargetFactoryImpl;
 import org.jboss.weld.serialization.BeanHolder;
 import org.jboss.weld.util.reflection.HierarchyDiscovery;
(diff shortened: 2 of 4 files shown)


RussiaVk:
> Weld is an implementation of the CDI specification and type variables in injection points are forbidden by the specification. Besides, you can just declare the injection point as `IService` or use a wildcard with bounds, both of which are permitted and should work fine.
> 
> > If an injection point type is a type variable, the container automatically detects the problem and treats it as a definition error.
> 
> Here's the relevant specification part - https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#legal_injection_point_typeI got it. Thank you for your explanation

> Weld is an implementation of the CDI specification and type variables in injection points are forbidden by the specification. Besides, you can just declare the injection point as `IService` or use a wildcard with bounds, both of which are permitted and should work fine.
> 
> > If an injection point type is a type variable, the container automatically detects the problem and treats it as a definition error.
> 
> Here's the relevant specification part - https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#legal_injection_point_types

I got it. Thank you for your explanation

PR #2946, opened 2024-03-13
WELD-2780 collection of changes for the invokers API

manovotn:
PR is still WIP and will require new Weld API version (see https://github.com/weld/api/pull/191)

Contains:
* Logging changes so that we use proper `Logger` interface
* Implement changes allowing to use Weld APIs without the need to type cast
* Remove leftover playground classes
  * After some more thinking I decided we should keep them in; might save a headache or two in the future
* Documentation changes - we should mention the new API in our docs
* Revisit deployment validation of lookups and see if we could perform it differently
  * This is IMO done correctly
  * invokers do not use classical `InjectionPoint` (which is correct due to their nature) and validating them separately through `WeldInstance` works just fine
  * Notably, this should respect the BM-per-archive structure that Weld uses and should correctly indicate if some of the beans isn't available in given bean archive
* Update the micro benchmark setup to use latest API
  * Done in `https://github.com/weld/weld-core-benchmarks/tree/invokableMethods`
  * This of course need Weld core release before it's mergeable

### DIFF
--- a/weld-lite-extension-translator/src/main/java/org/jboss/weld/lite/extension/translator/ExtensionInvoker.java
+++ b/weld-lite-extension-translator/src/main/java/org/jboss/weld/lite/extension/translator/ExtensionInvoker.java
@@ -30,7 +30,6 @@ class ExtensionInvoker {
         for (Class<? extends BuildCompatibleExtension> extensionClass : extensions) {
             SkipIfPortableExtensionPresent skip = extensionClass.getAnnotation(SkipIfPortableExtensionPresent.class);
             if (skip != null) {
-                // TODO only if the corresponding portable extension exists!
                 continue;
             }
 
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -30,7 +30,7 @@
     </developers>
 
     <properties>
-        <weld.api.bom.version>6.0.Beta2</weld.api.bom.version>
+        <weld.api.bom.version>6.0.Beta3</weld.api.bom.version>
         <gpg.plugin.version>3.2.0</gpg.plugin.version>
         <jboss.releases.repo.url>https://repository.jboss.org/nexus/service/local/staging/deploy/maven2/</jboss.releases.repo.url>
         <jboss.snapshots.repo.url>https://repository.jboss.org/nexus/content/repositories/snapshots/</jboss.snapshots.repo.url>
--- a/docs/reference/pom.xml
+++ b/docs/reference/pom.xml
@@ -112,7 +112,6 @@
                     <doctype>book</doctype>
                     <sourceDirectory>${basedir}/src/main/asciidoc</sourceDirectory>
                     <templateEngine>slim</templateEngine>
-                    <headerFooter>true</headerFooter>
                     <attributes>
                         <docinfo>true</docinfo>
                         <experimental>true</experimental>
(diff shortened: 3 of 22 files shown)


weld-lite-extension-translator/src/main/java/org/jboss/weld/lite/extension/translator/ExtensionInvoker.java
```diff
@@ -30,7 +30,6 @@ class ExtensionInvoker {
         for (Class<? extends BuildCompatibleExtension> extensionClass : extensions) {
             SkipIfPortableExtensionPresent skip = extensionClass.getAnnotation(SkipIfPortableExtensionPresent.class);
             if (skip != null) {
-                // TODO only if the corresponding portable extension exists!
```
Ladicek, line 33: This TODO is legit. If you don't feel like fixing it, I'd say you should keep it in place :-)

Ladicek:
I originally added the playground classes on purpose, because they allow much easier experimentation with various parts of the `MethodHandle`-based implementation of invokers. If you don't want to keep them around, feel free to delete them, but I predict that once there's a bug, you will go hunting for them :laughing:

manovotn:
> I originally added the playground classes on purpose, because they allow much easier experimentation with various parts of the `MethodHandle`-based implementation of invokers. If you don't want to keep them around, feel free to delete them, but I predict that once there's a bug, you will go hunting for them 😆

Yea, I don't feel strongly about keeping or removing these classes; it's just that I had this removal note in the JIRA issue so I added it here.
You're right tho that with some bugs, I'm pretty sure I'll be fairly nonplussed :)

weld-lite-extension-translator/src/main/java/org/jboss/weld/lite/extension/translator/ExtensionInvoker.java
manovotn: Hm, then I probably don't recall the case in which this would be an issue.
How would you be able to reference the class in the annotation without it existing?

Ladicek: That's ... a very good point :-) Unless there's a way for a portable extension to be present but not executed, you're right that it's enough.

manovotn: Actually, there is - you can have it on CP but not declared as `META-INF/services` which should cause it not to execute.
But I am not sure it then makes sense to have BCE with this annotation in place :thinking: 
Mainly because containers that cannot run PE will ignore them anyway (and you will never do the skipping) and those that can execute then and use this annotation should run the PE.

Ladicek: Yeah, the annotation should probably have been `@SkipInCdiFull` or something like that. The class-valued annotation member servers as a documentation link and that's probably enough...

manovotn:
I've decided to put back in the `*Playground` classes; they might save me a headache if future me can find them :)

Last but not least, I finally got to review the deployment validation of lookups.
Currently implemented approach is IMO the best way as the validation needs to respect whether given bean is present/visible in given bean archive (and therefore if the BM will be able to resolve it).
Furthermore, the fact that invoker lookups are *not* `InjectionPoint` is also right due to their nature - unlike standard bean IPs, here you can have several invokers for the same method with any combination of instance and args being lookup up in each of them.
Last but not least, there is no need to modify invokers "injection points" via portable extensions. Each integrator is supposed to register their invoker if they want one and they can fine tune it during creation.

This PR should therefore be merged once CI passes.

PR #2958, opened 2024-04-17
WELD-2771 Revisit Weld examples, update deps, verify they work

manovotn:
Examples are very outdated and don't even work with latest JDKs.
We should pick some minimal set of examples that demonstrate - EE usage, servlet usage, SE usage (, maybe the Groovy SE example?).

This can be done by keeping the Numberguess example for both, SE and JSF variants.

Things to do:
* ~~Remove JSF login and translator examples~~
  * ~~Remove mentions of the above in our docs~~
  * ~~Add a link to WFLY quickstarts as a source of more examples; many of them use CDI anyway~~
* ~~Update JSF Numberguess to work with WFLY, Tomcat and Jetty, update its README~~
* ~~Update test setup of JSF Numberguess to strip away Arq.~~
  * ~~Copied over from the same example under WFLY - kept some basic assertions against running deployed application~~
* ~~Update and test SE Numberguess sample~~
* ~~Update and test SE Groovy example~~
* ~~Scan Weld docs for any other mention of example usage to make sure we don't have leftovers~~
* ~~Think if and how we can test this in CI~~
  * ~~We can surely test a build but deployments might be tricky - each setup requires starting and stopping a container~~


manovotn:
The failures are related to update of some dependencies which have min. JDK version set to 17
We could extract servlet testing from `Weld CI / Weld In-container Tests - JDK 11` into a separate job or simply skip that part if JDK is 11.
As for `Weld CI / CDI TCK - JDK 11`, this is because we are currently relying on the pre-split version of TCK which has web tests as well and is therefore affected as well. I need to re-check the status of tests moving into umbrella TCK and if there is a release we can consume already.

manovotn:
I've adjusted the CI in the following way:
* Examples are now only built (not tested yet) and this is done with JDK 17 only
  * We should still look into how and if we can test it too, which is TBD
* CDI TCKs are run with all TCKs and should pass since the web profile is now not present at all
  * I didn't find any release of the platform TCK yet but we need to reintroduce that ASAP
  * I should at least try some local setup with snapshots and store it under a JIRA to keep track
* In-container tests on JDK 11 now skip servlet modules which would require JDK 17+
  * JDK 17 and above tests are not affected in any way

manovotn:
The last thing needed in this PR is testing the examples (running their tests) - I'll give it a spin next week to see if we can get at least some simple setup running.

manovotn:
I've added CI setup that performs a Maven build on examples and then starts WFLY as background process, deploys JSF numberguess and runs tests against it. This should give us at least basic level of confidence.
We could add the same with Tomcat/Jetty but the version keeping would be more difficult and job longer due to re-download of servlets on every run. With WFLY it's easier as we already have the patched version prepared from initial job.

PR #2959, opened 2024-04-25
WELD-2787 Remove legacy integration points for Jetty

manovotn:
Based on the discussion in linked Arq. jetty adapter, I think the integration can be simplified to the following.
Note that removing these legacy parts in 5.x doesn't really matter as they play no role in current Jetty versions anyway and you cannot use this version of Weld with older versions of Jetty.


environments/servlet/core/src/main/java/org/jboss/weld/environment/jetty/JettyContainer.java
```diff
@@ -62,21 +59,14 @@ protected String classToCheck() {
...
     @Override
     public void initialize(ContainerContext context) {
         try {
             ServletContext servletContext = context.getServletContext();
             String mode = (String) servletContext.getAttribute(JETTY_CDI_ATTRIBUTE);
-            if (mode == null) {
-                mode = DECORATING_LISTENER_MODE;
-            }
             switch (mode) {
                 case CDI_SPI_DECORATOR_MODE:
```
manovotn, line 71: According to https://github.com/arquillian/arquillian-container-jetty/issues/197#issuecomment-2077383639, this is Jetty's internal class. I suppose the only purpose it serves here is telling apart an undefined state (i.e. throwing an exception) versus this state which, for us, is just a no-op.
I'll add a code comment for my future self to remember that.

manovotn: I've added a code comment clarification.

PR #2961, opened 2024-04-25
WELD-2786 OM configurator without overriden notify method should respect Reception settings

manovotn:


impl/src/main/java/org/jboss/weld/bootstrap/events/configurator/ObserverMethodConfiguratorImpl.java
```diff
@@ -82,7 +83,11 @@ public ObserverMethodConfiguratorImpl(Extension extension) {
     public ObserverMethodConfiguratorImpl(ObserverMethod<T> observerMethod, Extension extension) {
         this(extension);
         read(observerMethod);
-        notifyWith(e -> observerMethod.notify(e));
+        if (observerMethod instanceof ObserverMethodImpl<?, ?>) {
```
manovotn, line 86: This isn't very good as it relies on our impl of OM underneath while technically the impl can be arbitrary.
However, I don't think we can achieve this with any other setup anyway as the public API for `ObserverMethod` doesn't allow us to `setReception()`. Thoughts or ideas @mkouba?

mkouba: I think that it's OK because it should only work for configurators which are based on non-synthetic observer methods, i.e. observer methods backed by a Java method of a class bean which is always a Weld implementation of OM.

manovotn: > i.e. observer methods backed by a Java method of a class bean which is always a Weld implementation of OM.

That is a good point, thank you.

PR #2977, opened 2024-05-09
WELD-2788 Use Lock instead of synchronized in ApplicationScopedContextualInstanceStrategy to be virtual thread-friendly

ljnelson:
This pull request uses a `Lock` instead of a `synchronized` block to prevent thread pinning in virtual thread scenarios. See [WELD-2788](https://issues.redhat.com/browse/WELD-2788) for details.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/ContextualInstanceStrategy.java
+++ b/impl/src/main/java/org/jboss/weld/bean/ContextualInstanceStrategy.java
@@ -18,6 +18,8 @@ package org.jboss.weld.bean;
 
 import java.lang.annotation.Annotation;
 import java.util.Set;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.context.ConversationScoped;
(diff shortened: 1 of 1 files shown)


manovotn (approved):

mkouba (approved):
Looks good but there are a lot of places with `synchronized` in the Weld codebase: https://github.com/search?q=repo%3Aweld%2Fcore%20synchronized&type=code such as https://github.com/weld/core/blob/0b806e8ebfbfd3d04fc32d61857ef6f97c89cbe4/impl/src/main/java/org/jboss/weld/util/LazyValueHolder.java#L44. In other words, it would probably require some deeper analysis and "global" revision to make Weld really virtual thread-friendly...

manovotn:
> Looks good but there are a lot of places with `synchronized` in the Weld codebase: https://github.com/search?q=repo%3Aweld%2Fcore%20synchronized&type=code such as
> 
> https://github.com/weld/core/blob/0b806e8ebfbfd3d04fc32d61857ef6f97c89cbe4/impl/src/main/java/org/jboss/weld/util/LazyValueHolder.java#L44
> . In other words, it would probably require some deeper analysis and "global" revision to make Weld really virtual thread-friendly...

I also did a `grep` for synchronized blocks yesterday and there is surely more.
However, we are not running primarily in an environment with virt. threads so nailing these down would be partially guesswork.
I expect we will see more of these issues and I am open to fixing these problems as they pop up.

ljnelson:
There are _absolutely_ more. This was one in particular where badly-behaved user code could have an impact on the container performance so it seemed worth addressing in isolation. Thanks for the merge.

PR #2979, opened 2024-05-15
Weld-2772 Allow to register BCE in Weld SE without discovery

manovotn:
Add the ability to manually register BCE in Weld SE without the need to perform discovery.

I believe this should eventually bubble up to `SeContainerInitializer` and be part of the CDI SE API (cc @Ladicek; this is something I mentioned a [long] while ago)
We already have similar methods in place for portable extensions.


Ladicek:
I don't really know much about CDI SE, but that makes sense to me.

PR #2980, opened 2024-05-15
Weld-2772 Allow to register BCE in Weld SE without discovery

manovotn:
Weld 6 counterpart of https://github.com/weld/core/pull/2979


mkouba (approved):
Looks good!

PR #2995, opened 2024-06-21
Release setup changes

manovotn:
Updated:
Introduced similar setup to other Weld projects, make use of GH secrets.
Moved the gradle build script to core repo - this isn't needed but I thought I'd be easier than having it in a separate repo where it was up until now.
Setup CI release job so that it builds the dist and exposes the result as a downloadable zipped file.


manovotn:
This should be a good starter setup - most of release work would be done via CI.

However, there is a release build issue with Weld shaded artifacts that needs to be resolved first. I am looking into that.

PR #3034, opened 2024-08-19
My attempt to fix e.getClass().equals(BeanManager.class), where Strin…

dk2k:
…g is compared with Class. However, two parts combined with && looks ridiculous. There is no class, which is subtype of both NameClassPair and BeanManager. Please consider this PR as bug report


manovotn (approved):
Thanks, this does look like an oversight.

PR #3037, opened 2024-08-21
WELD-2796 For built-in beans, the proxy's originalClass should be the implemented type

manovotn:
This will affect the behavior in WFLY where we retrieve a module based on the original class, see https://github.com/wildfly/wildfly/blob/main/weld/subsystem/src/main/java/org/jboss/as/weld/services/bootstrap/ProxyServicesImpl.java#L187

In case of `ServletContextBeanTest`, one can observe that with this change we would now place the proxy into user deployment instead of into Weld module. As a result, such proxy cannot be reused in another deployments and repeated test executions on a running WFLY instance will trigger repeated proxy creations.


manovotn:
Closing in favor of https://github.com/weld/core/pull/3038 and https://github.com/weld/core/pull/3039 after further discussion on the issue.

PR #3055, opened 2024-09-30
WELD-2798 Correct injection into injected @Resource fields; add automated test

manovotn:
JIRA - https://issues.redhat.com/browse/WELD-2798

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Set up JDK
-        uses: actions/setup-java@v3.12.0
+        uses: actions/setup-java@v4.2.1
         with:
           java-version: 11
           distribution: 'temurin'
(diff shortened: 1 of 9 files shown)


impl/src/main/java/org/jboss/weld/injection/ResourceInjectionProcessor.java
```diff
@@ -90,8 +90,7 @@ protected <T> Set<ResourceInjection<?>> createResourceInjections(Bean<?> declari
         }
 
         Class<? extends Annotation> marker = getMarkerAnnotation(processorContext);
-
-        final Collection<EnhancedAnnotatedField<?, ? super T>> fields = type.getDeclaredEnhancedFields(marker);
+        final Collection<EnhancedAnnotatedField<?, ? super T>> fields = type.getEnhancedFields(marker);
         final Collection<EnhancedAnnotatedMethod<?, ? super T>> methods = type.getDeclaredEnhancedMethods(marker);
```
manovotn, line 94: For methods, the `EnhancedAnnotatedType` seems to have different logic and "declared" methods contain all of the methods of the bean while trying to use `getEnhancedMethods(marker)` fails as that is a map containing only entries for a subset of annotations.

I am not sure if that's an optimization or an oversight; I will keep poking around that.
Nonetheless, the fix in this PR is sufficient to fix the problem in the linked issue.

mkouba (approved):

PR #3066, opened 2024-10-09
WELD-2799 Implement validation for unproxyability of sealed types

manovotn:
JIRA - https://issues.redhat.com/browse/WELD-2799


manovotn:
The failure is due to impsort plugin not being setup for JDK 17 - which is something coming from Weld parent.
I'll perform a release there and then update the version as part of this PR.

PR #3069, opened 2024-10-26
[WELD-2800] Don't use jar cache in ServiceLoader.loadServiceFile

WolfgangHG:
This should resolve https://issues.redhat.com/browse/WELD-2800

`java.net.URL.openStream` uses a jar cache. Thus the jar file is kept in memory as long as the Glassfish server is alive, and thus the jar file in "glassfish\domains\domain1\applications" cannot be deleted on windows when the app is undeployed.

Workaround: set `URLConnection.setUseCaches(false)` before accessing the InputStream:

```java
URLConnection jarConnection = serviceFile.openConnection();
jarConnection.setUseCaches(false);
is = jarConnection.getInputStream();
```

I pushed it to the 5.1 branch as GlassFish 7.0.18 uses Weld 5.1.3.Final, and I hope that this fix makes it to next Glassfish.

I don't have much git experience and never pushed to a branch, so please be patient with me if I did this wrong ;-)

### DIFF
--- a/impl/src/main/java/org/jboss/weld/util/ServiceLoader.java
+++ b/impl/src/main/java/org/jboss/weld/util/ServiceLoader.java
@@ -175,7 +175,8 @@ public class ServiceLoader<S> implements Iterable<Metadata<S>> {
         InputStream is = null;
         try {
             URLConnection jarConnection = serviceFile.openConnection();
-            //Don't cache the file (avoids file leaks on GlassFish).
+            //Don't cache the file; in combination with Windows OS, this can cause file leaks on some EE servers (GlassFish)
+            // See WELD-2800 for more details
             jarConnection.setUseCaches(false);
             is = jarConnection.getInputStream();
             BufferedReader reader = new BufferedReader(new InputStreamReader(is, "UTF-8"));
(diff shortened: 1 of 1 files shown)


manovotn:
Hello,

I am currently on PTO and unless @mkouba has some spare cycles, it won't get reviewed right away.
I'll keep it on my TODO list once I am back though :)

manovotn:
Interesting find and thank you for the PR @WolfgangHG!
We will want to address this in both, `5.1` and `master` branch which will both have this problem.

> thus the jar file in "glassfish\domains\domain1\applications" cannot be deleted on windows when the app is undeployed.

So this is a Windows-only problem but does that mean that what we do is wrong and should be done differently?
Turning off the caching seems like a workaround more so than a proper solution.
I don't have anything against turning off the caching as such (I don't think it does much here), just curious as you seem to have done quite some investigation already.

WolfgangHG:
I can only provide a bit of information that I f
are my findings when 
openStream is a shortcut to openConnection().getInputStream(), and openConnection() returns a java.net.URLConnection, which has a subclass sun.net.www.protocol.jar.JarURLConnection (code e.g. [here](https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/net/www/protocol/jar/JarURLConnection.java)). This class uses a cache by default (see method connect). This cache causes the file leak. There is no API to flush the cache.

manovotn (approved):

PR #3088, opened 2024-11-27
WELD-2804 Remove deprecated code parts.

manovotn:
Deprecate and mark for removal `WeldContainer#instance()`; change all remaining SE tests using this method. 
Remove deprecation from methods obtaining qualifiers on `EnhancedAnnotated` - those have been deprecated since early versions but had meanwhile become core part of this API.


mkouba (approved):
Looks good. Let's prune the deprecated stuff!

PR #3122, opened 2025-02-27
bump to CDI 5.0 + add support for sealed classes to the implementation of CDI Lang Model

Ladicek:


Ladicek:
The first commit depends on https://github.com/weld/api/pull/225 and again, doesn't necessarily has to be correct.

The second commit should be good.

manovotn:
I've sent https://github.com/weld/core/pull/3132 to upgrade project version (and similar for API).
Though we'd still need some Alpha releases to avoid the snapshots :)

tests-arquillian/src/test/java/org/jboss/weld/tests/extensions/SimpleExtension.java
```diff
@@ -34,7 +34,7 @@ public static boolean isObservedBeforeBeanDiscovery() {
         return observedBeforeBeanDiscovery;
     }
 
-    public void observeBeforeShutdown(@Observes BeforeShutdown beforeShutdown, BeanManager beanManager) {
+    public void observeBeforeShutdown(@Observes BeforeShutdown beforeShutdown, ELAwareBeanManager beanManager) {
```
manovotn, line 37: ATM, Weld is checking strictly for `BeanManager` type as an optional argument in observer methods (not even `WeldManager` works).
I am not sure why that is the case but it should be fairly simple to fix.
I have created https://issues.redhat.com/browse/WELD-2811 to track this and will fix it for 6.x as well as 7.x

Ladicek: I would guess that at least specification defined types should work, so not just `BeanManager`, but also `BeanContainer` and `ELAwareBeanManager`.

manovotn: I agree. I've already sent PRs for this, see https://github.com/weld/core/pull/3134

manovotn:
I have updated this branch and rebased it onto `master` so that it has the fix for lifecycle OM and `BeanManager` param.
It is still running with `SNAPSHOT` dependencies, so failures are expected. 

Though running it locally, I can now see all the tests passing :+1:

Ladicek:
You really should feel free to bump to CDI 5.0 API / Weld 7.0 API elsewhere, if that suits you better. The core value of this PR has always been the implementation of the sealed classes support in the lang model.

manovotn:
Updated Weld API and CDI TCK versions; moved from draft to ready for review.
CI should now pass.

manovotn:
Uff, these failures are related but I am not sure how much we can do now.

There are calls to `BM#getELResolver()` which is removed in CDI 5. Now, these calls can be either from WFLY integration code or faces impl, none of which can be easily modified. Even WFLY doesn't yet have a version for EE 11, so none that would reliably operate on CDI 4.1 to have access to the new EL API :thinking: 

I'll poke around WFLY internals to at least get clarity on what the problems are and create some tracking issue for them...

manovotn:
Few notes to myself after poking around this today:
* Mojarra has this correctly updated for EE11 version (4.1.x mojarra) 
  * Standard WFLY doesn't use this version yet - preview does
    * We could use the preview here and workaround it but it is only  a part of the problem
* WFLY internals also need changing
  * There is integration code using `BM#getELResolver()` and that I cannot change because the codebase is still EE10 based (so no CDI 4.1)
  * Apart from that, many modules will also need to declare deps on CDI EL API as they consume Weld artifacts that now implement it too.

To begin with, we should use the preview WFLY artifact but that does expose other problems such as adding a JMS queue that some tests are using. Apparently, the preview WFLY has different configuration files and the relevant subsystem is no longer pre-configured.
There are some example files that contain the configuration and it seems those work fine. So we either need to setup CI to copy that over or find a way to start WFLY with a custom file that's not in the default dir.

manovotn (approved):

PR #3136, opened 2025-03-18
Attempt to use latest Jakarta Platform TCKs

manovotn:
I assume this will crash as html unit and selenium were updated...


manovotn:
Superseeded by https://github.com/weld/core/pull/3138 that also updates CI setup with Preview WFLY

PR #3138, opened 2025-03-24
Move CI setup to WFLY preview; drop profile that patched Jakarta deps in WFLY manually. Update Jakarta TCK version

manovotn:


manovotn:
After upgrading the platform TCK version, the failing test was:

> [ERROR] Failures: 
[ERROR]   LongRunningConversationPropagatedByFacesContextTest>Arquillian.run:138->testConversationPropagatedAjax:136 » DevTools You are using a no-op implementation of the CDP. The most likely reason for this is that Selenium was unable to find an implementation of the CDP protocol that matches your browser. Please be sure to include an implementation on the classpath, possibly by adding a new (maven) dependency of `org.seleniumhq.selenium:selenium-devtools-vNN:4.24.0` where `NN` matches the major version of the browser you're using.
Build info: version: '4.24.0', revision: '748ffc9bc3'
System info: os.name: 'Linux', os.arch: 'amd64', os.version: '6.8.0-55-generic', java.version: '21.0.3'
Driver info: DevTools Connection

I have added a commit with an exclusion for this test for now.
More information can be found in https://github.com/jakartaee/platform-tck/issues/1228#issuecomment-2725633366

PR #3314, opened 2026-03-05
WELD-2834 Optimize in which cases Weld keeps reference to dependent instnace in CreationalContext

manovotn:
An attempt to bring Weld closer to what ArC does - avoid keeping dependent bean references in `CreationalContext` unless the bean has some form of pre-destroy callback.

Without this PR, Weld always puts bean's interceptor/decorator into the CC of that given bean.
This in turn means that the bean then always has dependents and cannot be further optimized.
In theory, it should be possible to circumvent this as we only really need the reference if there is some destroy logic that we'll need to execute.


manovotn:
FTR, I have also tested this with the reproducer from WELD-2834 and the memory leak no longer shows.
The application is still wrong in that it *should* handle bean destruction properly but at least we aren't leaking.

PR #3337, opened 2026-05-04
Changes for Alpha6 release of CDI API

manovotn:
JIRA - https://redhat.atlassian.net/browse/WELD-2836


manovotn:
There might be issues with the incontainer tests. WFLY changed to EE 11 and this PR doesn't reflect that.

manovotn:
> There might be issues with the incontainer tests. WFLY changed to EE 11 and this PR doesn't reflect that.

Rebased onto master which has the WFLY setup corrected - the CI should now pass.

manovotn:
Added a small commit that improves logging in order to be able to distinguish failure in bean destruction from failure in calling its `close()` method.

PR #3361, opened 2026-06-02
Add support for injecting an EntityAgent

### LINKED ISSUE #3367: Jakarta Persistence 4.0 Integration: add capability of injecting an EntityAgent
JPA 4 adds an `EntityAgent` to provide a “stateless alternative to an EntityManager". And it also defines ways of obtaining it in the EE environment. It is quite similar to what `@PersistenceContext` does for the `EntityManager`.

More details and links in Weld API issue - Weld API issue - https://github.com/weld/api/issues/259

marko-bekhta:
Hey 🙂 👋🏻 

This one is based on 
* https://github.com/weld/api/pull/258

so I'll open it as a draft for now as we'd need the release of API first.

### DIFF
--- a/impl/src/main/java/org/jboss/weld/injection/ResourceInjectionFactory.java
+++ b/impl/src/main/java/org/jboss/weld/injection/ResourceInjectionFactory.java
@@ -111,6 +111,7 @@ public final class ResourceInjectionFactory implements Service, Iterable<Resourc
     private void initializeProcessors() {
         resourceInjectionProcessors.add(new PersistenceUnitResourceInjectionProcessor());
         resourceInjectionProcessors.add(new PersistenceContextResourceInjectionProcessor());
+        resourceInjectionProcessors.add(new PersistenceAgentResourceInjectionProcessor());
         resourceInjectionProcessors.add(new ResourceResourceInjectionProcessor());
         resourceInjectionProcessors.add(new WebServiceResourceInjectionProcessor());
     }
@@ -195,6 +196,47 @@ public final class ResourceInjectionFactory implements Service, Iterable<Resourc
             // ugly hack that allows application servers to support hibernate session injection while at the same time
             // the injection points are validated by Weld for invalid types (required by the TCK)
             return !apiAbstraction.SESSION_NAME.equals(getRawType(getResourceInjectionPointType(member)).getName());
+        }
+    }
+
+    /**
+     * JPA persistence agent resource injection processor.
+     */
+    private static class PersistenceAgentResourceInjectionProcessor extends
+            ResourceInjectionProcessor<JpaInjectionServices, PersistenceApiAbstraction> {
+
+        @Override
+        protected <T> ResourceReferenceFactory<T> getResourceReferenceFactory(InjectionPoint injectionPoint,
+                JpaInjectionServices injectionServices, PersistenceApiAbstraction apiAbstraction) {
+
+            if (!injectionPoint.getType().equals(apiAbstraction.ENTITY_AGENT_CLASS)) {
+                throw BeanLogger.LOG.invalidResourceProducerType(injectionPoint.getAnnotated(),
+                        apiAbstraction.ENTITY_AGENT_CLASS);
+            }
+            return Reflections.<ResourceReferenceFactory<T>> cast(injectionServices
+                    .registerPersistenceAgentInjectionPoint(injectionPoint));
+        }
+
+        @Override
+        protected Class<? extends Annotation> getMarkerAnnotation(PersistenceApiAbstraction apiAbstraction) {
+            return apiAbstraction.PERSISTENCE_AGENT_ANNOTATION_CLASS;
+        }
+
+        @Override
+        protected PersistenceApiAbstraction getProcessorContext(BeanManagerImpl manager) {
+            return manager.getServices().get(PersistenceApiAbstraction.class);
+        }
+
+        @Override
+        protected JpaInjectionServices getInjectionServices(BeanManagerImpl manager) {
+            return manager.getServices().get(JpaInjectionServices.class);
+        }
+
+        @Override
+        protected boolean accept(AnnotatedMember<?> member, PersistenceApiAbstraction apiAbstraction) {
+            // ugly hack that allows application servers to support hibernate stateless session injection while at the same time
+            // the injection points are validated by Weld for invalid types (required (?) by the TCK)
+            return !apiAbstraction.STATELESS_SESSION_NAME.equals(getRawType(getResourceInjectionPointType(member)).getName());
         }
     }
 
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -30,7 +30,7 @@
     </developers>
 
     <properties>
-        <weld.api.bom.version>7.0.Beta1</weld.api.bom.version>
+        <weld.api.bom.version>7.0.Beta2</weld.api.bom.version>
         <gpg.plugin.version>3.2.8</gpg.plugin.version>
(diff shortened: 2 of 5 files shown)


manovotn:
Hello @marko-bekhta and thank you for the PR.

We will need to create tracking issues for this (both, core and api) - GH issues will do.
More importantly though, I am definitely not savvy  in JPA so I don't know whether this actually does what you hope it to do. Is there a way we can test this? Should there be some coverage on Weld side or is there coverage elsewhere and you tried it with the SNAPSHOT version?

marko-bekhta:
hey hello @manovotn 👋🏻 🙂 

Sorry, it took me a while to get to this email 🙈 
I've created https://redhat.atlassian.net/browse/WELD-2837 (should I create another one for API/JUnit extension?)

good questions!

> I don't know whether this actually does what you hope it to do. 

so it should be somewhat similar to what `@PersistenceContext` does for an enity manager, `@PersistenceAgent` should inject an `EntityAgent`. (`EntityManagerFactory` has the capability to create instances of this new `EntityAgent`).

> Is there a way we can test this? Should there be some coverage on Weld side or is there coverage elsewhere and you tried it with the SNAPSHOT version?

that's where I may have got lost a bit, my understanding was that the SPI implementation lives in WildFly, so we'd need to make changes to it before we could do a full integration test ? Or more likely I missed something important and we actually can test it within Weld impl itself 🤞🏻 🫣 🙂 ?

As to how I tried to test these changes for now: I also updated the junit 5 extension to use the snapshots based on these two PRs and used the snapshot of that JUnit extension to run some tests (Jakarta Data TCK to be more precise 🫣🙂). The way I used it there was:

```java
@ExtendWith(WeldJunit5Extension.class)
public class StandaloneEntityTests extends EntityTests {

    @WeldSetup
    public WeldInitiator weld = WeldInitiator.from(EntityTests.class,
                    _Boxes.class,
                    _MultipleEntityRepo.class,
                    _AsciiCharacters.class,
                    _NaturalNumbers.class,
                    _PositiveIntegers.class,
                    EntityManagerFactoryProducer.class)
            .activate(RequestScoped.class)
            .inject(this)
            .setPersistenceUnitFactory(ip -> CDI.current().select(EntityManagerFactory.class).get())
            .setPersistenceContextFactory(ip -> CDI.current().select(EntityManagerFactory.class).get().createEntityManager())
            .setPersistenceAgentFactory(ip -> CDI.current().select(EntityManagerFactory.class).get().createEntityAgent())
            .build();

}
```
where the EMF producer is something like:
```java
@ApplicationScoped
public class EntityManagerFactoryProducer {

    @Produces
    @ApplicationScoped
    public SessionFactory createEntityManagerFactory(BeanManager beanManager) {
        HashMap<String, Object> properties = new HashMap<>();
        // set some EMF properties ...
        SessionFactory sessionFactory
                = Persistence.createEntityManagerFactory("jakarta-data-tck", properties)
                .unwrap(SessionFactory.class);
        return sessionFactory;
    }
}
```
(let me know if I should go ahead and also open a PR for weld-testing to include these updates to the extension)


I hope this clarifies things a bit 🤞🏻

manovotn:
> I've created https://redhat.atlassian.net/browse/WELD-2837 (should I create another one for API/JUnit extension?)

I am trying to move away from JIRA and use GH issues which should now be enabled on all Weld repos.
I don't mind whichever way we create it but there should be a linkage between the sent PRs, that's all that matters.

> that's where I may have got lost a bit, my understanding was that the SPI implementation lives in WildFly, so we'd need to make changes to it before we could do a full integration test ?

That is very likely. Lots of these things require juggling quite a few snapshots to get it going unfortunately.
Does it actually make sense to test outside of WFLY (or other EE server for that matter)?

The JUnit test you did will just verify the mock injection, right?
I guess that's also good in terms of whether the plumbing works, the details are then up to the impl in EE server 🤔 

> As to how I tried to test these changes for now: I also updated the junit 5 extension to use the snapshots based on these two PRs

Ah, that's good reminder, The JUnit extension is something I didn't yet get to - I should update that to CDI 5/Weld 7 as well.

> (let me know if I should go ahead and also open a PR for weld-testing to include these updates to the extension)

Sure, if you have the code, that would be nice.
Set it up as a draft until we get the new API/core versions in.

> I hope this clarifies things a bit 

It does indeed. This week I have some time to look into this, so I can move these PRs forward.

impl/src/main/java/org/jboss/weld/injection/ResourceInjectionFactory.java
```diff
@@ -198,6 +199,40 @@ protected boolean accept(AnnotatedMember<?> member, PersistenceApiAbstraction ap
         }
     }
 
+    /**
+     * JPA persistence agent resource injection processor.
+     */
+    private static class PersistenceAgentResourceInjectionProcessor extends
```
manovotn, line 205: More of a question than anything else - all other `Processors` defined here also override the `accept` method. This seems to be a workaround because of JPA TCKs, do you not need it here as well?

My understanding is that there is (was?? maybe it is gone??) a requirement where a certain injection annotation, say `@PersistenceContext`, can only be declared for a certain type, like `EntityManager`.
Putting it onto any other type should result in a validation failure.
Overriding the `accept` method let's Weld skip validation for certain IPs, such as:

```java
@PersistenceContext
Session s; // Hibernate session
```

marko-bekhta: ahh, that's a good point and a reminder ... we don't have any tests for CDI in the JPA TCK (the standalone one we run as we work on changes) ... but we probably need to add something to the platform TCK to test this new CDI/JPA integration ...

manovotn:
@marko-bekhta Weld API Beta2 was just released and will land in Central shortly (~1h) - if you can bump that version in the parent POM file, you can then un-draft this and we can move it forward.
And feel free to squash the commits into a single one.

I assume you want some Weld Core release ASAP as well so that you work with it, right?

marko-bekhta:
nice ! thanks 🙂 🥳 

ohh if we can get it out + the testing extension that would be great 😊🙂
(for now, we just run the Data TCK from local snapshots, but if we have a published version, we'd be able to integrate that into CI 🙂 )

Oh Central is nice today: https://repo1.maven.org/maven2/org/jboss/weld/weld-api/7.0.Beta2/ 🥳

manovotn:
> ohh if we can get it out + the testing extension that would be great 😊🙂

Sure, I'll try to do both, Core and JUnit extension, within this week.

manovotn:
Hmm, the test failures are related - those are in-container tests (against WFLY).
The test in question is `org.jboss.weld.tests.resources.EMFFactoryTest`.

> Caused by: java.lang.ClassNotFoundException: jakarta.persistence.EntityManager$CreationOption

I suppose this is another new class in the JPA and something is trying to load it while WFLY still has the old version.
I am not sure we can fix now; even sneaking in new JPA into WFLY might just give more errors but might be worth a shot.

marko-bekhta:
ahh I bet it's this change:

https://github.com/jakartaee/persistence/commit/3a416efc2c24f4715c1f5534584be7012d3f2295#diff-3ed47919f6e32958add334ab278cb53dde8cb59ac1c5f07d1348b0a39f604c34L180-R182

hmm ... 3.2 and 4.0 aren't compatible but we cannot just patch WildFly (and replace the jpa api), we run the tests against... since that also means we'd need to bring a new version of ORM 😖

manovotn:
I've added `@Ignore` to the test with a comment. I don't think we can do anything else now.

manovotn:
Thank you @marko-bekhta!

PR #3364, opened 2026-06-03
Allow contravariant wildcards in Event and covariant wildcards in Instance

### LINKED ISSUE #3362: Validation of wildcard `Event` and `Instance` injection is too strict
ATM (and since forever) Weld has been rejecting injection of any `Event`/`Instance` with wildcard in it.
However, as has been pointed out recently in https://github.com/jakartaee/cdi/issues/888#issuecomment-4578091411, this is theoretically incorrect.
I say theoretically, because there is no actual added value in being able to `@Inject Event<? super Something>` versus just `@Inject Event<Something>`. You'd still fire the same events from both.

That said, `Event` is naturally contravariant and should therefore allow `? super X`, while `Instance` is covariant and should allow `? extends Y`.
At the same time, we want to keep rejecting all the other variants such as unbound wildcard.

This change should go to Weld 6 and 7 alike.
CDI TCK - https://github.com/jakartaee/cdi-tck/pull/727
CDI - https://github.com/jakartaee/cdi/pull/982

manovotn:
Fixes #3362

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/builtin/AbstractFacade.java
+++ b/impl/src/main/java/org/jboss/weld/bean/builtin/AbstractFacade.java
@@ -20,6 +20,7 @@ import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.lang.reflect.WildcardType;
 import java.util.Set;
 
 import jakarta.enterprise.context.spi.CreationalContext;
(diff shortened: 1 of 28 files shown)


manovotn:
The failures are expected, there are TCK changes to be made, see https://github.com/jakartaee/cdi-tck/pull/727
This PR should probably wait for the next TCK Beta release so that we won't need to have exclusions.

manovotn:
Superseded by #3462

PR #3500, opened 2026-09-03
WELD-2836 Relocate client proxies for public third-party normal-scope…

### LINKED ISSUE #3502: Relocate client proxies for public third-party normal-scoped beans to avoid requiring `--add-opens`
On the Java module path, Weld requires `--add-opens` (WELD-2828 fixed the `--add-reads`) on a third-party module whenever an application injects a normal-scoped bean whose type lives in that module — e.g. a `@RequestScoped` producer that produces a public interface owned by another named module. Weld already avoids this for `java.*` and `jakarta.*` types by relocating their proxies into `org.jboss.weld.generated.proxies.*`. This request is to generalize that relocation to any public type in a named module whose package is not opened to Weld, eliminating the `--add-opens` requirement for those beans.

A CDI producer exposes a Vert.x type as a normal-scoped bean:

```java
@Produces
@RequestScoped
RoutingContext routingContext() { ... }   // io.vertx.ext.web.RoutingContext, module io.vertx.web
```

`RoutingContext` is a public interface, exported from the `io.vertx.web` module. Because `@RequestScoped` is a normal scope, Weld generates a client proxy. On the module path this currently fails unless the application launches with:

```
--add-opens io.vertx.web/io.vertx.ext.web=weld.core.impl
```

By contrast, injecting a jakarta.ws.rs-owned normal-scoped type works with no flags, because those proxies are relocated.

Root cause (verified against source)

weld-core-impl, org.jboss.weld.bean.proxy:

1. `ProxyFactory.getProxyClass()` relocates a proxy into Weld's own namespace only for java/jakarta prefixes:

```java
if (proxyClassName.startsWith(JAVA)) {
    proxyClassName = proxyClassName.replaceFirst(JAVA, WELD_PROXY_PREFIX);
} else if (proxyClassName.startsWith(JAKARTA)) {
    proxyClassName = proxyClassName.replaceFirst(JAKARTA, WELD_PROXY_PREFIX);
}

```
   For a producer, `createCompoundProxyName()` forces the proxy package to the produced type's package (mostSpecificClass.getPackage().getName()), so a `RoutingContext` proxy is named in io.vertx.ext.web.* and never matches the prefix check.
2. `WeldDefaultProxyServices.defineClass()` then dispatches purely on the proxy's name prefix:

```java
if (className.startsWith(ProxyFactory.WELD_PROXY_PREFIX)) {
    return defineWithClassLoader(className, ...);        // Weld's own CL — needs nothing from target module
} else {
    return defineWithMethodLookup(className, ..., originalClass, ...);
}
```

   Non-relocated names go through `defineWithMethodLookup`, which does `MethodHandles.privateLookupIn(targetType, ...)` and defines the proxy into the target package. That is what forces:
   - opens: `io.vertx.web` must open `io.vertx.ext.web` to `weld.core.impl` (for `privateLookupIn` + `defineClass`), and
   - reads: the target module must read weld.core.impl and weld.api so the generated proxy can link the Weld interfaces it implements.

The `defineWithClassLoader` path (used by relocated `java`/`jakarta` proxies) needs neither.

jamezp:
…d beans to avoid requiring --add-opens.

resolves #3502

### DIFF
--- a/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
+++ b/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
@@ -403,6 +403,10 @@ public class ProxyFactory<T> {
             proxyClassName = proxyClassName.replaceFirst(JAVA, WELD_PROXY_PREFIX);
         } else if (proxyClassName.startsWith(JAKARTA)) {
             proxyClassName = proxyClassName.replaceFirst(JAKARTA, WELD_PROXY_PREFIX);
+        } else if (bean != null && shouldRelocateProxy(bean.getBeanClass())) {
+            // Target lives in a named module that doesn't open its package to Weld.
+            // Define the proxy into Weld's own package to avoid the opens requirement to the named module.
+            proxyClassName = WELD_PROXY_PREFIX + "." + proxyClassName;
         }
         Class<T> proxyClass = null;
         Class<?> originalClass = bean != null ? bean.getBeanClass() : proxiedBeanType;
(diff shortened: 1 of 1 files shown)


jamezp:
Converted to a draft as I got local failures I missed as I couldn't get the 6.0 branch to work and forgot :)

PR #3532, opened 2026-09-22
Adapt Core to Weld API cleanup

### LINKED ISSUE #3533: Cleanup before Weld 7 Final release
There is a few tasks I am aware of that should be done before we air the Final release.
Those include but are not limited to:
* Removal of deprecated parts
* Deprecation of APIs that were moved into CDI over the years
* Documentation (not a full overhaul, lack time for that, but I am aware of some outdated pieces)
* Some dep. changes

This will span across both API and core repos.

manovotn:
Adaptation of core to https://github.com/weld/api/pull/335.

Contains snapshot version of Weld API, hence draft.


manovotn:
Failures are related to WFLY still using the now removed `LiteExtensionTranslator` constructor variant.
Fix should be trivial and exists since Weld 5.1 so it should cover all the cases in which current WFLY uses Weld.

manovotn:
WFLY PR - https://github.com/wildfly/wildfly/pull/20429

I ran the incontainer/TCK tests locally against patched WFLY and got a pass.

